PR #27, opened 2020-02-16
Refactor Dockerfiles

inductor:
- Use `COPY` rather than `ADD` for cache strategy(ADD never generates cache)
- Use Node.js 12(and newer than 12.15 for important CVE fix) as it is the current LTS version


gustavosbarreto:
Thank you for your pull request

gustavosbarreto:
@all-contributors please add @inductor for infra

PR #34, opened 2020-02-26
Add the device features page

leonardojoao:


leonardojoao:
Problem with others bottoms in list devices, because does not work edit name, terminal and delete not function.

PR #56, opened 2020-03-05
api: move db operations to mongo package

eduardoveiga:


eduardoveiga:
to rename an index in mongoDB it's necessary drop and recreate the index
example

```
Version: 3,
		Up: func(db *mongo.Database) error{
		    _, err := db.Collection("users").UpdateMany(context.Background(), bson.M{}, bson.M{"$rename": bson.M{"username": "user_name"}})
		_, err = db.Collection("users").Indexes().DropOne(context.TODO(), "username")    
		 mod := mongo.IndexModel{
                Keys:    bson.D{{"user_name", 1}},
                Options: options.Index().SetName("user_name").SetUnique(true),
```

PR #60, opened 2020-03-13
ssh, api: detect when device goes offline

eduardoveiga:


api/main.go
```diff
@@ -84,6 +83,19 @@ func main() {
 
 	publicAPI := e.Group("/api")
 	internalAPI := e.Group("/internal")
+	internalAPI.GET("/devices/:uid/offline", func(c echo.Context) error {
```
gustavosbarreto: Using GET or HEAD on a resource URL, should NEVER change the resource. Please change it to POST

pkg/revdial/revdial.go
```diff
@@ -405,3 +408,7 @@ func ConnHandler(upgrader websocket.Upgrader) http.Handler {
 		d.matchConn(wsconnadapter.New(wsConn))
 	})
 }
+
+func (d *Dialer) Isonline() bool {
```
gustavosbarreto: IsOnline instead of Isonline

PR #63, opened 2020-03-18
api,ssh: update dependencies and remove unused ones 

eduardoveiga:


gustavosbarreto:
goherokuname is no longer used by api then you can safely remove from glide

PR #65, opened 2020-03-24
api: add pagination to Devices and Sessions list

eduardoveiga:


api/pkg/store/mongo/store.go
```diff
@@ -210,8 +218,46 @@ func (s *Store) UpdateDeviceStatus(ctx context.Context, uid models.UID, online b
 	return nil
 }
 
-func (s *Store) ListSessions(ctx context.Context) ([]models.Session, error) {
+func (s *Store) CountDevices(ctx context.Context) (int64, error) {
+
+	if tenant := store.TenantFromContext(ctx); tenant != nil {
```
gustavosbarreto: We need to count all devices even if tenant id is not available in the context

api/pkg/store/mongo/store.go
```diff
@@ -210,8 +218,46 @@ func (s *Store) UpdateDeviceStatus(ctx context.Context, uid models.UID, online b
...
+		return count, err
+	}
+
+	return 0, nil
+
+}
+
+func (s *Store) CountSessions(ctx context.Context) (int64, error) {
+
+	if tenant := store.TenantFromContext(ctx); tenant != nil {
```
gustavosbarreto: We need to count all sessions even if tenant id is not available in the context

PR #74, opened 2020-04-15
Add policy to the only owner for rename and delete device

leonardojoao:


api/main.go
```diff
@@ -172,16 +172,23 @@ func main() {
 		if err != nil {
 			return err
 		}
-
```
otavio, line 175: Non related change.

leonardojoao: Done

PR #82, opened 2020-04-23
Add wss in terminalDialog

leonardojoao:


ui/src/components/terminal/TerminalDialog.vue
```diff
@@ -133,7 +142,7 @@ export default {
         })
         .join('&');
 
-      this.ws = new WebSocket(`ws://${location.host}/ws/ssh?${params}`);
+      this.ws = new WebSocket(this.protocolConnectionURL+`://${location.host}/ws/ssh?${params}`);
```
otavio, line 145: As this seems to be used only here, wouldn't be better to use a local variable here and use it to generate the URL?

gustavosbarreto: @leonardojoao

leonardojoao: I did it this way to avoid asynchronism, it was guaranteed that we would not have a problem.

ui/src/components/terminal/Terminal.vue
```diff
@@ -19,7 +19,7 @@ export default {
 
   data() {
     return {
-      protocolConnectionURL: ''
+      // protocolConnectionURL: ''
```
otavio, line 22: No dead code please.

leonardojoao: Sorry, done.

Issue #79, opened 2020-04-22
UI: devices and sessions Ordering is not working (pagination is not working properly)


S-Chinthana:
HI, Shellhub v0.2.4 is now working flawlessly, but there is some sort of UI issues on v0.2.4-UI. Now In new version device page and session page have pagination in it. but we can’t sort it according to ascending or descending order. Ordering is not working. 

In the session page, first, it shows the oldest session. I think this can be a small issue. also can't load all sessions into one page.  And I think it'll be much convenient if the number of items per page has a selection like [10, 20, 50, all] rather than [5,10,15, all].

gustavosbarreto:
Hi @S-Chinthana,
We're happy with your feedback.

> Now In new version device page and session page have pagination in it. but we can’t sort it according to ascending or descending order. Ordering is not working.

For now, we are considering to disable sorting in the UI because the backend doesn't support it. In a future release, we will support sorting sessions and devices properly.

> In the session page, first, it shows the oldest session. I think this can be a small issue. 

Fixed in f28c918c6a085035818a4a2235511b0c91ccf4db

> And I think it'll be much convenient if the number of items per page has a selection like [10, 20, 50, all] rather than [5,10,15, all].

I agree.

otavio:
This issue was fixed on #80. @S-Chinthana please give it a go and reopen this issue if need.

S-Chinthana:
Thanks a lot, guys, Keep up the good work!

S-Chinthana:
Hi @gustavosbarreto  and @otavio ,

I think the pagination option is only working on the session page. On the devices page, it's not working. After adding the 11th device into the shell hub server. Devices list only showing first 10 devices. I think the main issue is still there.

PR #85, opened 2020-04-28
ui: Pass pagination arguments when fetching devices

leonardojoao:


gustavosbarreto:
> Commit log is wrong.
> 
> A more meaningful would be:
> 
> ```
> ui: Pass pagination arguments when fetching devices
> 
> The pagination was not working as the UI component was not doing the request
> with the correct arguments.
> 
> Refs: #79 
> ```

@leonardojoao

PR #109, opened 2020-05-13
api: avoid devices with same name

eduardoveiga:
Restoring getDeviceByName  function to avoid renaming a device
to an already taken name.

Fixes: d44e0c884dfb39e7209252623bbd99fb0e94c447


api/pkg/services/deviceadm/service.go
```diff
@@ -68,7 +68,11 @@ func (s *service) RenameDevice(ctx context.Context, uid models.UID, name string,
 		if device.Name != name {
 			device.Name = name
 			if err := validate.Struct(device); err == nil {
-				return s.store.RenameDevice(ctx, uid, name)
+				otherDevice, _ := s.store.GetDeviceByName(ctx, name, tenant)
+				if otherDevice == nil {
+
```
otavio, line 73: Useless newline.

PR #117, opened 2020-05-15
ui: Appending linux distributions icons

noreng-jg:
Missing linux distributions icons were appended
according to o.s. release ID


ui/src/components/device/DeviceDetails.vue
```diff
@@ -158,10 +158,29 @@ export default {
       hide:true,
       device: null,
       deviceIcon: {
+        alpine: 'fl-alpine',
         arch: 'fl-archlinux',
+        centos: 'fl-centos',
+        coreos: 'fl-coreos',
+        debian: 'fl-debian',
+        devuan: 'fl-devuan',
+        'elementary OS': 'fl-elementary',
```
otavio, line 167: This seems wrong. Check the ID value to ensure it is fine.

PR #119, opened 2020-05-18
ui: search devices by name

leonardojoao:
It's now possible for the user to search devices on the view device list.


gustavosbarreto:
Can you please squash two commits into one?

PR #121, opened 2020-05-18
api: support to multiple filters

eduardoveiga:
the operations "and" and "or" are supported in the filter

### DIFF
--- a/api/pkg/store/mongo/store.go
+++ b/api/pkg/store/mongo/store.go
@@ -2,6 +2,7 @@ package mongo
 
 import (
 	"context"
+	"errors"
 	"strconv"
 	"strings"
 	"time"
@@ -12,6 +13,8 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
+
+var ErrWrongParamsType = errors.New("wrong parameters type")
 
 type Store struct {
 	db *mongo.Database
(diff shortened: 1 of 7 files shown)


api/pkg/store/mongo/store.go
```diff
@@ -25,37 +25,58 @@ func NewStore(db *mongo.Database) *Store {
...
-			if err != nil {
-				return nil, 0, err
 			}
-			queryFilter = append(queryFilter, bson.M{
-				filter.Params.Name: bson.M{
-					"$eq": operator,
-				},
-			})
 		}
+		if filter.Type == "operator" {
```
otavio, line 59: We have the switch case, this should be reworked and keep using it as it makes the code easier to read.

Please see if you can get this to work.

otavio: It seems a cleaned code, preserving your logic, would be:

```diff
diff --git a/api/pkg/store/mongo/store.go b/api/pkg/store/mongo/store.go
index c320c7c..beddd6d 100644
--- a/api/pkg/store/mongo/store.go
+++ b/api/pkg/store/mongo/store.go
@@ -28,7 +28,8 @@ func (s *Store) ListDevices(ctx context.Context, perPage, page int, filters []mo
 	var queryMatch []bson.M
 	var queryFilter []bson.M
 	for _, filter := range filters {
-		if filter.Type == "property" {
+		switch filter.Type {
+		case "property":
 			switch filter.PropertyParams.Operator {
 			case "like":
 				queryFilter = append(queryFilter, bson.M{
@@ -53,26 +54,24 @@ func (s *Store) ListDevices(ctx context.Context, perPage, page int, filters []mo
 						"$eq": operator,
 					},
 				})
-
 			}
-		}
-		if filter.Type == "operator" {
+		case "operator":
+			operator := nil
 			switch filter.OperatorParams.Name {
 			case "and":
-
-				queryMatch = append(queryMatch, bson.M{
-					"$match": bson.M{"$and": queryFilter},
-				})
-				queryFilter = []bson.M{}
+				operator = "$and"
 			case "or":
-				queryMatch = append(queryMatch, bson.M{
-					"$match": bson.M{"$or": queryFilter},
-				})
-				queryFilter = []bson.M{}
+				operator = "$or"
 			}
-		}
 
+			queryMatch = append(queryMatch, bson.M{
+				"$match": bson.M{operator: queryFilter},
+			})
+
+			queryFilter = []bson.M{}
+		}
 	}
+
 	if len(queryFilter) > 0 {
 		queryMatch = append(queryMatch, bson.M{
 			"$match": bson.M{"$or": queryFilter},
```

However, it is not clear to me why you are resetting `queryFilter` when selecting the operators. It seems odd to me.

eduardoveiga: QueryFilter is an intermediate slice to to append each filter until I got the operator
in this moment I append all values acumulated in QueryFilter to respective and/or groups in QueryMatch slice. 

To restart this routine I reset QueryFilter values.  @gustavosbarreto do you know another aproach to implement this task? This was my first

api/pkg/store/mongo/store.go
```diff
@@ -18,6 +18,47 @@ var wrongType error = errors.New("wrong parameters type")
 
 const match string = "$match"
 
+var lookupDevices bson.M = bson.M{
+
```
otavio, line 22: Drop newline.

api/pkg/store/mongo/store.go
```diff
@@ -18,6 +18,47 @@ var wrongType error = errors.New("wrong parameters type")
...
+	"$lookup": bson.M{
+		"from":         "connected_devices",
+		"localField":   "uid",
+		"foreignField": "uid",
+		"as":           "online",
+	},
+}
+
+var lookupUsers bson.M = bson.M{
+
```
otavio, line 32: Here as well.

api/pkg/store/mongo/store.go
```diff
@@ -13,6 +14,10 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+var wrongType error = errors.New("wrong parameters type")
```
otavio, line 17: typeErr

api/pkg/store/mongo/store.go
```diff
@@ -13,6 +14,10 @@ import (
...
+
+const match string = "$match"
```
otavio, line 19: The idea is good but it mixes string and consts. I'd probably avoid adding this const so the query uses a single style.

api/pkg/store/mongo/store.go
```diff
@@ -94,29 +133,9 @@ func (s *Store) ListDevices(ctx context.Context, perPage, page int, filters []mo
...
-		},
-		{
-			"$addFields": bson.M{
-				"online":    bson.M{"$anyElementTrue": []interface{}{"$online"}},
-				"namespace": "$namespace.username",
-			},
-		},
+		lookupDevices,
+		lookupUsers,
+		onlineDevices,
```
otavio, line 138: This rework is very nice; it'd be good if it comes before the addition of the multiple filters so it is even easier to follow.

ui/src/components/device/DeviceRename.vue
```diff
@@ -35,6 +35,7 @@
           You entered an invalid hostname,
           it must follow the
           <a
+            rel="noopener noreferrer"
```
gustavosbarreto, line 38: ????

api/pkg/store/mongo/store.go
```diff
@@ -13,6 +14,8 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+var typeErr error = errors.New("wrong parameters type")
```
gustavosbarreto, line 17:
```
var ErrWrongParamsType = errors.New("Wrong parameters type")
````

PR #131, opened 2020-05-22
ui: rename name to hostname, removing unused code and blank line

leonardojoao:
It's removing the unsed code and add code to list hostname.
Centralizing two itens, hostname and Operating System.

### DIFF
--- a/ui/src/components/device/DeviceList.vue
+++ b/ui/src/components/device/DeviceList.vue
@@ -53,31 +53,13 @@
             </v-tooltip>
           </template>
 
-          <template v-slot:item.uid="{ item }">
-            <v-chip
-              class="short"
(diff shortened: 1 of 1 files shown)


ui/src/components/device/DeviceList.vue
```diff
@@ -161,14 +143,15 @@ export default {
           align: 'center'
         },
         {
-          text: 'Name',
-          value: 'name'
+          text: 'Hostname',
+          value: 'hostname',
```
gustavosbarreto, line 147: Shouldn't the value be 'name' instead of  'hostname'?

leonardojoao: In the future, the API will send 'name' or 'hostname'?

PR #138, opened 2020-05-25
api: Refactor migration code to avoid using duplicated code

otavio:

### DIFF
--- a/api/pkg/store/mongo/migrations.go
+++ b/api/pkg/store/mongo/migrations.go
@@ -24,36 +24,30 @@ var migrations = []migrate.Migration{
 	{
 		Version: 2,
 		Up: func(db *mongo.Database) error {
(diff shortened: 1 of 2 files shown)


api/pkg/store/mongo/migrations.go
```diff
@@ -9,6 +9,11 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+func renameCollumn(db *mongo.Database, collectionName, collumnOrig, collumnDest string) error {
```
gustavosbarreto, line 12: `renameCollumn` refers to a table-based database.
`renameField` seems more appropriate for a document-based database.

otavio: Good catch! I made this change locally and will see if I can improve it more before resending it.

PR #139, opened 2020-05-26
rework: Minor API code rework to reduce duplication

otavio:


api/pkg/store/mongo/store.go
```diff
@@ -23,37 +23,8 @@ func NewStore(db *mongo.Database) *Store {
...
-			queryFilter = append(queryFilter, bson.M{
-				filter.Params.Name: bson.M{
-					"$eq": operator,
-				},
-			})
-		}
-	}
-
-	query := []bson.M{
+func queryDevices() []bson.M {
```
gustavosbarreto, line 26: Functions in a file should be grouped by receiver and non-receiver (and unexported) functions should appear in the end of file.

otavio: I made this change, please take a look if it looks good now.

otavio:
This code had a bunch of rework, so we can close this now.

PR #146, opened 2020-05-27
Minor fixes for cicd workflow

otavio:
We reworked cicd workflow, especially the commit message one, to allow the use of numbers, spaces, and comma for commit scope as this happens often.

Another change was as a consequence of most PR being marked as failing, while this was not really the case. To fix, we restrict the workflow to the 'push' event only.


otavio:
I marked people for review so you guys are aware of the changes. @gustavosbarreto please review and see if you agree.

PR #147, opened 2020-05-27
ui: creating test for the getters and removing unused code

leonardojoao:
The test is considering default variables


leonardojoao:
@gustavosbarreto what do you think of the modifications?

leonardojoao:
It's ok. It doesn't split into multiple files, because I thought of creating a getters file. But your idea is good, I'll splitting

PR #170, opened 2020-06-09
ui: removing unused code in component terminal

leonardojoao:


noreng-jg:
It had run ok, but for that I needed to delete
the folder __snapshots__ in each subdirectory inside
tests/units

PR #171, opened 2020-06-12
ssh, api: record ssh session

eduardoveiga:
records ssh session and pass to a new route in api to save all commands and outputs


pkg/models/session.go
```diff
@@ -15,9 +15,19 @@ type Session struct {
 	LastSeen      time.Time `json:"last_seen" bson:"last_seen"`
 	Active        bool      `json:"active" bson:",omitempty"`
 	Authenticated bool      `json:"authenticated" bson:"authenticated"`
+	Log           bool      `json:"log" bson:"log"`
```
gustavosbarreto, line 18: Recorded

PR #180, opened 2020-06-15
ui: Updating eslint packages

### LINKED ISSUE #177: ui: Update eslint and eslint-plugin-vue
The versions we're using are quite old and it'd be good if we upgrade them.

```
eslint              5.16.0  5.16.0   7.2.0  ui
eslint-plugin-vue    5.2.3   5.2.3   6.2.2  ui
```

noreng-jg:
The plugins of eslint and eslist-plugin-vue were updated to the versions
7.2.0 and 6.2.2, respectively. Correction to some files were also set
to the default rules.
Closes #177


ui/src/components/session/SessionList.vue
```diff
@@ -56,13 +56,6 @@
           </template>
 
           <template v-slot:item.username="{ item }">
-            <v-tooltip
-              v-if="!item.authenticated"
-              bottom
-            >
-              <template #activator="{ on }" />
-              <span>Unauthorized</span>
-            </v-tooltip>
```
leonardojoao, line 65: Why did you remove?

otavio: This is not used. Please check the code.

ui/src/components/session/SessionList.vue
```diff
@@ -56,7 +56,23 @@
...
+                  v-on="on"
+                >{{ item.username }}</span>
+              </template>
+              <span v-if="!item.authenticated">Unauthorized</span>
+            </v-tooltip>
+            <template
+              v-if="item.authenticated"
+            >
+              {{ item.username }}
+            </template>
```
otavio, line 75: Change the commit log for something more meaningful, similar to:

```
ui: Fix unauthenticated tooltip regression

The unauthenticated tooltip were not being shown when we hover
the mouse over the user icon while it should.

The issue has been spot by the eslint update and this commit fixes it.
```

also, this commit should be done before the update, if it comes in this PR.

PR #187, opened 2020-06-18
ui: create component to close session

leonardojoao:
This remove confirm dialog, because the project used the convention to use
components in actions.

### DIFF
--- a/ui/src/components/session/SessionClose.vue
+++ b/ui/src/components/session/SessionClose.vue
@@ -1,0 +1,76 @@
+<template>
+  <fragment>
+    <v-tooltip bottom>
+      <template #activator="{ on }">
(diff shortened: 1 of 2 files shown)


ui/src/components/session/SessionClose.vue
```diff
@@ -0,0 +1,76 @@
...
+      v-model="dialog"
+      max-width="400"
+    >
+      <v-card>
+        <v-card-title class="headline grey lighten-2 text-center">
+          Are you sure?
+        </v-card-title>
+
+        <v-card-text class="mt-4 mb-3 pb-1">
+          You are going to remove connection for this device
```
gustavosbarreto, line 25: s/remove/close/g

PR #192, opened 2020-06-18
api: add pending flag to Device

eduardoveiga:
when adding a new device it will be waiting the user aproval before
allow connections

created routes to allow a new device and list only pending devices


api/main.go
```diff
@@ -142,11 +154,14 @@ func main() {
 			Page    int    `query:"page"`
 			PerPage int    `query:"per_page"`
 			Filter  string `query:"filter"`
+			Pending int    `query:"pending"`
```
otavio, line 157: Why this is not a boolean? Using an integer here seems wrong. What `pending` as 25 means?

api/pkg/services/deviceadm/service.go
```diff
@@ -41,8 +42,8 @@ func (s *service) ListDevices(ctx context.Context, perPage, page int, filterB64
 	if err := json.Unmarshal([]byte(raw), &filter); len(raw) > 0 && err != nil {
 		return nil, 0, err
 	}
-
-	return s.store.ListDevices(ctx, perPage, page, filter)
+	pendingBool := (pending == 1)
```
otavio, line 45: This is also going to avoid this conversion.

leonardojoao:
@gustavosbarreto can you check the UI?

PR #193, opened 2020-06-23
ui: Update lists in pagination

### LINKED ISSUE #189: ui: not update the list session when closing the connection
You can close the connection in the list session, and that connection is terminated with the device. But the list data is not updated.

### LINKED ISSUE #190: ui: not update the list devices when delete a device
The situation:

The user has 13 devices and 10 devices per page. It has two pages and the first page has 10 devices on the list. When deleting a device, the list is not updated and the page has 9 devices and 2 devices in other page.

noreng-jg:
The lists of sessions and devices are now automatically
updated through an event handler, without the need to refresh page
to see them.
This closes #189 and closes #190.

### DIFF
--- a/ui/src/components/device/DeviceDelete.vue
+++ b/ui/src/components/device/DeviceDelete.vue
@@ -73,9 +73,11 @@ export default {
   methods: {
     async remove() {
       await this.$store.dispatch('devices/remove', this.uid);
+      this.dialog = !this.dialog;
       if (this.redirect) {
         this.$router.push('/devices');
       }
+      this.$emit('update');
     },
   },
 };
--- a/ui/src/components/session/SessionList.vue
+++ b/ui/src/components/session/SessionList.vue
@@ -128,6 +128,7 @@
             <SessionClose
               v-if="item.active"
               :session="item"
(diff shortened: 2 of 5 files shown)


ui/src/components/device/DeviceDelete.vue
```diff
@@ -76,6 +76,7 @@ export default {
       if (this.redirect) {
         this.$router.push('/devices');
       }
+      this.$emit('update', true);
```
leonardojoao, line 79: Why not just use `this. $ Emit ('update')`?

ui/src/components/session/SessionList.vue
```diff
@@ -221,6 +219,20 @@ export default {
...
+
+    refresh(params) {
+      if (params) {
+        this.getSessions();
+      }
+    },
+
+    async getSessions() {
+      const data = { perPage: this.pagination.itemsPerPage, page: this.pagination.page };
+      await this.$store.dispatch('sessions/fetch', data);
```
leonardojoao, line 231: The line 231 and 233 isn't equal?

PR #202, opened 2020-07-03
cicd: starting the filter of ui

gomex:
- The option "fetch-depth" is required by sonarscan
- dorny/paths-filter@v2.2.0 was used to filter by folder on each job
- The changes on Golang folder won't affect the checks on UI and vice-versa
- Added sonarscan for UI too (Specific project)

PS: Please don't merge this PR until we have specific project on sonarcloud for UI

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -23,6 +23,9 @@ jobs:
 
     - name: Check out code
       uses: actions/checkout@v2
+      with:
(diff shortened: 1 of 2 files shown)


otavio:
Is this still intended to be merged? Besides the Sonar Cloud, which we are dropping out, the rest has anything which might be good to add to the current workflow?

PR #204, opened 2020-07-09
ui: add settings view

leonardojoao:


gustavosbarreto:
See #215

ui/src/router/index.js
```diff
@@ -64,6 +64,21 @@ const router = new Router({
       name: 'firewalls',
       component: () => import('./../views/FirewallRules.vue'),
     },
+    {
+      path: '/settings',
+      name: 'settings',
+      component: () => import(/* webpackChunkName: 'devices' */ './../views/Settings.vue'),
```
leonardojoao, line 70: webpackChunkName: 'devices' or /* webpackChunkName: 'settings'?

api/user/service.go
```diff
@@ -24,12 +24,8 @@ func NewService(store store.Store) Service {
...
 	user, _ := s.store.GetUserByTenant(ctx, tenant)
 	if user != nil {
-			if username == "" && email == "" { //change in password
-				return s.store.UpdateUser(ctx, user.Username, user.Email, password, tenant)
-			} else if password == "" { //changes in the name and email
-			if user.Username != username || user.Email != email {
-				return s.store.UpdateUser(ctx, username, email, user.Password, tenant)
-			}
+		if user.Username != username || user.Email != email {
+			return s.store.UpdateUser(ctx, username, email, user.Password, tenant)
```
otavio, line 28: Could it be done like this?

```
func (s *service) UpdateDataUser(ctx context.Context, username, email, password, tenant string) error {
	user, _ := s.store.GetUserByTenant(ctx, tenant)
	if user == nil {
                return ErrUnauthorized
        }
	
        return s.store.UpdateUser(ctx, username, email, user.Password, tenant)
}
```

api/store/mongo/store.go
```diff
@@ -827,6 +827,34 @@ func (s *Store) GetRecord(ctx context.Context, uid models.UID) ([]models.Recorde
...
+
+    if password!="" {
+			if _, err := s.db.Collection("users").UpdateOne(ctx, bson.M{"tenant_id": tenant}, bson.M{"$set": bson.M{"password": password}}); err!= nil {
+					return err
+				}
+			}
+
+    return nil
+}
+
```
eduardoveiga, line 857: By this way all logic of what will be updated is inside store, not services. @gustavosbarreto it's ok?

eduardoveiga, line 857: By this way all logic of what will be updated is inside store, not services. @gustavosbarreto it's ok?

ui/src/components/setting/SettingProfile.vue
```diff
@@ -0,0 +1,304 @@
...
+        newPassword: this.newPassword,
+      };
+
+      try {
+        await this.$store.dispatch('users/put', data);
+        this.$store.dispatch('modals/showSnackbarSuccess', true);
+      } catch {
+        this.$store.dispatch('modals/showSnackbarError', true);
+      }
+      this.enableEdit();
```
leonardojoao, line 299: I would put in try, because the request fails the dialog not close.

PR #231, opened 2020-07-21
api: store tests

gustavosbarreto:


otavio:
I think the baseline of code is very good. Do you intend to extend the tests before merging this or add it later?

gustavosbarreto:
> I think the baseline of code is very good. Do you intend to extend the tests before merging this or add it later?

add it later

PR #266, opened 2020-08-05
agent: Use NegotiateAPIVersion for docker client

gustavosbarreto:
WithAPIVersionNegotiation makes the Docker client negotiate down to a lower
version if Docker's current API version is newer than the server version.

### DIFF
--- a/agent/selfupdater/updater_docker.go
+++ b/agent/selfupdater/updater_docker.go
@@ -178,6 +178,8 @@ func NewUpdater(_ string) (Updater, error) {
 		return nil, err
 	}
 
+	api.NegotiateAPIVersion(context.Background())
+
 	return &dockerUpdater{api: api}, nil
 }
 


sixhills:
I have tested this as follows, in an attempt to reproduce an issue reported by @Buadhai when using an older apt-installed version of Docker on a Raspberry Pi.

The SSH Gateway (server) is a private Debian 10 host.  The agent is to be installed on a Raspberry Pi 4 with the latest Raspberry Pi OS, based on Debian 10 Buster.  Docker not previously installed on this device.

```
# apt update
# apt upgrade
# apt install docker.io
# docker --version
Docker version 18.09.1, build 4c52b90
# apt install docker-compose
# docker-compose --version
docker-compose version 1.21.0, build unknown
# usermod -aG docker pi
```
This is the same version used by @Buadhai.

Check Docker installation works:

`# docker run hello-world`

Attempted to register the agent by running the curl command specified by the server.  This returned no errors but the registration failed: the device pending and rejected lists on the server were empty.  These are the symptoms reported by @Buadhai but he also saw an error in syslog:

> msg="Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.39"

This error was NOT seen in syslog during this test.

Repeated the registration attempt using the new agent version:

```
# docker rm shellhub
# docker run -d \
       --name=shellhub \
       --restart=on-failure \
       --privileged \
       --net=host \
       --pid=host \
       -v /:/host \
       -v /var/run/docker.sock:/var/run/docker.sock \
       -v /etc/passwd:/etc/passwd \
       -v /etc/group:/etc/group \
       -e SERVER_ADDRESS=http://<address of server> \
       -e PRIVATE_KEY=/host/etc/shellhub.key \
       -e TENANT_ID=<tenant id> \
       gustavosbarreto/shellhub-agent:arm64v8
```
This apparently fails to start:

```
# docker ps
CONTAINER ID        IMAGE                                    COMMAND             CREATED             STATUS                          PORTS               NAMES
46efbce17d59        gustavosbarreto/shellhub-agent:arm64v8   "/bin/agent"        22 minutes ago      Restarting (1) 43 seconds ago                       shellhub
```
Nothing is seen on the server.

Repeated in the foreground:

```
# docker stop shellhub
# docker rm shellhub
# docker run \
    --name-shellhub \
   etc
```
This reported:

> standard_init_linux.go:207: exec user process caused "exec format error"

otavio:
@gustavosbarreto I think Raspian uses ARM32. So it needs to be the other arch.

@sixhills could you check the output of:
```
file /bin/bash
```

So we check the file format your system is using?

gustavosbarreto:
> @gustavosbarreto I think Raspian uses ARM32. So it needs to be the other arch.
> 
> @sixhills could you check the output of:
> 
> ```
> file /bin/bash
> ```
> 
> So we check the file format your system is using?

@otavio @sixhills 

Replace the `arm64v8` tag with `arm32v6` in the `docker run` command.

sixhills:
@gustavosbarreto @otavio Yes, Raspberry Pi is 32-bit ARM.  There's a 64-bit beta available but not widely used and not yet recommended.

```
# file /bin/bash
/bin/bash: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=3e5e2847bbc51da2ab313bc53d4bdcff0faf2462, stripped
```

Replacing the arm64v8 tag with arm32v6 in the docker run command works!  The device is registered on the server succesfully and I can ssh to it via the server.  Great work.  My only concern is that I wasn't able to reproduce the "client version 1.40 is too new" error seen by @Buadhai.

otavio:
So it seems it worked just fine. I think we can merge this PR then.

gustavosbarreto:
> My only concern is that I wasn't able to reproduce the "client version 1.40 is too new" error seen by @Buadhai.

@sixhills Where you unable to reproduce using the default install script or with custom docker run command?

sixhills:
@gustavosbarreto  The error message "client version 1.40 is too new" wasn't logged to syslog at all during my testing: not with the default install script and not with the custom docker run command.  I'm wondering whether there's something non-standard on @Buadhai's system that causes this.

PR #268, opened 2020-08-05
Custom keepalive interval

### LINKED ISSUE #215: [Feature] Custom keep alive intervall
To limit bandwidth we would like to change the interval the device is calling home. For us it would be totally fine if the device only contacts the server once every hour. 

To access the shell there could be two options:
- let the server know to establish a ssh connection next time he calls home and notify the user that a ssh session has been established
- let the user predefine commands he likes to sent to the server. Next time the server calls home he will then execute them.

Related [question](https://github.com/shellhub-io/shellhub/issues/102)

gustavosbarreto:
To set a custom keepalive interval just append `keepalive_interval` param (value in seconds) to the install script, e.g.

`http://localhost/install.sh?tenant_id=xxx-xxxx....&keepalive_interval=60`

This closes #215

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -28,9 +28,10 @@ import (
 var AgentVersion string
 
 type ConfigOptions struct {
-	ServerAddress string `envconfig:"server_address"`
-	PrivateKey    string `envconfig:"private_key"`
-	TenantID      string `envconfig:"tenant_id"`
+	ServerAddress     string `envconfig:"server_address"`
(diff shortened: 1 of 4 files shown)


gustavosbarreto:
@jwillmer Will this fits for your use case?

jwillmer:
Nice! Yes, I think that will solve it. Next step would be to be able to change it in the UI but for now I think this will do 👍

PR #269, opened 2020-08-06
api: Return total firewall count when listing firewalls

leonardojoao:


api/routes/firewall.go
```diff
@@ -26,7 +27,9 @@ func GetFirewallRuleList(c apicontext.Context) error {
 	// TODO: normalize is not required when request is privileged
 	query.Normalize()
 
-	rules, _, _ := svc.ListRules(c.Ctx(), *query)
+	rules, count, _ := svc.ListRules(c.Ctx(), *query)
```
gustavosbarreto, line 30: Return error if err is not nil

PR #283, opened 2020-08-11
Topic/add try catch device

leonardojoao:


ui/tests/index.js
```diff
@@ -1,5 +1,9 @@
 import Vue from 'vue';
 import Vuetify from 'vuetify';
 
+import SnackbarError from '@/components/snackbar/SnackbarError';
+
+Vue.component('SnackbarError', SnackbarError);
+
```
otavio, line 7: This commit should be included in the previous one. Please make sure tests are running on all commits.

leonardojoao: Should I fixup the last two commits?

otavio: Yes.

PR #285, opened 2020-08-12
api: only allow lowercase hostname

### LINKED ISSUE #284: Upper-case letters in device names cause "Invalid session target" failure
I'm not sure if this is a minor bug or a undocumented restriction but device names containing upper-case letters don't work. For example, if I rename a device in the web UI from its default MAC address to something more readable, like "Pi41", and attempt to ssh to it from another host using the "ssh user@SSHID" syntax, the connection fails with "Invalid session target". Renaming the device to "pi41" works.

eduardoveiga:
fix #284

### DIFF
--- a/api/store/mongo/migrations.go
+++ b/api/store/mongo/migrations.go
@@ -2,7 +2,9 @@ package mongo
 
 import (
 	"context"
+	"strings"
 
+	"github.com/shellhub-io/shellhub/pkg/models"
 	migrate "github.com/xakep666/mongo-migrate"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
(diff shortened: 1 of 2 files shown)


api/store/mongo/migrations.go
```diff
@@ -141,6 +143,31 @@ var migrations = []migrate.Migration{
...
+			defer cursor.Close(context.TODO())
+			for cursor.Next(context.TODO()) {
+				device := new(models.Device)
+				err := cursor.Decode(&device)
+				if err != nil {
+					return err
+				}
+
+				device.Name = strings.ToLower(device.Name)
+				_, err = db.Collection("devices").UpdateOne(context.TODO(), bson.M{"uid": device.UID}, bson.M{"$set": bson.M{"name": strings.ToLower(device.Name)}})
```
otavio, line 162: If it fails here, we need to do something. We cannot blindly ignore the error.

eduardoveiga: now i'm returning thr error everytime the update fails. It's ok?

PR #288, opened 2020-08-12
ui: add new snackbarsuccess and refactor snackbarerror

noreng-jg:
Now it's possible to provide a custom name to snackbar prop and
choose another option in case of success.


ui/src/components/snackbar/SnackbarSuccess.vue
```diff
@@ -0,0 +1,34 @@
...
+  </fragment>
+</template>
+
+<script>
+
+export default {
+  name: 'SnackbarError',
+
+  props: {
+    ok: {
```
leonardojoao, line 22: What do you think 'success'?

PR #295, opened 2020-08-14
api: Create method in store to return users

eduardoveiga:


api/store/mongo/store.go
```diff
@@ -712,13 +712,7 @@ func (s *Store) CreateFirewallRule(ctx context.Context, rule *models.FirewallRul
...
 
 func (s *Store) ListFirewallRules(ctx context.Context, pagination paginator.Query) ([]models.FirewallRule, int, error) {
-	query := []bson.M{
-		{
-			"$sort": bson.M{
-				"priority": 1,
-			},
-		},
-	}
+	query := []bson.M{}
```
otavio, line 715: This is not related to this commit. Please move it to another PR.

PR #333, opened 2020-08-24
cicd: workflow refactor

gomex:
the build agent shouldn't push image on PR

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -39,6 +39,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: login to docker hub
+        if: "github.event_name != 'pull_request'"
         run: echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
 
       - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/build-agent.yml
```diff
@@ -9,12 +9,6 @@ on:
     - 'agent/**'
     - '.github/workflows/build-agent.yml'
 
-  pull_request:
-    branches: [ master ]
-    paths:
-    - 'agent/**'
-    - '.github/workflows/build-agent.yml'
```
otavio, line 16: It must be build but Docker image should not to be pushed.

.github/workflows/build-agent.yml
```diff
@@ -8,7 +8,7 @@ on:
     paths:
     - 'agent/**'
     - '.github/workflows/build-agent.yml'
-
+
```
otavio, line 11: Useless change.

.github/workflows/build-agent.yml
```diff
@@ -47,21 +48,29 @@ jobs:
...
         run: echo ::set-env name=RELEASE_VERSION::${GITHUB_REF#refs/*/}
 
-      - name: build image
-        if: "!contains(github.ref, 'refs/tags/v')"
+      - name: build image 
+        if: ${{ github.event_name == 'pull_request' }} 
+        run: |
+          docker buildx build -f agent/Dockerfile.${{ matrix.dockerfile }} \
+            --tag shellhubio/agent:${{ github.sha }}-${{ matrix.dockerfile }} \
+            --platform linux/${{ matrix.arch }} .
```
otavio, line 56: This and the below task seem the same. Aren't they?

.github/workflows/commit.yml
```diff
@@ -3,6 +3,9 @@ on:
   push:
     branches-ignore:
       - 'dependabot/*'
+  
+  pull_request:
+    branches: [ master ]
```
otavio, line 8: Please split the commit check in another PR so it can go in fast; also this PR is changing two different things which are not a good practice and we should avoid whenever possible.

.github/workflows/build-agent.yml
gomex: No, this step don't have "--push" option.

.github/workflows/build-agent.yml
gomex: Done

.github/workflows/commit.yml
gomex: Removed.

.github/workflows/build-agent.yml
otavio: Couldn't you have a docker push as a specific task?

gomex: I can try, but my idea is keep it simple. The buildx execution would be responsible for everything. 
I will work to separate these steps

PR #350, opened 2020-08-28
agent: Don't set ctty fd

gustavosbarreto:
`Ctty` in `SysProcAttr` should be set to a fd in a child process, but we were setting it to a fd in the parent process.
This was working by accident, but now a panic is raised by `os/exec`.

See: https://github.com/golang/go/issues/29458


gustavosbarreto:
@all-contributors please add @gustavosbarreto  for code

PR #374, opened 2020-09-01
api, ui: Add form errors when update user settings

noreng-jg:


api/user/service.go
```diff
@@ -17,25 +18,42 @@ type service struct {
 	store store.Store
 }
 
-func NewService(store store.Store) Service {
-	return &service{store}
+type Invalid struct {
```
otavio, line 21: My biggest concern is about the naming. This seems to be not clear about what is indeed "invalid". I think if it was named 'InvalidField' it'd be more meaningful.

api/user/service.go
```diff
@@ -17,25 +18,42 @@ type service struct {
...
+	Field string
```
otavio, line 22: I'd use 'name'

api/user/service.go
```diff
@@ -17,25 +18,42 @@ type service struct {
...
 }
 
-func (s *service) UpdateDataUser(ctx context.Context, username, email, currentPassword, newPassword, tenant string) error {
+func NewService(store store.Store) (Service *service) {
+	return &service{store} 
+}
+
+func (s *service) UpdateDataUser(ctx context.Context, username, email, currentPassword, newPassword, tenant string) ([]Invalid, error) {
+	var invalids []Invalid
```
otavio, line 30: invalid_fields

api/user/service.go
```diff
@@ -17,25 +18,42 @@ type service struct {
...
 	}
+
+	var checkName, checkMail bool
+
 	user, err = s.store.GetUserByUsername(ctx, username)
 	if err == nil && user.TenantID != tenant {
-		return ErrUnauthorized
+		invalidDuplicated.Field = "username"
+		checkName = true
+		invalids = append(invalids, invalidDuplicated)
```
otavio, line 47: I think you can do something like:

```go
invalid_fields = append(invalid_fields, InvalidField { name: "username })
```

ui/src/components/setting/SettingProfile.vue
```diff
@@ -281,10 +283,25 @@ export default {
...
+        this.$store.dispatch('modals/showSnackbarSuccessAction', this.$success.profileData);
         this.enableEdit();
-      } catch {
-        this.$store.dispatch('modals/showSnackbarErrorDefault');
+      } catch (error) {
+        if (error.response.status === 409) { // user data already exists
+          error.response.data.forEach((item) => {
+            if (item.Field === 'username') {
+              this.$refs.data.setErrors({
+                username: ['This username already exists'],
```
otavio, line 293: The error message should come from the API so it is just rendered here.

api/user/service.go
```diff
@@ -11,7 +11,7 @@ var ErrUnauthorized = errors.New("unauthorized")
 var ErrConflict = errors.New("conflict")
 
 type Service interface {
-	UpdateDataUser(ctx context.Context, username, email, currentPassword, newPassword, tenant string) error
+	UpdateDataUser(ctx context.Context, username, email, currentPassword, newPassword, tenant string) ([]Invalid, error)
```
otavio, line 14: Squash on the initial commit.

api/user/service.go
```diff
@@ -8,34 +8,56 @@ import (
...
+	UpdateDataUser(ctx context.Context, username, email, currentPassword, newPassword, tenant string) ([]InvalidField, error)
 }
 
 type service struct {
 	store store.Store
 }
 
+const (
+	conflictName = "This username already exists"
+	conflictMail = "This email already exists"
```
otavio, line 23: conflictEMail

PR #383, opened 2020-09-03
pkg: set leveled logger for retryablehttp client

gustavosbarreto:


otavio (changes requested):
Please improve the commit log. A nice thing to have is a small quote of how it was before and after this change.

gustavosbarreto:
> Please improve the commit log. A nice thing to have is a small quote of how it was before and after this change.

There are no significant changes until `WithLogger` is used

otavio (approved):

PR #385, opened 2020-09-05
cicd: add filter for monorepo trigger

gomex:
Expected behavior:

Changes to the golang folders trigger go jobs.
Changes to the ui folder trigger js jobs.

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -23,9 +23,22 @@ jobs:
 
     - name: Check out code
       uses: actions/checkout@v2
+    
+    - uses: dorny/paths-filter@v2.2.0
+      id: filter
+      with:
+        # inline YAML or path to separate file (e.g.: .github/filters.yaml)
+        filters: |
+          ui:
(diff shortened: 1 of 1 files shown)


.github/workflows/qa.yml
```diff
@@ -23,17 +23,30 @@ jobs:
...
+            - 'ui/**'
+          go:
+            - 'api/**'
+            - 'agent/**'
+            - 'pkg/**'
+            - 'ssh/**'
 
     - name: Set up Go 1.x [Go]
-      if: matrix.project != 'ui'
+      if: matrix.project != 'ui' && steps.filter.outputs.go == 'true'
```
otavio, line 41: This way if it changes both it won't be triggered.

otavio (changes requested):

.github/workflows/qa.yml
otavio: We ought to just check the filter here.

otavio (approved):

PR #390, opened 2020-09-10
api: delete device from connected_device on removeDevice

eduardoveiga:
this fix avoid GetStats from count the removed device as an online device

### DIFF
--- a/api/store/mongo/store.go
+++ b/api/store/mongo/store.go
@@ -193,7 +193,8 @@ func (s *Store) DeleteDevice(ctx context.Context, uid models.UID) error {
 		return err
 	}
 
-	return nil
+	_, err := s.db.Collection("connected_devices").DeleteMany(ctx, bson.M{"uid": uid})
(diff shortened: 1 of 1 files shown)


api/store/mongo/store.go
```diff
@@ -193,7 +193,8 @@ func (s *Store) DeleteDevice(ctx context.Context, uid models.UID) error {
 		return err
 	}
 
-	return nil
+	_, err := s.db.Collection("connected_devices").DeleteMany(ctx, bson.M{"uid": uid})
```
otavio, line 196: Shouldn't this to be `DeleteOne`?

otavio (changes requested):

api/store/mongo/store.go
gustavosbarreto: No, because there can be more than one document.

gustavosbarreto (approved):

PR #391, opened 2020-09-10
ui: reset validation firewall dialog

noreng-jg:
Form errors in firewall rules are now reseted when closing the dialog.


leonardojoao (changes requested):
You can call close inside update, because you are going to reuse the code.

leonardojoao (approved):

PR #422, opened 2020-09-21
ui: Update unit tests in device list

### LINKED ISSUE #417: [unit-test] ui: Add missing tests in ui for rendering each item in device list
You can create a test to verify items of the v-data-table. To this, use the declared variable on scope describe, or create const with data related to use in all tests.

noreng-jg:
A verification of data-table items in devicelist was added.
This closes #417.

### DIFF
--- a/ui/src/components/device/Device.vue
+++ b/ui/src/components/device/Device.vue
@@ -39,6 +39,7 @@
             <v-badge
               :content="getNumberPendingDevices"
               :value="getNumberPendingDevices"
+              data-test="badge-field"
               overlap
               inline
               color="success"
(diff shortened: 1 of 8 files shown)


noreng-jg:
I couldn't find a way to guarantee a comparison of the object fields names individually
in each template tag inside the data-table.
These tests only grant that the device items are loaded as expected.

leonardojoao (changes requested):
You are verifying the number of items in the list, but you are not verifying the items rendered in the template. It's important to confirm the rendered items with what we expect.

otavio (dismissed):

noreng-jg:
It seems that the slot data in internal template tags of data-table rendered in
each item don't appear mounted in the list components inside
snapshot folder, data can be accessed alternatively with propsData
following this approach.

leonardojoao (approved):

PR #466, opened 2020-09-30
api: add test for api routes

eduardoveiga:


otavio (changes requested):
@eduardoveiga this looks promising ;-)

The only concern I have is that the tests ought to be refactored with small functions that cover each aspect of the API so we can better understand the use case and have more context.

PR #468, opened 2020-09-30
Topic/modal to snackbar refactor

noreng-jg:


leonardojoao (changes requested):
You can join commits, as both are related to modification.

leonardojoao (approved):

Issue #317, opened 2020-08-18
Record login activity to utmp when a pty session is created


gustavosbarreto:
Most of unix login systems record login activity to utmp when a pty session is created.
The commands like `w` and `who` read from utmp to show the users currently logged into the system.

Suggested by @sixhills on Gitter chat channel.

sixhills:
Also wtmp as a historical utmp.  I'm looking into doing this as an exercise, but new to Go and git, so steep learning curve!

sixhills:
As a first step, I have implemented session start and session end logging into the agent.   My locally-built agent now logs, when a session starts:

`time="2020-08-19T16:21:29Z" level=info msg="Session started" localaddr="192.168.11.141:46724" pty=xterm remoteaddr="157.230.203.130:443" user=pi`

and, when a session ends:

`time="2020-08-19T16:32:36Z" level=info msg="Session ended" localaddr="192.168.11.141:46724" pty=xterm remoteaddr="157.230.203.130:443" user=pi`

This satisfies my personal requirement that there should be a record of inbound connections.

I want to implement utmp updating but wondered if this logging would be a useful addition by itself?

otavio:
@sixhills I think so; it is a great addition for sure. Also, it could be the begin of #184 as it is related.

sixhills:
OK, thanks.  Small improvement to log the device:

`time="2020-08-19T18:42:46Z" level=info msg="Session ended" localaddr="192.168.11.141:56566" pty=/dev/pts/3 remoteaddr="157.230.203.130:443" user=pi
`

But it is curious that the agent always uses the names /dev/pts/0, 1, 2, ... even if these are already in use by other processes.

otavio:
It is running inside Docker? if so, it's likely under a namespace.

sixhills:
Yes, it is and that makes sense.  Sorry for silly questons!

otavio:
No problem at all; it is nice you are contributing. Looking forward for the first PR :-)

sixhills:
[...]

PR #553, opened 2020-11-23
agent: Add utmp and wtmp recording of pty sessions

sixhills:
This change implements recording of login activity to the utmp
and wtmp databases so that sessions are reported by standard
utilities such as "who", "last", "w" and "users".

The code uses the glibc functions pututline(3) and updwtmp(3)
to access the utmp and wtmp files but these functions are not
implemented within the musl libc included within the alpine
base used by the Docker agent build process. musl libc defines
PATH_UTMP as /dev/null/utmp and PATH_WTMP as /dev/null/wtmp
and pututline(3) and updwtmp(3) as null functions. The code runs
without error within an Alpine Docker container but has no
effect. It updates utmp and wtmp only when run natively or
inside a Docker container built with glibc support.

For this reason, this change is not suitable for merging but
may be useful as a basis for a more complete implementation,
either using https://skarnet.org/software/utmps/ as suggested
by @otavio or by porting pututline(3) and updwtmp(3) to Go and
including them within the agent code.

Comments and feedback on the code is very welcome as I'm not
a Go programmer and I'm sure the code can be improved.

Partially fixes https://github.com/shellhub-io/shellhub/issues/317

### DIFF
--- a/agent/sshd/server.go
+++ b/agent/sshd/server.go
@@ -1,7 +1,11 @@
 package sshd
 
+//#include <utmp.h>
+//#include <paths.h>
+//#include <stdlib.h>
+import "C"
+
 import (
-	"C"
 	"fmt"
 	"io"
 	"net"
@@ -10,10 +14,19 @@ import (
 	"strconv"
 	"sync"
 	"time"
+	"strings"
+	"unsafe"
 
 	sshserver "github.com/gliderlabs/ssh"
 	"github.com/shellhub-io/shellhub/agent/pkg/osauth"
 	"github.com/sirupsen/logrus"
+)
+
+const (
+	UserProcess = C.USER_PROCESS   // 0x7
+	DeadProcess = C.DEAD_PROCESS   // 0x8
+	PathUtmp = C._PATH_UTMP        // /var/run/utmp
+	PathWtmp = C._PATH_WTMP        // /var/log/wtmp
 )
 
 type sshConn struct {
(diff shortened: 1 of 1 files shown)


sixhills:
One point to note is that the code writes the agent pid into the utmp/wtmp records.  It would be better to use the login process pid but I don't know how to get this within server.go.  Although a pid is part of the utmp record, nothing actually uses it as far as I know, so it may not matter.

gustavosbarreto:
@sixhills awesome! I'm working on other features, but I'll look at it as soon as possible.

otavio:
@sixhills when you can an opportunity, please rebase this on top of master and take a look at the CI errors.

sixhills:
The CI errors are because musl defines one of the elements of struct utmp as "unsigned ut_addr_v6[4]" but glibc defines it as "int32_t ut_addr_v6[4]".  I don't know why this is different in the two libraries.  The code builds OK on a glibc system but needs one line changing for Alpine musl (see comment in the code).  Perhaps there's a way of expressing the assignment that works on both?  I'll need to read up about rebase - I'm still learning about git.

sixhills:
I have now removed the dependency on glibc by implementing the utmp/wtmp file handling within Go, so the same code now works on 32 and 64-bit systems and when built natively and within the Alpine Docker environment.

I looked at https://skarnet.org/software/utmps/ but this requires a daemon to be run with the agent and adds complexity, especially when the agent is run natively, so I felt it was simpler to include the required code within the agent.

If the project is happy with this approach, I'll withdraw this PR and submit a new PR with the revised code,

otavio:
Obviusly @sixhills ... that's great news indeed. Please send the new PR and thank you for your commitment on this.

PR #582, opened 2020-12-17
agent: Add utmp and wtmp recording of pty sessions (Revised)

sixhills:
This change implements recording of login activity to the utmp
and wtmp databases so that sessions are reported by standard
utilities such as "who", "last", "w" and "users".

This revised PR removes the dependency on glibc, so that the
code works when the Agent is run in an Alpine docker
environment as well as when it's run natively.

Comments and feedback on the code is very welcome as I'm
new to Go programming.

Fixes https://github.com/shellhub-io/shellhub/issues/317

### DIFF
--- a/agent/sshd/server.go
+++ b/agent/sshd/server.go
@@ -122,19 +122,19 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 
 		os.Chown(pts.Name(), uid, -1)
 
-		remoteAddr := session.RemoteAddr().String()
+		remoteAddr := session.RemoteAddr()
 
 		logrus.WithFields(logrus.Fields{
 			"user":       session.User(),
 			"pty":        pts.Name(),
 			"remoteaddr": remoteAddr,
-			"localaddr":  session.LocalAddr().String(),
+			"localaddr":  session.LocalAddr(),
 		}).Info("Session started")
 
 		ut := utmpStartSession(
 			pts.Name(),
 			session.User(),
-			remoteAddr,
+			remoteAddr.String(),
 		)
 
 		s.mu.Lock()
@@ -148,8 +148,8 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 		logrus.WithFields(logrus.Fields{
 			"user":       session.User(),
 			"pty":        pts.Name(),
-			"remoteaddr": session.RemoteAddr().String(),
-			"localaddr":  session.LocalAddr().String(),
+			"remoteaddr": remoteAddr,
+			"localaddr":  session.LocalAddr(),
 		}).Info("Session ended")
 
 		utmpEndSession(ut)
@@ -162,8 +162,8 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 
 		logrus.WithFields(logrus.Fields{
 			"user":        session.User(),
-			"remoteaddr":  session.RemoteAddr().String(),
-			"localaddr":   session.LocalAddr().String(),
+			"remoteaddr":  session.RemoteAddr(),
+			"localaddr":   session.LocalAddr(),
 			"Raw command": session.RawCommand(),
 		}).Info("Command started")
 
@@ -185,8 +185,8 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 
 		logrus.WithFields(logrus.Fields{
 			"user":        session.User(),
-			"remoteaddr":  session.RemoteAddr().String(),
-			"localaddr":   session.LocalAddr().String(),
+			"remoteaddr":  session.RemoteAddr(),
+			"localaddr":   session.LocalAddr(),
 			"Raw command": session.RawCommand(),
 		}).Info("Command ended")
 	}
--- a/agent/sshd/server.go
+++ b/agent/sshd/server.go
@@ -122,12 +122,20 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 
 		os.Chown(pts.Name(), uid, -1)
 
+		remoteAddr := session.RemoteAddr().String()
+
 		logrus.WithFields(logrus.Fields{
 			"user":       session.User(),
 			"pty":        pts.Name(),
-			"remoteaddr": session.RemoteAddr().String(),
+			"remoteaddr": remoteAddr,
 			"localaddr":  session.LocalAddr().String(),
 		}).Info("Session started")
+
+		ut := utmpStartSession(
+			pts.Name(),
+			session.User(),
+			remoteAddr,
+		)
 
 		s.mu.Lock()
 		s.cmds[session.Context().Value(sshserver.ContextKeySessionID).(string)] = scmd
@@ -143,6 +151,8 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 			"remoteaddr": session.RemoteAddr().String(),
 			"localaddr":  session.LocalAddr().String(),
 		}).Info("Session ended")
+
+		utmpEndSession(ut)
 	} else {
 		u := osauth.LookupUser(session.User())
 		cmd := newCmd(u, "", "", s.deviceName, session.Command()...)
--- a/agent/sshd/utmp.go
+++ b/agent/sshd/utmp.go
@@ -1,0 +1,267 @@
+package sshd
+
+/*	The Utmpx struct is derived from the Linux definition (man 5 utmpx).
+
+	At session start, a utmp record is constructed containing the PID,
+	tty name, ID, username, remote host, source IP and time.
+	Type is set to UserProcess and the record is written to UtmpxFile.
+	If a record with the same ID exists, that record is overwritten;
+	otherwise the record is appended to the file.  The same record is
+	appended to WtmpxFile.
+
+	At session end, Type is set to DeadProcess, the User and Host fields
+	are cleared, time is updated and the record is wrtten to UtmpxFile,
+	overwriting the session start record. The same record with ID and
+	source IP address cleared is appended to WtmpxFile.
+*/
+
+import (
+	"os"
+	"unsafe"
+	"bytes"
+	"net"
+	"strings"
+	"encoding/binary"
+	"golang.org/x/sys/unix"
+	"github.com/sirupsen/logrus"
+)
+
+type ExitStatus struct {
+	ETermination	int16		// Process temination status - not used
+	EExit		int16		// Process exit status - not used
+}
+
+type Utmpx struct {
+	Type		int16		// UserProcess or DeadProcess
+	Padding		[2]byte		// Padding to align rest of struct
+	Pid		int32		// PID of the ShellHub agent
+	Line		[32]byte	// tty associated with the process
+	ID		[4]byte		// Index, last 4 characters of Line
+	User		[32]byte	// Username
+	Host		[256]byte	// Source IP address
+	Exit		ExitStatus	// Exit status - not used
+	Session		int32		// Session ID - not used
+	Tv		unix.Timeval	// Time entry was made
+	AddrV6		[4]uint32	// Source IP address. IPv4 in AddrV6[0]
+	Reserved	[20]byte	// Not used
+}
+
+const (
+	UtmpxFile   = "/var/run/utmp"
+	WtmpxFile   = "/var/log/wtmp"
+	UserProcess = 0x7		// Normal process
+	DeadProcess = 0x8		// Terminated process
+)
+
+// This function updates the utmp and wtmp files at the start of a user session
+func utmpStartSession(line, user, remoteAddr string) Utmpx {
+	var u Utmpx
+
+	u.Type = UserProcess
+	a := unix.Timeval{}
+	unix.Gettimeofday(&a)
+	u.Tv.Sec, u.Tv.Usec = a.Sec, a.Usec
+	u.Pid = int32(os.Getpid())
+
+	// remoteAddr has the form <IPv4 address>:<port>
+	// or [<IPv6 address>]:<port>
+	// Remove the port suffix and also the square brackets
+	// if IPv6, leaving the bare IPv4 or IPv6 address
+	host, _, err := net.SplitHostPort(remoteAddr)
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"ip": remoteAddr,
+		}).Warn("wrong remoteAddr format")
+	} else {
+		// Parse IP address to a standard 16-byte representation
+		ip := net.ParseIP(host)
+		// Check whether IPv4 or IPv6
+		if ip4 := ip.To4(); ip4 != nil {
+			// This is a 32-bit IPv4 address to be
+			// stored in the first element of u.AddrV6 
+			u.AddrV6[0] = binary.LittleEndian.Uint32(ip4)
+		} else {
+			// This is a 128-bit IPv6 address. Each 4 bytes
+			// of the address is stored as a 32-bit int in
+			// successive elements of u.AddrV6
+			u.AddrV6[0] = binary.LittleEndian.Uint32(ip[0:4])
+			u.AddrV6[1] = binary.LittleEndian.Uint32(ip[4:8])
+			u.AddrV6[2] = binary.LittleEndian.Uint32(ip[8:12])
+			u.AddrV6[3] = binary.LittleEndian.Uint32(ip[12:16])
+		}
+	}
+
+	line = strings.TrimPrefix(line,"/dev/")
+	// The index to the utmp record is the last 4 chars of line
+	id := line[len(line)-4:]
+
+	_ = copy(u.ID[:], id)
+	_ = copy(u.Line[:], line)
+	_ = copy(u.User[:], user)
+	_ = copy(u.Host[:], host)
+
+	updUtmp(u, id)
+	updWtmp(u)
+	return u
+}
+
+// This function updates the utmp and wtmp files at the end of a user session
+func utmpEndSession(u Utmpx) {
+	u.Type = DeadProcess
+	u.User = [32]byte{}
+	u.Host = [256]byte{}
+	a := unix.Timeval{}
+	unix.Gettimeofday(&a)
+	u.Tv.Sec, u.Tv.Usec = a.Sec, a.Usec
+
+	updUtmp(u, string(u.ID[:]))
+
+	u.ID = [4]byte{}
+	u.AddrV6 = [4]uint32{}
+
+	updWtmp(u)
+}
+
+// This function updates the utmp file by overwriting the record with index
+// id if present; otherwise by appending the new record to the file
+func updUtmp(u Utmpx, id string) {
+	file, err := os.OpenFile(
+			UtmpxFile,
+			os.O_RDWR|os.O_CREATE,
+			0644)
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"file": UtmpxFile,
+			"err": err,
+		}).Warn("Open failed")
+		return
+	}
+
+	defer file.Close()
+
+	// Lock the file
+	lk := unix.Flock_t{
+		Type: int16(unix.F_WRLCK),
+		Pid:  int32(os.Getpid()),
+		}
+
+	err = unix.FcntlFlock(file.Fd(), unix.F_SETLKW, &lk)
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"file": UtmpxFile,
+			"err": err,
+		}).Warn("Lock failed")
+		return
+	}
+
+	var ut Utmpx
+
+	// Read through the utmp file looking for a record with index id
+	for {
+		offset, err := file.Seek(0, os.SEEK_CUR)
+		if err != nil {
+			logrus.WithFields(logrus.Fields{
+				"file": UtmpxFile,
+				"err": err,
+			}).Warn("Null seek failed")
+			return
+		}
+
+		err = binary.Read(file, binary.LittleEndian, &ut)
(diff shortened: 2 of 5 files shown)


sixhills:
Should I fix the style issues reported by Revive?

otavio:
Yes, please fix the style issues so it is clean. Also, do you have a reference you used to implement this? I'd like to take a look at this so I use it as reference while reviewing the code.

sixhills:
Miscellaneous references:
man utmp 5
man pututline 3
https://code.woboq.org/userspace/glibc/login/
https://gogs.blitter.com/RLabs/goutmp
https://godoc.org/github.com/EricLagerg/go-gnulib/utmp

otavio (changes requested):
@sixhills, thanks for implementing this. It looks quite good.

First, I think this is big enough to deserve a specific file, inside the sshd server package. This will allow for a more self-contained implementation and make it easy to understand how it is glued together with the rest of the app.

Next, it'd be good to provide a small comment to explain where the definition is gathered from so anyone intending to change this code have this at disposal.

agent/sshd/server.go
```diff
@@ -33,6 +38,38 @@ type sshConn struct {
 	ctx           sshserver.Context
 }
 
+type TimeVal struct {
+	Sec  int32	// All struct fields capitalized for
+	Usec int32	// export to the encoding/binary library
+}
```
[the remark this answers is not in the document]
sixhills: Thanks for the comments.  I'll move the new code into file utmp.go and delete the TimeVal struct.  Should I place the suggested comment at the top of the new file?  How much detail is needed? Just a reference to utmpx.h or a more complete description about how the files are updated (overwrite record if present in utmp, append to wtmp) and which fields are set on session start and session end?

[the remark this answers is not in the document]
otavio: Hard to know. When you update the PR and provide it, I'll do another review and we see if it provide enough detail for me to find all information during the review process.

Also, please squash all changes into a single commit so we start the review process again.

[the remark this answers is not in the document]
sixhills: I've made the changes requested and squashed into a single commit (hope I've done that correctly!)

agent/sshd/server.go
```diff
@@ -122,13 +123,43 @@ func (s *Server) sessionHandler(session sshserver.Session) {
...
+		}
+
+		bits := strings.Split(host, ".")
+
+		b0, _ := strconv.Atoi(bits[0])
+		b1, _ := strconv.Atoi(bits[1])
+		b2, _ := strconv.Atoi(bits[2])
+		b3, _ := strconv.Atoi(bits[3])
+
+		var ip uint32
```
otavio, line 149: I believe this does not belong to the server.go file but instead should be calculated internally on `utmp` file. Either way, I think we can simplify this: https://play.golang.org/p/Q9hYrjSEDyz

agent/sshd/server.go
```diff
@@ -122,13 +123,43 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 
 		os.Chown(pts.Name(), uid, -1)
 
+		hostport := session.RemoteAddr().String()
```
otavio, line 126: `remoteAddr` seems to be a more meaningful variable name.

agent/sshd/server.go
```diff
@@ -122,13 +123,43 @@ func (s *Server) sessionHandler(session sshserver.Session) {
...
+			"remoteaddr": hostport,
 			"localaddr":  session.LocalAddr().String(),
 		}).Info("Session started")
 
+		var host string
+		if !strings.Contains(hostport, "[") {
+			host = strings.Split(hostport, ":")[0]
+		} else {
+			host = strings.Split(strings.Split(hostport, "[")[1], "]")[0]
+		}
```
otavio, line 140: This string manipulation is confusing, we need to simplify this or add a comment block explaining what it does.

agent/sshd/utmp.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+type ExitStatus struct {
+	ETermination	int16
+	EExit		int16
+}
+
+type Utmpx struct {
+	Type		int16		// UserProcess or DeadProcess
+	Pad		[2]byte
```
otavio, line 34: If `Pad` stands for padding, please use `Padding` as name here and also add the comment.

otavio (changes requested):

agent/sshd/utmp.go
sixhills: Thanks for further comments.  I'll amend tomorrow.

agent/sshd/server.go
sixhills: Thanks.  That's a much better way of doing this.  I'll modify the method for IPv6 and move it to the `utmp` file.

agent/sshd/server.go
sixhills: Agreed.

agent/sshd/server.go
sixhills: The purpose is to extract the IP address from a string that may be IPv4 or IPv6 and contains a port suffix.  I'll add a comment block but can't see a simplification.  Suggestions welcome, of course.

agent/sshd/utmp.go
sixhills: Done.  Revised PR to follow.

agent/sshd/server.go
otavio: I think you could use [LastIndexByte](https://golang.org/pkg/strings/#LastIndexByte) to lookup for last `:` and use the string before that, in [ParseIP](https://golang.org/pkg/net/#ParseIP), which is generic over IPv4 and IPv6.

sixhills: But ParseIP doesn't accept IPv6 enclosed in square brackets, so I'd still have to remove those before calling ParseIP.  Please see [https://play.golang.org/p/S8ONQyXNbu2]( https://play.golang.org/p/S8ONQyXNbu2)

otavio: In fact, the `net` package has a  specific method for this. See the below code:
```go
func newCode(remoteAddr string) {
	host, port, err := net.SplitHostPort(remoteAddr)
	fmt.Println("Extracted IP: ", host, " port: ", port, " err: ", err)
}
```

sixhills: I like that. Simple to understand and does exactly what's needed, so I'll put that in the code.

agent/sshd/utmp.go
```diff
@@ -0,0 +1,247 @@
...
+		ip4 := ip.To4()
+		if ip4 != nil {
+			u.AddrV6[0] = binary.LittleEndian.Uint32(ip4)
+		} else {
+			ip16 := ip.To16()
+			u.AddrV6[0] = binary.LittleEndian.Uint32(ip16[0:4])
+			u.AddrV6[1] = binary.LittleEndian.Uint32(ip16[4:8])
+			u.AddrV6[2] = binary.LittleEndian.Uint32(ip16[8:12])
+			u.AddrV6[3] = binary.LittleEndian.Uint32(ip16[12:16])
+		}
```
otavio, line 81: Doesn't something like this (https://github.com/openshift/coredns-mdns/blob/d3d4bb1f4a4a0e1619a419a26a3d8dcf2959d990/mdns_test.go#L17-L21) work?

otavio (changes requested):

agent/sshd/utmp.go
sixhills: I can't see an efficient way of using append to do this.

otavio: Indeed; I just think we need to add some comment here explaining both code paths. Also, the commit log needs to be fixed as it ended being lost during the revisions.

sixhills: I'm not sure how to fix the commit log. I think I've been doing the commits incorrectly but I'm not clear about how to do it.  I've been doing "git add agent/sshd/utmp.go", "git commit" with a comment, then "git push origin utmp".  I then squash the commits with "git reset --soft HEAD-2", "git commit" and "git push origin +utmp".  I'm still new to git, so would welcome advice on the correct process.

otavio: Take a look at https://linuxize.com/post/change-git-commit-message/

Otherwise, ping me on Gitter and we do it together.

sixhills: I've done a commit to reset the commit log but without any further code changes.  Further commit to come with minor code changes and additional comments.

sixhills: Commit done with improved comments.

agent/sshd/utmp.go
```diff
@@ -0,0 +1,273 @@
...
+const (
+	UtmpxFile   = "/var/run/utmp"
+	WtmpxFile   = "/var/log/wtmp"
+	UserProcess = 0x7		// Normal process
+	DeadProcess = 0x8		// Terminated process
+)
+
+// This function updates the utmp and wtmp files at the start of a user session
+func utmpStartSession(line, user, remoteAddr string) Utmpx {
+
```
otavio: This empty line should be dropped.

agent/sshd/utmp.go
```diff
@@ -0,0 +1,273 @@
...
+	unix.Gettimeofday(&a)
+	u.Tv.Sec, u.Tv.Usec = a.Sec, a.Usec
+	u.Pid = int32(os.Getpid())
+
+	// remoteAddr has the form <IPv4 address>:<port>
+	// or [<IPv6 address>]:port
+
+	// Remove the port suffix and also the square brackets
+	// if IPv6, leaving the bare IPv4 or IPv6 address
+
```
otavio: Drop the empty line and make this a single comment block such as:

```go
// foo
//
// bar
```

agent/sshd/utmp.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"ip": remoteAddr,
+		}).Warn("wrong remoteAddr format")
+	} else {
+		// Parse IP address to a standard 16-byte representation
+		ip := net.ParseIP(host)
+		// Check whether IPv4 or IPv6
+		ip4 := ip.To4()
+		if ip4 != nil {
```
otavio:
```go
if ipv4 := ip.To4(); ipv4 != nil {
...
```

agent/sshd/utmp.go
```diff
@@ -0,0 +1,267 @@
...
+
+	u.Type = UserProcess
+	a := unix.Timeval{}
+	unix.Gettimeofday(&a)
+	u.Tv.Sec, u.Tv.Usec = a.Sec, a.Usec
+	u.Pid = int32(os.Getpid())
+
+	// remoteAddr has the form <IPv4 address>:<port>
+	// or [<IPv6 address>]:port
+
```
otavio, line 67: add `//`

otavio (changes requested):

agent/sshd/server.go
```diff
@@ -122,13 +123,21 @@ func (s *Server) sessionHandler(session sshserver.Session) {
...
 		logrus.WithFields(logrus.Fields{
 			"user":       session.User(),
 			"pty":        pts.Name(),
-			"remoteaddr": session.RemoteAddr().String(),
+			"remoteaddr": remoteAddr,
 			"localaddr":  session.LocalAddr().String(),
 		}).Info("Session started")
 
+		ut := utmpStartSession(
+			strings.TrimPrefix(pts.Name(),"/dev/"),
```
otavio, line 136: I belive the trimming of the prefix is an implementation detail and belongs to the `utmpStartSession(...)` code. So here we pass the `pty` and be done with that.

agent/sshd/server.go
```diff
@@ -122,13 +123,21 @@ func (s *Server) sessionHandler(session sshserver.Session) {
...
 		os.Chown(pts.Name(), uid, -1)
 
+		remoteAddr := session.RemoteAddr().String()
+
 		logrus.WithFields(logrus.Fields{
 			"user":       session.User(),
 			"pty":        pts.Name(),
-			"remoteaddr": session.RemoteAddr().String(),
+			"remoteaddr": remoteAddr,
 			"localaddr":  session.LocalAddr().String(),
```
otavio, line 132: please check if it really needs to use `.String()` here and on `remoteAddr` as I belive it will call it internally.

otavio (changes requested):

agent/sshd/utmp.go
sixhills: Done.

agent/sshd/utmp.go
sixhills: Done.

agent/sshd/utmp.go
sixhills: Agreed - more elegant.

agent/sshd/utmp.go
sixhills: Added <>

agent/sshd/server.go
sixhills: Done.

otavio: I think you did not push the changes. I assumed it'd be replaced by `pts.Name()` and inside the function the prefix would be trimmed.

agent/sshd/server.go
sixhills: I'll check but `.String()`  was in the original code,so I left it.

agent/sshd/server.go
sixhills: Now done and pushed.

agent/sshd/server.go
sixhills: I've confirmed that `.String()` isn't needed in the calls of logrus.WithFields, so I've deleted it.

sixhills:
When the latest changes have been reviewed, should I squash all the commits into one (retaining the commit log)?

otavio:
> When the latest changes have been reviewed, should I squash all the commits into one (retaining the commit log)?

Yes, except the `.String()` removal as it is an isolated change. Thanks for all the patience and good work.

gateway/kickstart.sh
```diff
@@ -46,6 +46,8 @@ $SUDO docker run -d \
        -v /var/run/docker.sock:/var/run/docker.sock \
        -v /etc/passwd:/etc/passwd \
        -v /etc/group:/etc/group \
+       -v /var/run:/var/run \
```
gustavosbarreto, line 49: Older versions will also have to add this volume to the new container when doing the auto-update. See https://github.com/shellhub-io/shellhub/blob/master/agent/selfupdater/updater_docker.go#L83

E.g.
```
if v.LessThan(v0_4_2) {
  parent.info.HostConfig.Mounts = []mount.Mount{
    mount.Mount{
      Type:   mount.TypeBind,
      Source: "/var/run",
      Target: "/var/run",
    },
    mount.Mount{
      Type:   mount.TypeBind,
      Source: "/var/log",
      Target: "/var/log",
    },
  }
}
```

Or you can set `UtmpxFile  = "/host/var/run/utmp"` and `WtmpxFile  = "/host/var/log/wtmp"` when building using `docker` tag

gustavosbarreto (changes requested):

gateway/kickstart.sh
otavio: Good catch @gustavosbarreto; I completely forgot about this :-)

I have mixed feelings about the approach to take. We'd need to handle this on `kickstart.sh` on the gateway as well.

It seems adding the fine inside the `/host` volume is less hackish. We can later come up with a generic way to map to the host files to abstract this.

sixhills: I don't know how to set the file name constants differently when building using docker.  Guidance appreciated!

gustavosbarreto: @sixhills You can create a file for each platform type (docker and native) to set different values for `UtmpxFile` and `WtmpxFile`.

See:
https://github.com/shellhub-io/shellhub/blob/master/agent/init_docker.go
https://github.com/shellhub-io/shellhub/blob/master/agent/init_native.go

gustavosbarreto: `utmp_docker.go`:

```
// +build docker

const (
	UtmpxFile   = "/host/var/run/utmp"
	WtmpxFile   = "/host/var/log/wtmp"
)
```

`utmp_native.go`:

```
// +build !docker

const (
	UtmpxFile   = "/var/run/utmp"
	WtmpxFile   = "/var/log/wtmp"
)
```

sixhills: Thanks - I understand now,

sixhills: I don't think this will work - unless I'm misunderstanding, which is quite possible.

Despite using `-v /:/host` in the `docker run` command, the agent inside docker reports:

`level=warning msg="Open failed" err="open /host/var/run/utmp: no such file or directory" file=/host/var/run/utmp`

Does `docker  run -v ...` mount only the specified directory and not the tree beneath it?

otavio: @sixhills what host distribution are you using?

otavio: ... and please post the output of:
```sh
mount
```

sixhills: Raspberry Pi OS (aka Raspbian).  I can try others if there may be differences.  Mount output follows:

dev/mmcblk0p2 on / type ext4 (rw,noatime)
devtmpfs on /dev type devtmpfs (rw,relatime,size=1827800k,nr_inodes=97773,mode=755)
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
proc on /proc type proc (rw,relatime)
securityfs on /sys/kernel/security type securityfs (rw,nosuid,nodev,noexec,relatime)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000)
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
tmpfs on /run/lock type tmpfs (rw,nosuid,nodev,noexec,relatime,size=5120k)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
cgroup2 on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate)
cgroup on /sys/fs/cgroup/systemd type cgroup (rw,nosuid,nodev,noexec,relatime,xattr,name=systemd)
none on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)
cgroup on /sys/fs/cgroup/net_cls,net_prio type cgroup (rw,nosuid,nodev,noexec,relatime,net_cls,net_prio)
cgroup on /sys/fs/cgroup/blkio type cgroup (rw,nosuid,nodev,noexec,relatime,blkio)
cgroup on /sys/fs/cgroup/freezer type cgroup (rw,nosuid,nodev,noexec,relatime,freezer)
cgroup on /sys/fs/cgroup/devices type cgroup (rw,nosuid,nodev,noexec,relatime,devices)
cgroup on /sys/fs/cgroup/perf_event type cgroup (rw,nosuid,nodev,noexec,relatime,perf_event)
cgroup on /sys/fs/cgroup/pids type cgroup (rw,nosuid,nodev,noexec,relatime,pids)
cgroup on /sys/fs/cgroup/cpuset type cgroup (rw,nosuid,nodev,noexec,relatime,cpuset)
systemd-1 on /proc/sys/fs/binfmt_misc type autofs (rw,relatime,fd=28,pgrp=1,timeout=0,minproto=5,maxproto=5,direct)
sunrpc on /run/rpc_pipefs type rpc_pipefs (rw,relatime)
debugfs on /sys/kernel/debug type debugfs (rw,relatime)
mqueue on /dev/mqueue type mqueue (rw,relatime)
configfs on /sys/kernel/config type configfs (rw,relatime)
/dev/mmcblk0p1 on /boot type vfat (rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=391976k,mode=700,uid=1000,gid=1000)
gvfsd-fuse on /run/user/1000/gvfs type fuse.gvfsd-fuse (rw,nosuid,nodev,relatime,user_id=1000,group_id=1000)
fusectl on /sys/fs/fuse/connections type fusectl (rw,relatime)
nsfs on /run/docker/netns/default type nsfs (rw)

gustavosbarreto: @sixhills What is the output of `readlink -f /var/run/utmp`?

sixhills: That's a good question.  `readlink -f /var/run/utmp` reports: `/run/utmp`.

sixhills: Sorry this is proving rather more complicated than expected. I don't want to use too much of your time on this.

otavio: Ok; so we won't be capable of handling it using the `/host` volume. You'll need to add `-v /var/run/utmp` on upgrade and installation script :-(

sixhills: OK, thanks.

otavio: You need to handle the upgrade case as well. See https://github.com/shellhub-io/shellhub/pull/582/commits/505733e6843ed619f92f18cef2c0f177caf222c3#r551440473

gustavosbarreto:
@sixhills Thank you!!

PR #594, opened 2020-12-30
agent: Small code rework

otavio:
This PR is all about code rework for cleanness so it is not intended to add new features or do massive reworks, just small ones.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -15,6 +15,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// Agent version to be embed inside the binary. This is injected using `-ldflags`
+// build option (e.g: `go build -ldflags "-X main.AgentVersion=1.2.3"`).
+//
+// If set to `latest`, the auto-updating mechanism is disabled. This is intended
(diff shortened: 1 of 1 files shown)


otavio:
@sixhills please take a look here as well.

sixhills:
Excellent! Comments help to understand the way the code works.

agent/main.go
```diff
@@ -102,12 +122,22 @@ func main() {
 				continue
 			}
 
+			namespace := agent.authData.Namespace
+			tenantName := agent.authData.Name
+			sshEndpoint := agent.serverInfo.Endpoints.SSH
+
+			sshid := strings.NewReplacer(
```
gustavosbarreto, line 129: Why not? 
```fmt.Sprintf("%s.%s@%s", namespace, tenantName, strings.Split(sshEndpoint, ":")[0])```

gustavosbarreto (approved):

agent/main.go
otavio: Even nicer. :-)

PR #597, opened 2021-01-03
Remove CGO dependency in agent

### LINKED ISSUE #596: Possibility of removing CGO usage from the agent
The osauth package in the agent uses CGO. This blocks cross compilation using GOOS and GOARCH and makes it much harder to do at all.

As the package does mainly parse files and perform some crypto, can it not be replaced with a pure go implementation ? Or was there a specific reason why this was implemented in CGO ?

If you consider it possible to replace it I would be willing to try and submit a PR :-)

Greetings

s00500:
Closes #596 

CGO is really not necessary for the simple file parsing
github.com/GehirnInc/crypt can provide the needed crypto functions

Removing the CGO dependency for the agent makes cross compilation much easier, and can enable easier workflows when the agent is not running inside of docker

Currently I have left a big comment that parses additional parts of /etc/shadow
In my opinion it is unnecessary, but it would make sense to be in there for completeness or later use... So please let me know

Also my change to the User type, changing UID and GUID to ints is not strictly necessary and therefore up for discussion

I did test the new implementation on a Pi running RaspberryPiOS lite and it works fine

Let me know what you think =D

### DIFF
--- a/agent/pkg/osauth/auth.go
+++ b/agent/pkg/osauth/auth.go
@@ -1,55 +1,117 @@
 package osauth
 
-/*
-#cgo LDFLAGS: -lcrypt
-#define _GNU_SOURCE 1
-#include <stdlib.h>
-#include <stdio.h>
-#include <shadow.h>
-#include <string.h>
-#include <pwd.h>
-#include <crypt.h>
-*/
-import "C"
-
 import (
-	"unsafe"
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/GehirnInc/crypt"
+	_ "github.com/GehirnInc/crypt/md5_crypt"    // GehirnInc/crypt uses blank imports for crypto subpackages
+	_ "github.com/GehirnInc/crypt/sha256_crypt" // GehirnInc/crypt uses blank imports for crypto subpackages
+	_ "github.com/GehirnInc/crypt/sha512_crypt" // GehirnInc/crypt uses blank imports for crypto subpackages
+	"github.com/sirupsen/logrus"
 )
 
 var DefaultShadowFilename = "/etc/shadow"
 
-func AuthUser(user, passwd string) bool {
-	cuser := C.CString(user)
-	defer C.free(unsafe.Pointer(cuser))
+type ShadowEntry struct {
+	Username    string // Login name
+	Password    string // Hashed password
+	Lastchanged int    // Days since Jan 1, 1970 that password was last changed
+	Minimum     int    // The minimum number of days required between password changes i.e. the number of days left before the user is allowed to change his/her password
+	Maximum     int    // The maximum number of days the password is valid (after that user is forced to change his/her password)
+	Warn        int    // The number of days before password is to expire that user is warned that his/her password must be changed
+	Inactive    int    // The number of days after password expires that account is disabled
+	Expire      int    // Days since Jan 1, 1970 that account is disabled i.e. an absolute date specifying when the login may no longer be used.
+}
 
-	cpasswd := C.CString(passwd)
-	defer C.free(unsafe.Pointer(cpasswd))
-
-	cfilename := C.CString(DefaultShadowFilename)
-	defer C.free(unsafe.Pointer(cfilename))
-
-	cmode := C.CString("r")
-	defer C.free(unsafe.Pointer(cmode))
-
-	f := C.fopen(cfilename, cmode)
-	defer C.fclose(f)
-
-	var pwd *C.struct_spwd
-	for {
-		if pwd = C.fgetspent(f); pwd == nil {
-			return false
-		}
-
-		if C.strcmp(cuser, pwd.sp_namp) == 0 {
-			break
-		}
-	}
-
-	if pwd == nil {
+func AuthUser(username, passwd string) bool {
+	shadowFile, err := os.Open(DefaultShadowFilename)
+	if err != nil {
+		// TODO: log error
+		logrus.Error("Could not open /etc/shadow")
 		return false
 	}
 
-	crypted := C.crypt(cpasswd, pwd.sp_pwdp)
+	defer shadowFile.Close()
 
-	return C.strcmp(crypted, pwd.sp_pwdp) == 0
+	entries, err := parseShadowReader(shadowFile)
+	if err != nil {
+		logrus.Printf("Could not parse shadowfile %v", err)
+		return false
+	}
+
+	if entry, ok := entries[username]; ok {
+		if entry.Password != "" {
+			crypt := crypt.NewFromHash(entry.Password)
+			if crypt == nil {
+				logrus.Error("Could not detect password crypto algorithm from shadow entry")
+				return false
+			}
+
+			err := crypt.Verify(entry.Password, []byte(passwd))
+			return err == nil
+		}
+		logrus.Error("Password entry is empty")
+		return false
+	}
+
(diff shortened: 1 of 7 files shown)


agent/pkg/osauth/auth.go
```diff
@@ -1,55 +1,145 @@
...
+		return result, fmt.Errorf("Shadow line had wrong number of parts %d != 9", len(parts))
 	}
+	result.Username = strings.TrimSpace(parts[0])
+	result.Password = strings.TrimSpace(parts[1])
+	/*
+	   	lastchanged, err := strconv.Atoi(parts[2])
+	   	if err != nil {
+	   		return result, fmt.Errorf("Shadow line had badly formatted lastchanged %s", parts[2])
+	   	}
+	   	result.Lastchanged = lastchanged
```
otavio, line 101: I think we ought to include this code so it becomes fully featured. The hassle of adding it back will be bigger than adding it now.

otavio (changes requested):

agent/pkg/osauth/auth.go
s00500: Ok, added it back in, also made it a little nicer

otavio: Cool, could you rebase the patches and squash this last one onto the original one? There is no reason for it to be done in a specific commit.

s00500: ok, I can also squash the entire PR if that makes more sense ?

s00500:
Can you maybe retrigger the CI run ? I am really not sure why this one failed...

s00500:
Seems like a download failed in building the image

agent/pkg/osauth/auth.go
otavio: No; small and logical commits are nicer for repo history.

otavio:
> Can you maybe retrigger the CI run ? I am really not sure why this one failed...

Just did it.

agent/sshd/cmd.go
```diff
@@ -25,7 +22,7 @@ func newCmd(u *osauth.User, shell, term, host string, command ...string) *exec.C
 		groups = append(groups, uint32(igid))
 	}
 	if len(groups) == 0 {
-		groups = append(groups, uint32(gid))
+		groups = append(groups, uint32(u.GID))
```
otavio, line 25: I believe this commit ought to be squashed on the previous commit otherwise it will fail to build.

agent/sshd/server.go
```diff
@@ -118,9 +116,7 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 
 		u := osauth.LookupUser(session.User())
 
-		uid, _ := strconv.Atoi(u.UID)
-
-		os.Chown(pts.Name(), uid, -1)
+		os.Chown(pts.Name(), u.UID, -1)
```
otavio, line 119: So does this commit.

agent/sshd/cmd_docker.go
```diff
@@ -12,10 +12,7 @@ import (
 )
 
 func newCmd(u *osauth.User, shell, term, host string, command ...string) *exec.Cmd {
-	uid, _ := strconv.Atoi(u.UID)
-	gid, _ := strconv.Atoi(u.GID)
-
-	nscommand, _ := nsenterCommandWrapper(uid, gid, fmt.Sprintf("/host/%s", u.HomeDir), command...)
+	nscommand, _ := nsenterCommandWrapper(u.UID, u.GID, fmt.Sprintf("/host/%s", u.HomeDir), command...)
```
otavio, line 15: here as well.

s00500:
True these individual commits can fail, I squashed them now

But that should not be the reason for the ci failure, that also seems to be working again now... =D

agent/sshd/cmd_docker.go
otavio: Doesn't this commit also belong to the previous one?

s00500: squashed =D

otavio (approved):

agent/sshd/cmd.go
```diff
@@ -37,6 +34,6 @@ func newCmd(u *osauth.User, shell, term, host string, command ...string) *exec.C
 	}
 	cmd.Dir = u.HomeDir
 	cmd.SysProcAttr = &syscall.SysProcAttr{}
-	cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uint32(uid), Gid: uint32(gid), Groups: groups}
+	cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uint32(u.UID), Gid: uint32(u.GID), Groups: groups}
```
otavio, line 37: Couldn't this use `u.UID` and `u.GID` directly?

agent/sshd/cmd.go
otavio, line 25: Is the cast still required?

agent/sshd/cmd.go
s00500: Yes it is, because the fields in the user struct use ints. I thought of changing that but it requieres casts in other places back to int as the os package wants ints, whereas the syscall package wants uint32

Either way there will be casts, let me know if you consider the other way around (casting when parsing) cleaner

agent/sshd/cmd.go
s00500: same applies as in previous comment above

agent/sshd/cmd.go
otavio: If the syscall requires it to be an `uint32` we could do the cast where we gather the information. So we do it once and avoid redoing it all over the place.

agent/sshd/server.go
```diff
@@ -9,19 +9,17 @@ import (
 	"sync"
 	"time"
 
-	sshserver "github.com/gliderlabs/ssh"
-	"github.com/shellhub-io/shellhub/agent/pkg/osauth"
-	"github.com/sirupsen/logrus"
-)
-import (
 	"crypto"
```
otavio, line 12: This belongs to the original commit. Not this rework patch right?

You could squash this as well, if preferred.

s00500: done

agent/pkg/osauth/user.go
```diff
@@ -26,14 +26,14 @@ type User struct {
 func LookupUser(username string) *User {
 	passwdFile, err := os.Open(DefaultPasswdFilename)
 	if err != nil {
-		logrus.Error("Could not open %s", DefaultPasswdFilename)
+		logrus.Errorf("Could not open %s", DefaultPasswdFilename)
```
otavio, line 29: This is changing the code included in the previous commit so we should squash this.

s00500: Is there any particular reason not to squash the entire PR ?

otavio: No, please squash it all.

s00500:
Weird CI failure again...

otavio:
@s00500 the PR is good to go, but we'll merge it once we do the 0.5.0 release. We'd like to keep it in use for some time to reduce the risk of regressions.

otavio (approved):

s00500:
I see, sounds good,  thanks 👍

gustavosbarreto:
@s00500 Can you please resolve conflicts?

s00500:
@gustavosbarreto fixed

otavio (approved):

gustavosbarreto:
@all-contributors please add @s00500 for code

PR #613, opened 2021-01-11
ui: Display only 10 characters in the namespace bar

leonardojoao:


ui/src/components/app_bar/namespace/NamespaceMenu.vue
```diff
@@ -248,6 +248,13 @@ export default {
         this.$store.dispatch('snackbar/showSnackbarErrorLoading', this.$errors.namespaceSwitch);
       }
     },
+
+    displayOnlyTenCharacters(string) {
+      if (string !== undefined) {
+        if (string.length > 10) return `${string.substring(0, 10)}...`;
```
otavio, line 254: `.substr(0, 10);` doesn't need to check the size.

PR #615, opened 2021-01-11
ui: Reshape the profile configuration layout

leonardojoao:


noreng-jg (commented):
The new layout looks cool now.
I think that we could also remove the form errors when the cancel button
is pressed, and return the component variables to their original state.

The following method, seems to do the job:

```js
   cancel() {
      if (this.editDataStatus) {
        this.setData();
        this.$refs.data.reset();
        this.editDataStatus = !this.editDataStatus;
      } else if (this.editPasswordStatus) {
        this.currentPassword = '';
        this.newPassword = '';
        this.newPasswordConfirm = '';
        this.$refs.pass.reset();
        this.editPasswordStatus = !this.editPasswordStatus;
      }
  }
```
And change the template:
```vue
                  <v-btn
                    class="mr-2"
                    outlined
                    @click="cancel"
                  >
                    Cancel
                  </v-btn>

```

PR #617, opened 2021-01-12
single user login mode

shawn111:
Not all systems run docker, and it is dangerous to run as root in native mode.
This patch provide a simple mode, normal user can run agent.

We only allow the account to login into this system.
It is useful for most cases and can keep system safer.
The other point is once the change apply native mode might have more users.
I also provide some usage hint in main.go.

Example:

setup
  user$ SIMPLE_PASSWORD=123 ./agent
connection
  $ ssh user@sshid_hostname@server
  user@sshid_hostname@server's password: 123


shawn111:
https://github.com/shellhub-io/shellhub/pull/622 replace this one

Issue #618, opened 2021-01-13
[Enhancement] agent run as non-root mode


shawn111:
Use docker to deploy is easy, but it will be better if support native mode.
However, in native mode root permission is dangerous.

PR #622, opened 2021-01-13
agent: support non-root to run

shawn111:
ShellHub run as single user mode.

Since in the mode, not allow to be other accounts.
We don't need root permission to switch users.

Closes: #618

```
$ SHELLHUB_SIMPLE_PASSWORD=$(openssl passwd -1) ./agent
Password: 
Verifying - Password: 
Verify failure
This application is configured via the environment. The following environment
variables can be used:

KEY                            TYPE       DEFAULT    REQUIRED    DESCRIPTION
SHELLHUB_SERVER_ADDRESS        String                true        
SHELLHUB_PRIVATE_KEY           String                true        
SHELLHUB_TENANT_ID             String                true        
SHELLHUB_KEEPALIVE_INTERVAL    Integer    30                     
SHELLHUB_PREFERRED_HOSTNAME    String                            
SHELLHUB_SIMPLE_PASSWORD       String                            SHELLHUB_SIMPLE_PASSWORD=$(openssl passwd -1) ./agent
FATA[0000] required key SERVER_ADDRESS missing value   
```

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -44,14 +44,13 @@ type ConfigOptions struct {
 	// Set the device preferred hostname. This provides a hint to the server to
 	// use this as hostname if it is available.
 	PreferredHostname string `envconfig:"preferred_hostname"`
+
+	// Force act as SimplePassword mode. The hash password could be generated
+	// by ```openssl passwd```.
+	SimplePassword string `envconfig:"simple_password"`
 }
 
 func main() {
-	if os.Geteuid() != 0 {
-		logrus.Error("ShellHub must be run as root")
-		os.Exit(1)
-	}
-
 	opts := ConfigOptions{}
 
 	// Process unprefixed env vars for backward compatibility
@@ -61,6 +60,14 @@ func main() {
 		// show envconfig usage help users to run agent
 		envconfig.Usage("shellhub", &opts) // nolint:errcheck
 		logrus.Fatal(err)
+	}
+
+	if (os.Geteuid() == 0) != (opts.SimplePassword == "") {
+		logrus.Error("ShellHub agent must be run as root w/o SIMPLE_PASSWORD")
+		logrus.Error("or as normal user with SIMPLE_PASSWORD")
+		logrus.Error("example:")
+		logrus.Error(" SIMPLE_PASSWORD=$(openssl passwd -1) ./agent")
+		os.Exit(1)
 	}
 
 	updater, err := selfupdater.NewUpdater(AgentVersion)
@@ -96,6 +103,10 @@ func main() {
 	}
 
 	sshserver := sshd.NewServer(agent.cli, agent.authData, opts.PrivateKey, opts.KeepAliveInterval)
+
+	if opts.SimplePassword != "" {
+		sshserver.ForceSingleMode(opts.SimplePassword)
+	}
 
 	tunnel := NewTunnel()
 	tunnel.connHandler = func(w http.ResponseWriter, r *http.Request) {
--- a/agent/pkg/osauth/auth.go
+++ b/agent/pkg/osauth/auth.go
@@ -28,6 +28,22 @@ type ShadowEntry struct {
 	Expire      int    // Days since Jan 1, 1970 that account is disabled i.e. an absolute date specifying when the login may no longer be used.
 }
 
+func VerifyPasswordHash(hashPassword, passwd string) bool {
+	if hashPassword == "" {
+		logrus.Error("Password entry is empty")
+		return false
+	}
+
+	crypt := crypt.NewFromHash(hashPassword)
+	if crypt == nil {
+		logrus.Error("Could not detect password crypto algorithm from shadow entry")
+		return false
+	}
+
+	err := crypt.Verify(hashPassword, []byte(passwd))
+	return err == nil
+}
+
 func AuthUser(username, passwd string) bool {
 	shadowFile, err := os.Open(DefaultShadowFilename)
 	if err != nil {
@@ -45,18 +61,7 @@ func AuthUser(username, passwd string) bool {
 	}
 
 	if entry, ok := entries[username]; ok {
-		if entry.Password != "" {
-			crypt := crypt.NewFromHash(entry.Password)
-			if crypt == nil {
-				logrus.Error("Could not detect password crypto algorithm from shadow entry")
-				return false
-			}
-
-			err := crypt.Verify(entry.Password, []byte(passwd))
-			return err == nil
-		}
-		logrus.Error("Password entry is empty")
-		return false
+		return VerifyPasswordHash(entry.Password, passwd)
 	}
 
 	logrus.Warn("User not found")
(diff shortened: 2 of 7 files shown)


agent/sshd/server.go
```diff
@@ -78,7 +78,6 @@ func NewServer(api client.Client, authData *models.DeviceAuthResponse, privateKe
 			return &sshConn{conn, closeCallback, ctx}
 		},
 	}
-
```
otavio, line 81: Unrelated change, please undo it.

agent/main.go
```diff
@@ -51,9 +52,10 @@ type Information struct {
 
 func main() {
 	if os.Geteuid() != 0 {
-		logrus.Error("ShellHub must be run as root")
-		os.Exit(1)
+		logrus.Info("ShellHub run as single user mode.")
+		osauth.DefaultShadowFilename = ".shadow"
```
otavio, line 56: What is the use case for this file?

otavio (changes requested):
Please improve the commit log. The message you did for the PR is nice, and it'd provide a good base for it.

agent/sshd/server.go
shawn111: @otavio My patch is based on master branch, do you mean I need change the base branch to commit.
If that, what branch should it base on?

shawn111: thanks, removed it.

agent/main.go
shawn111: To explain and help users to use, I created a script shadow_helper.sh.

```
# This script is a helper to generate shadow file.
#
# shellhub agent support to define shadow path.
# The passwords of system and shellhub could be different.
#
# In single user mode,
# it allows different accounts to login but same system account.
# You can review the session log in web console.
#
#
# example:
#  $ USER=abc PASSWD=123 SHADOW_FILEANAME=.shadow ./shadow_helper.sh
#
#  $ SHADOW_FILEANAME=.shadow ./agent


```

agent/shadow_helper.sh
```diff
@@ -0,0 +1,55 @@
...
+   exit 1
+fi
+
+if [ -z ${SHADOW_FILEANAME+x} ]; then
+   echo "SHADOW_FILEANAME is unset"
+   exit 1
+fi
+
+ENCRYPTED_PASSWD=$(echo "$PASSWD" | mkpasswd --method=SHA-512 --stdin)
+echo $USER:$ENCRYPTED_PASSWD::::::: >> ${SHADOW_FILEANAME}
```
otavio, line 49: The shadow file generation could be done using a sub-command inside `shellhub-agent` itself.  This allows the deployment of a single binary. 

For example:

```
./agent create-shadow-file <username> -o <destination file>
```

and it could read the password from stdin, for example. @gustavosbarreto what do you think?

shawn111: It is a good idea.

If https://github.com/shellhub-io/shellhub/pull/597 be merged, that would be easy to implement.

shawn111: @otavio @gustavosbarreto it is good to have sub-commands inside shellhub-agent.

Do you think let put commands/sub-commands related code into pkg/cmd/<sub-commands>.go?
Or even consider to import cobra.

shawn111:
@otavio Could you help me do the review again?
Since 0.5.0 release, I did rebase

otavio:
There are some checks which failed. Could you look at this?

shawn111:
Oh...got it, let me fix it. thanks

shawn111:
@otavio it's all pass. :)

otavio:
Hello @shawn111, I hope you are doing well.

In our last team meeting, we discussed this PR, and we'd like to understand this reasoning. We've been trying to provide a consistent user experience when using the ShellHub agent inside Docker and as a standalone process. This PR seems to conflict with this consistency.

So we'd need to understand:
1. what use case your trying to cover?
2. what is your current way of using the agent? Is it in Docker or as a standalone process?

shawn111:
@otavio,
1. In some cases, we don't have docker. Like in Raspberry pi zero (ARMv6), or Fedora/CentOS or MacOS don't install docker by default. Shellhub is useful, I thought docker mode as default mode is good for most users but don't need to must have docker.

2. standalone Process and don't need root permission. Read provided shadow file instead of system shadow file. Although, that means we just can login as the same system user even we provide different login names.

Once, this mode could work, we can also provide agent binaries directly, not just docker image. More people have chances to use it.

@otavio I know non-root mode might confuse new users. Maybe just let this mode for developers or test mode.

otavio:
It might make sense for simple use, but in this case, we could use an environment variable to set the password. This would be good for development tests and helps in your desired use case. However, for use outside of Docker it is already supported. People can install it as a direct daemon.

We are working on adding packages for some distros and it'd solve your second use case.

shawn111:
@otavio got it, let's use environment variable to set the password. it is much simple.
I'll update it.

otavio (changes requested):

agent/pkg/osauth/auth.go
```diff
@@ -28,35 +28,33 @@ type ShadowEntry struct {
...
-	defer shadowFile.Close()
+	err := crypt.Verify(hashPassword, []byte(passwd))
+	return err == nil
+}
 
+func AuthUser(username, passwd string) bool {
+        shadowFile, err := os.Open(DefaultShadowFilename)
+        if err != nil {
+		return false
+	}
```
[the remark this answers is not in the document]
shawn111: Added

agent/pkg/osauth/auth.go
```diff
@@ -28,35 +28,33 @@ type ShadowEntry struct {
...
-func AuthUser(username, passwd string) bool {
-	shadowFile, err := os.Open(DefaultShadowFilename)
-	if err != nil {
-		// TODO: log error
-		logrus.Error("Could not open /etc/shadow")
+func PasswdVerify(hashPassword, passwd string) bool {
+	if hashPassword == "" {
+		logrus.Error("Password entry is empty")
+		return false
+        }
```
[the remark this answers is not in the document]
shawn111: go fmt, fixed.

agent/pkg/osauth/auth.go
```diff
@@ -28,35 +28,33 @@ type ShadowEntry struct {
 	Expire      int    // Days since Jan 1, 1970 that account is disabled i.e. an absolute date specifying when the login may no longer be used.
 }
 
-func AuthUser(username, passwd string) bool {
-	shadowFile, err := os.Open(DefaultShadowFilename)
-	if err != nil {
-		// TODO: log error
-		logrus.Error("Could not open /etc/shadow")
+func PasswdVerify(hashPassword, passwd string) bool {
```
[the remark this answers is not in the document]
shawn111: agent/sshd/server.go use it, change it to VerifyPasswordHash

gustavosbarreto:
@shawn111 can you please resolve conflicts so we can merge?

shawn111:
@gustavosbarreto  the conflicts solved, thanks.

otavio (approved):

otavio:
@gustavosbarreto please take a look at this; I think it is ready for merge.

PR #671, opened 2021-01-28
env: rename SHELLHUB_HOSTED to SHELLHUB_ENTERPRISE

gustavosbarreto:


otavio (approved):
This is indeed nicer :-)

PR #681, opened 2021-02-01
api: add delete user in store

eduardoveiga:


api/main.go
```diff
@@ -91,7 +91,7 @@ func main() {
 	internalAPI.POST(routes.RecordSessionURL, apicontext.Handler(routes.RecordSession))
 	publicAPI.GET(routes.PlaySessionURL, apicontext.Handler(routes.PlaySession))
 
-publicAPI.GET(routes.GetStatsURL,
+	publicAPI.GET(routes.GetStatsURL,
```
gustavosbarreto, line 94: This route should be handled by `admin-api`

PR #703, opened 2021-02-09
api:  Blocks actions when the user does not own the namespace

leonardojoao:


api/deviceadm/service.go
```diff
@@ -52,15 +52,43 @@ func (s *service) GetDevice(ctx context.Context, uid models.UID) (*models.Device
...
+	}
+
+	namespace, err := s.store.GetNamespace(ctx, tenant)
+	if err != nil {
+		return err
+	}
+
+	if user.ID != namespace.Owner {
+		return ErrUnauthorized
+	}
```
otavio, line 68: This is being repeated, so it's better to rework this to be a generic function called multiple times. Something like: `IsNamespaceOnwer(tenant, username) bool` or similar.

otavio (changes requested):

PR #704, opened 2021-02-09
Make property filter generic

gustavosbarreto:
Instead of having a filter property for each type of data, uses a generic property filter which convert the data according to the type.

Also rework ListUsers from mongostore to use generic filter property


gustavosbarreto:
@leonardojoao @eduardoveiga This PR changes `Store` interface which may break `cloud-api`

PR #708, opened 2021-02-10
api: fix partial delete

eduardoveiga:
when a user is deleted,  your namespaces need to be deleted too.  And when a namespaces is deleted, your devices need to be deleted too

### DIFF
--- a/api/store/mongo/store.go
+++ b/api/store/mongo/store.go
@@ -924,6 +924,23 @@ func (s *Store) UpdateUserFromAdmin(ctx context.Context, username, email, passwo
 func (s *Store) DeleteUser(ctx context.Context, ID string) error {
 	objID, _ := primitive.ObjectIDFromHex(ID)
 	_, err := s.db.Collection("users").DeleteOne(ctx, bson.M{"_id": objID})
+
+	findOptions := options.Find()
+	cursor, err := s.db.Collection("namespaces").Find(ctx, bson.M{"owner": ID}, findOptions)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


api/store/mongo/store.go
```diff
@@ -1340,9 +1358,43 @@ func (s *Store) CreateNamespace(ctx context.Context, namespace *models.Namespace
 	return namespace, err
 }
 func (s *Store) DeleteNamespace(ctx context.Context, namespace string) error {
```
gustavosbarreto, line 1360: You can use `DeleteMany` to removes all documents that match the filter from a collection instead of iterating through a cursor for a performance reasons

gustavosbarreto (changes requested):

api/store/mongo/store.go
```diff
@@ -1362,8 +1379,18 @@ func (s *Store) CreateNamespace(ctx context.Context, namespace *models.Namespace
...
-	_, err := s.db.Collection("namespaces").DeleteOne(ctx, bson.M{"tenant_id": namespace})
-	return err
+	if _, err := s.db.Collection("namespaces").DeleteOne(ctx, bson.M{"tenant_id": namespace}); err != nil {
+		return err
+	}
+
+	collections := [6]string{"devices", "sessions", "connected_devices", "firewall_rules", "public_keys", "recorded_sessions"}
```
gustavosbarreto, line 1386: It would be better to use slices instead of a literal array

gustavosbarreto (changes requested):

PR #723, opened 2021-02-17
topic: refactor queries to mongo 4.2

eduardoveiga:


wip
```diff
@@ -0,0 +1 @@
+wip
```
gustavosbarreto, line 1: Is this file really necessary?

gustavosbarreto (changes requested):

PR #724, opened 2021-02-17
api: Add support for API token

Asakiz:


api/routes/webtoken.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/shellhub-io/shellhub/api/apicontext"
+	"github.com/shellhub-io/shellhub/api/nsadm"
+	"golang.org/x/crypto/bcrypt"
+)
+
+const (
+	GetTokenListURL = "/token"
+)
+
+func GenerateToken(c apicontext.Context) string {
```
otavio, line 18: This need to handle errors and return it in case of failure.

api/routes/webtoken.go
```diff
@@ -0,0 +1,41 @@
...
+	hash, err := bcrypt.GenerateFromPassword([]byte(c.Tenant()), bcrypt.DefaultCost)
+	if err != nil {
+		log.Fatal(err)
```
otavio, line 21: Here we ought to return the error.

api/nsadm/service.go
```diff
@@ -173,3 +175,15 @@ func (s *service) GetDataUserSecurity(ctx context.Context, tenant string) (bool,
 	}
 	return false, ErrUnauthorized
 }
+
+func (s *service) ListTokens(ctx context.Context, namespace string) (*models.Token, error) {
+	ns, _ := s.store.GetNamespace(ctx, namespace)
+	if ns != nil {
```
otavio, line 181: Invest the logic here. This reduces the code identation and allow for easier reading of code. Something like:
```go
if ns == nil {
    return nil, ErrTokenNotFound
}
```

So rest of the code can take the assumption `ns` is valid.

otavio (changes requested):

api/nsadm/service.go
```diff
@@ -173,3 +175,17 @@ func (s *service) GetDataUserSecurity(ctx context.Context, tenant string) (bool,
...
+		tokens := []models.Token{}
+		for _, token := range ns.Token.(primitive.A) {
+			tokens = append(tokens, token.(models.Token))
+		}
+		return tokens, nil
+	}
+	return []models.Token{}, ErrTokenNotFound
+}
```
otavio, line 189: Please, use the _early return_ pattern for any new code. We'll later rework existing code to avoid this wrong code pattern to make the code worse.

This could be reworked:
```go
func (s *service) ListTokens(ctx context.Context, namespace string) ([]models.Token, error) {
	ns, err := s.store.GetNamespace(ctx, namespace)
	if err != nil {
		return nil, ErrTokenNotFound
	}

	tokens := []models.Token{}
	for _, token := range ns.Token.(primitive.A) {
		tokens = append(tokens, token.(models.Token))
	}

	return tokens, nil
}
```

otavio: Also, @gustavosbarreto mentioned we ought to avoid using MongoDB's primitives here. We ought to use it from the store API. So please take a look at this code so we start correctly.

api/routes/webtoken.go
otavio: This doesn't fix the comment. We ought to use `(string, error)` as the return type.

api/routes/token.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+
+	namespace := ""
+	if v := c.Username(); v != nil {
+		namespace = v.Username
+	}
+
+	if _, err := svc.CreateToken(c.Ctx(), namespace); err != nil {
+		if err == token.ErrUnauthorized {
```
otavio, line 49: To mach here, as well as other `err` special cases, it is better to use `switch` as it allows for same construction for one or more special cases and the explicit default one.

api/routes/token.go
```diff
@@ -0,0 +1,118 @@
...
+func DeleteToken(c apicontext.Context) error {
+	svc := token.NewService(c.Store())
+
+	namespace := ""
+	if v := c.Namespace(); v != nil {
+		namespace = v.Namespace
+	}
+
+	if err := svc.DeleteToken(c.Ctx(), c.Param("namespace"), namespace); err != nil {
+		if err == token.ErrUnauthorized {
```
otavio, line 83: Same here, prefer to use a `switch`.

api/token/service.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (s *service) ListTokens(ctx context.Context, namespace string) ([]models.Token, error) {
+	ns, _ := s.store.GetNamespace(ctx, namespace)
+	if ns == nil {
+		return []models.Token{}, ErrTokenNotFound
+	}
+
+	tokens := []models.Token{}
+	for _, token := range ns.Token.(primitive.A) {
```
otavio, line 42: Avoid using `primitive` as we ought to use the `store` to keep the storage use isolated.

api/apierr/service.go
```diff
@@ -0,0 +1,27 @@
+package apierr
```
otavio, line 1: error

api/apierr/service.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/shellhub-io/shellhub/api/apicontext"
+)
+
+var ErrUnauthorized = errors.New("unauthorized")
+var ErrUserNotFound = errors.New("user not found")
+var ErrResourceNotFound = errors.New("resource not found")
+var ErrDuplicateID = errors.New("user already member of this namespace")
+var ErrUserOwner = errors.New("cannot remove this user")
+
+func HandleErr(c apicontext.Context, err error) error {
```
otavio, line 16: ToApiError(...)

api/token/service.go
```diff
@@ -0,0 +1,67 @@
...
+type service struct {
+	store store.Store
+}
+
+func NewService(store store.Store) Service {
+	return &service{store}
+}
+
+func (s *service) ListTokens(ctx context.Context, namespace string) ([]models.Token, error) {
+	ns, _ := s.store.GetNamespace(ctx, namespace)
```
otavio, line 28: `ns, err` and check the `err`.

api/apierr/service.go
otavio, line 1: Has this been reverted from `error` on purpose?

api/apierr/service.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/shellhub-io/shellhub/api/apicontext"
+)
+
+var ErrUnauthorized = errors.New("unauthorized")
+var ErrUserNotFound = errors.New("user not found")
+var ErrResourceNotFound = errors.New("resource not found")
+var ErrDuplicateID = errors.New("user already member of this namespace")
+var ErrUserOwner = errors.New("cannot remove this user")
+
+func HandleError(c apicontext.Context, err error) error {
```
otavio, line 16: `ToApiError`

api/store/mongo/store.go
```diff
@@ -1469,6 +1469,54 @@ func (s *Store) GetSomeNamespace(ctx context.Context, ID string) (*models.Namesp
...
+
+	if result.ModifiedCount == 0 {
+		return ErrUserNotFound
+	}
+
+	return nil
+}
+
+func (s *Store) SetReadOnly(ctx context.Context, namespace string) error {
+	_, err := s.db.Collection("namespaces").UpdateOne(ctx, bson.M{"name": namespace}, bson.M{"&set": bson.M{"read_only": true}})
```
otavio, line 1512: You need to handle the case where the document is not found, don't you?

api/token/service.go
```diff
@@ -0,0 +1,91 @@
...
+	store store.Store
+}
+
+func NewService(store store.Store) Service {
+	return &service{store}
+}
+
+func (s *service) CreateToken(ctx context.Context, namespace string) (*models.Token, error) {
+	ns, err := s.store.GetNamespace(ctx, namespace)
+	if ns == nil {
```
otavio, line 31: I'd prefer if you handle the error case first and then the `ns` error. So we follow the same pattern in the whole code.

api/token/service.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return s.store.CreateToken(ctx, ns.Name, &models.Token{
+		ID:       hex.EncodeToString(hasher.Sum(nil)),
+		TenantID: ns.TenantID,
+		ReadOnly: true,
+	})
+}
+
+func (s *service) GetToken(ctx context.Context, namespace string) (*models.Token, error) {
+	ns, err := s.store.GetNamespace(ctx, namespace)
```
otavio, line 55: Same here. Please handle `err` and then `ns.`

api/token/service.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return s.store.GetToken(ctx, namespace)
+}
+
+func (s *service) DeleteToken(ctx context.Context, namespace string) error {
+	ns, err := s.store.GetNamespace(ctx, namespace)
+	if ns == nil {
```
otavio, line 69: ... and here.

api/token/service.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	return s.store.DeleteToken(ctx, namespace, ns.Token.(*models.Token))
+}
+
+func (s *service) SetReadOnly(ctx context.Context, namespace string) error {
+	ns, err := s.store.GetNamespace(ctx, namespace)
```
otavio, line 81: and here.

otavio (changes requested):

api/apierr/service.go
gustavosbarreto: “error” is a reserved word for the ERROR type

api/apierr/service.go
gustavosbarreto: HandleError handles errors and writes them according to our API conventions.

api/apierr/service.go
otavio: Ahh I see.

PR #726, opened 2021-02-19
fix /info to omit http host port from ssh host name

sammck:
If a nonstandard port is used for HTTP, it is included in ngx.var.http_host. Thus, when ":<ssh_port>" is appended, the result is that there are two port numbers appended to the value of the ssh endpoint in the /info api.  The solution is to use ngx.var.host rather than ngx.var.http_host for this case.

### DIFF
--- a/gateway/shellhub.conf
+++ b/gateway/shellhub.conf
@@ -182,11 +182,12 @@ server {
         default_type application/json;
 
         content_by_lua_block {
-            local host=ngx.var.http_host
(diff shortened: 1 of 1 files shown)


sammck:
Sorry about the commit message format nonconformity...

gustavosbarreto (approved):

gustavosbarreto:
@all-contributors please add @sammck for code

PR #736, opened 2021-02-23
On-demand connections

### LINKED ISSUE #717: On-demand agent connections
I'm interested in adapting shellhub to an environment with a very large number of client devices that have very intermittent connectivity, have battery life concerns, are infrequently accessed via SSH (only a very small number at one time), and for which the economics of frequently reestablishing and maintaining a persistent open socket from each agent to the shellhub server doesn't make sense.  These devices are on a network with an efficient but low-bandwidth/high latency notification/push channel (through HTTP polling, SMS notification or similar). It would be nice to have a configurable feature in shellhub whereby when an SSH client attempts to connect to an agent that does not have an active agent connection, a configured subprocess command could be invoked to initiate such an out-of-band push notification to the device, followed by waiting for a configurable timeout for the device agent to establish a connection before timing out the SSH connection attempt.

On the agent side, my custom out-of-band notification handler would receive the notification and start the shellhub device agent if it was not already started.  After a configurable period of inactivity (e.g., no active SSH connections), the shellhub agent would shut itself down. Alternatively, the shellhub server could disconnect from the device agent when its connection becomes idle, and the device agent would remain disconnected until another out-of-band notification is received.

Is there anything along these lines in your roadmap? Would you consider adding such a feature or be interested in related contributions?

gustavosbarreto:
Closes #717

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,9 @@ services:
       - PRIVATE_KEY=/run/secrets/ssh_private_key
       - SHELLHUB_ENTERPRISE=${SHELLHUB_ENTERPRISE}
       - RECORD_URL=${SHELLHUB_RECORD_URL}
+      - WEBHOOK_URL=${SHELLHUB_WEBHOOK_URL}
+      - WEBHOOK_PORT=${SHELLHUB_WEBHOOK_PORT}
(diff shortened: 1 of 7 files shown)


sammck:
Thanks for this 😊

PR #739, opened 2021-02-25
api: Return the number of devices in get namespace

leonardojoao:


otavio (dismissed):

api/store/mongo/store.go
```diff
@@ -1451,6 +1467,15 @@ func (s *Store) GetSomeNamespace(ctx context.Context, ID string) (*models.Namesp
 	return ns, nil
 }
 
+func (s *Store) CountDevicesNamespace(ctx context.Context, namespace string) (int, error) {
+	countDevice, err := s.db.Collection("devices").CountDocuments(ctx, bson.M{"tenant_id": namespace})
+	if err != nil {
+		return int(countDevice), err
+	}
```
otavio, line 1474: Here we ought to return 0, as this is an error case.

otavio (changes requested):

api/store/mongo/store.go
```diff
@@ -1451,6 +1466,15 @@ func (s *Store) GetSomeNamespace(ctx context.Context, ID string) (*models.Namesp
 	return ns, nil
 }
 
+func (s *Store) CountDevicesNamespace(ctx context.Context, namespace string) (int, error) {
```
gustavosbarreto, line 1469: is this function really needed?

gustavosbarreto (changes requested):

PR #752, opened 2021-03-01
ui: Fix to show only one add device button on the dashboard

leonardojoao:


ui/src/components/device/DeviceAdd.vue
```diff
@@ -76,6 +77,14 @@
 export default {
   name: 'DeviceAdd',
 
+  props: {
```
noreng-jg, line 80: Is this `smallButton` props really needed? Couldn't you just use `true` in `:small`?

otavio: Good point.

otavio (changes requested):

noreng-jg (approved):

ui/src/components/device/DeviceAdd.vue
otavio: @gustavosbarreto @leonardojoao this comment from @noreng-jg has not been resolved.

PR #758, opened 2021-03-03
ui: Pattern store code

leonardojoao:


ui/src/store/api/stats.js
```diff
@@ -1,5 +1,5 @@
 import http from '@/store/helpers/http';
 
-const getStats = async () => http().get('/stats');
+export const getStats = async () => http().get('/stats');
```
noreng-jg, line 3: Is there a need to export the same function twice?

leonardojoao: No, I will remove. I was trying to use export const, but we only have one function, we need to declare the default.

noreng-jg (approved):

PR #760, opened 2021-03-03
ui: Apply hyphens to urls

leonardojoao:


ui/src/store/api/public_keys.js
```diff
@@ -1,17 +1,17 @@
 import http from '@/store/helpers/http';
 
-export const postPublicKey = async (data) => http().post('/sshkeys/public_keys', {
+export const postPublicKey = async (data) => http().post('/sshkeys/public-keys', {
```
leonardojoao, line 3: @gustavosbarreto this part of the code, I switched to hyphens. But, to work it is necessary to make the exchange in the API. What do I do?

noreng-jg (approved):

PR #771, opened 2021-03-08
ui: revert vue-fragment

noreng-jg:


otavio:
@noreng-jg have you checked what changed from 1.5.1 and 1.5.2? The code change is very small so I am wondering if we have any misuse of it on ShellHub which need to be fixed?

If not, we ought to report the error on the repository so it can be fixed in a future release.

noreng-jg:
@otavio The latest version of the `vue-fragment` dependency has an issue with the `ui` DOM that has already been [reported](https://github.com/Thunberg087/vue-fragment/issues/61).
Also, the changes of this pr will prevent dependabot from generating pull requests while this issue is not fixed.

PR #782, opened 2021-03-10
ssh: compare hostname with a regex 

eduardoveiga:


api/routes/sshkeys.go
```diff
@@ -128,3 +129,24 @@ func CreatePrivateKey(c apicontext.Context) error {
...
+	pubKey, err := svc.GetPublicKey(c.Ctx(), c.Param("fingerprint"), c.Param("tenant"))
+	if err != nil {
+		return c.JSON(http.StatusForbidden, err)
+	}
+
+	if err := c.Bind(&device); err != nil {
+		return c.JSON(http.StatusForbidden, err)
+	}
+
+	ok, err := svc.EvaluateKeyHostname(c.Ctx(), pubKey, device)
```
otavio, line 146: You can move the `device` declaration just before its use here.

otavio, line 146: Check `err` and take `ok` as default `return` if all previous checks work.

pkg/api/client/client_internal.go
```diff
@@ -42,6 +43,21 @@ func (c *client) GetPublicKey(fingerprint, tenant string) (*models.PublicKey, er
 	return pubKey, nil
 }
 
+func (c *client) EvaluateKey(fingerprint string, dev *models.Device) (bool, error) {
+
+	resp, _, errs := c.http.Post(buildURL(c, fmt.Sprintf("/internal/sshkeys/public-keys/evaluate/%s", fingerprint))).Send(dev).End()
+
```
otavio, line 49: drop the empty lines.

ssh/server.go
```diff
@@ -244,10 +242,15 @@ func (*Server) publicKeyHandler(ctx sshserver.Context, pubKey sshserver.PublicKe
 
 	if ssh.FingerprintLegacyMD5(magicPubKey) != fingerprint {
 		apiClient := client.NewClient()
-		_, err = apiClient.GetPublicKey(fingerprint, device.Tenant)
+		_, err = apiClient.GetPublicKey(fingerprint, device.TenantID)
 		if err != nil {
 			return false
 		}
+
+		ok, _ := apiClient.EvaluateKey(fingerprint, device)
```
otavio, line 250: check `err` too.

ssh/ssh2ws.go
```diff
@@ -55,6 +55,13 @@ func HandlerWebsocket(ws *websocket.Conn) {
 			return
 		}
 
+		if ok, err := apiClient.EvaluateKey(fingerprint, device); !ok {
+			fmt.Println(err)
+			ws.Write([]byte("Permission denied\r\n")) // nolint:errcheck
+			ws.Close()
+			return
+		}
+
```
otavio, line 64: `ok` seems unused.

otavio (changes requested):

api/routes/sshkeys.go
eduardoveiga: this way?

api/routes/sshkeys.go
```diff
@@ -128,3 +129,24 @@ func CreatePrivateKey(c apicontext.Context) error {
...
+	if err != nil {
+		return err
+	}
+	return c.JSON(http.StatusOK, ok)
+
```
otavio, line 151: The newline should be before the return, not after.

pkg/api/client/client_internal.go
```diff
@@ -42,6 +43,19 @@ func (c *client) GetPublicKey(fingerprint, tenant string) (*models.PublicKey, er
 	return pubKey, nil
 }
 
+func (c *client) EvaluateKey(fingerprint string, dev *models.Device) (bool, error) {
+	var evaluate *bool
+	resp, _, errs := c.http.Post(buildURL(c, fmt.Sprintf("/internal/sshkeys/public-keys/evaluate/%s", fingerprint))).Send(dev).EndStruct(&evaluate)
+	if len(errs) > 0 {
+		return false, errs[0]
+	}
```
otavio, line 51: Newline after this very first block.

otavio (changes requested):

pkg/api/client/client_internal.go
otavio: I don't know if we're using the [`multierr`](https://github.com/uber-go/multierr) package. If not, we might look at it.

PR #806, opened 2021-03-18
api: fix error on pubkeys with different tenant

Asakiz:

### DIFF
--- a/api/sshkeys/service.go
+++ b/api/sshkeys/service.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/shellhub-io/shellhub/api/apicontext"
 	"github.com/shellhub-io/shellhub/api/store"
 	"github.com/shellhub-io/shellhub/pkg/api/paginator"
 	"github.com/shellhub-io/shellhub/pkg/models"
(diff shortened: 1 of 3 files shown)


api/sshkeys/service.go
```diff
@@ -49,10 +50,14 @@ func (s *service) CreatePublicKey(ctx context.Context, key *models.PublicKey) er
 
 	key.Fingerprint = ssh.FingerprintLegacyMD5(pubKey)
 
-	err = s.store.CreatePublicKey(ctx, key)
-	if err == store.ErrDuplicateFingerprint {
-		return ErrDuplicateFingerprint
+	returnedKey, err := s.store.GetPublicKey(ctx, key.Fingerprint, apicontext.TenantFromContext(ctx).ID)
+	if err == store.ErrRecordNotFound && returnedKey == nil {
```
otavio, line 54: Just check the err here as if this returns, we ought to break out.

otavio (changes requested):

api/sshkeys/service.go
Asakiz: we can't just return the error here, because the `recordNotFound` is a expected error, means there are no registered public keys with the same fingerprint and tenant_id.

PR #831, opened 2021-03-30
Topic/change message

leonardojoao:


noreng-jg (commented):
The "request has failed" message seems to be very generic, I believe a better approach
would be to use a specific message in snackbar or in the form, when the login response comes with the status 401, like "Invalid login" or "Username/password incorrect".

gustavosbarreto (approved):

PR #862, opened 2021-04-14
Topic/include agent version in http request

leonardojoao:


otavio (changes requested):
The agent version is always a semantic version right? If so, it'd be good to validate that.

@gustavosbarreto should we add those checks now?

gustavosbarreto:
> The agent version is always a semantic version right? If so, it'd be good to validate that.

No, not always. In development mode the version is hard-coded to "latest".

> @gustavosbarreto should we add those checks now?

Nice find, but this should be fixed in a separate commit.

gustavosbarreto:
@leonardojoao please append what issue this PR closes to the last commit message.

leonardojoao:
@gustavosbarreto 
Do I quote the same issue on both commits?

otavio:
> > The agent version is always a semantic version right? If so, it'd be good to validate that.
> 
> No, not always. In development mode the version is hard-coded to "latest".

Sure, in this case we can check if it is "latest" or a semver.

gustavosbarreto:
> @gustavosbarreto
> Do I quote the same issue on both commits?

No, just on the commit that changes the agent code.

PR #881, opened 2021-04-22
api: Returns conflict when inserting an existing namespace

leonardojoao:
This commit also removes the data return from the namespace, as it's
not necessary.

### DIFF
--- a/api/nsadm/service.go
+++ b/api/nsadm/service.go
@@ -21,6 +21,7 @@ var ErrUserNotFound = errors.New("user not found")
 var ErrNamespaceNotFound = errors.New("namespace not found")
 var ErrDuplicateID = errors.New("user already member of this namespace")
 var ErrUserOwner = errors.New("cannot remove this user")
+var ErrConflictName = errors.New("this name already exists")
 
 type Service interface {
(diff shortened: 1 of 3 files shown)


api/nsadm/service.go
```diff
@@ -77,7 +78,14 @@ func (s *service) CreateNamespace(ctx context.Context, namespace *models.Namespa
 		namespace.MaxDevices = -1
 	}
 
-	return s.store.NamespaceCreate(ctx, namespace)
+	err := s.store.NamespaceCreate(ctx, namespace)
+	if err != nil {
+		if strings.Contains(err.Error(), "duplicate key error") {
```
leonardojoao, line 83: I am looking for a solution for this type of verification.

otavio (dismissed):

cli/main.go
```diff
@@ -54,18 +54,14 @@ func main() {
...
-				ns, err := svc.NamespaceCreate(Parameters{
+				err := svc.NamespaceCreate(Parameters{
 					Namespace: args[0],
 					Username:  args[1],
 				})
 				if err != nil {
 					return err
 				}
 
-				fmt.Println("Namespace added: ", ns.Name)
```
gustavosbarreto, line 65: why was this removed?

gustavosbarreto (approved):

PR #882, opened 2021-04-22
cli: small fixes on cli

Asakiz:

### DIFF
--- a/bin/add-namespace
+++ b/bin/add-namespace
@@ -5,4 +5,4 @@
 NAMESPACE=$1
 OWNER=$2
 
-docker-compose exec cli ./cli add-namespace $NAMESPACE $OWNER
+docker-compose exec cli ./cli add-namespace "$NAMESPACE" "$OWNER" "$TENANT_ID"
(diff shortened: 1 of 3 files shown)


bin/add-namespace
```diff
@@ -5,4 +5,9 @@
 NAMESPACE=$1
 OWNER=$2
 
-docker-compose exec cli ./cli add-namespace $NAMESPACE $OWNER
+if [ -z "$TENANT_ID" ]
+then
+    docker-compose exec cli ./cli add-namespace $NAMESPACE $OWNER ""
+else
+    docker-compose exec cli ./cli add-namespace $NAMESPACE $OWNER $TENANT_ID
```
otavio, line 12: Not need. You can call it as:
```
docker-compose exec cli ./cli add-namespace "$NAMESPACE" "$OWNER" "$TENANT_ID"
```
This allow the removal of the `if` block.

cli/service.go
```diff
@@ -116,10 +117,14 @@ func (s *service) NamespaceCreate(data Parameters) (*models.Namespace, error) {
 		return nil, ErrDuplicateNamespace
 	}
 
+	if data.TenantID != "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" {
```
otavio, line 120: This kind of conditional worries me. What does this special tenant id offer?

@gustavosbarreto @eduardoveiga

otavio (changes requested):

cli/service.go
Asakiz: Same to me, i think this is not really necessary.

gustavosbarreto (approved):

PR #893, opened 2021-04-26
api: refactor to use early-return pattern

leonardojoao:


api/deviceadm/service.go
```diff
@@ -88,24 +88,30 @@ func (s *service) DeleteDevice(ctx context.Context, uid models.UID, tenant, user
...
-			device.Name = name
-			if err := validate.Struct(device); err == nil {
-				otherDevice, _ := s.store.DeviceGetByName(ctx, name, tenant)
-				if otherDevice == nil {
-					return s.store.DeviceRename(ctx, uid, name)
-				}
-			}
-		}
+	if device.Name == name {
+		return			// I need return the "Error name is equal"
```
otavio, line 102: return nil // No operation needed.

api/deviceadm/service.go
```diff
@@ -88,24 +88,30 @@ func (s *service) DeleteDevice(ctx context.Context, uid models.UID, tenant, user
...
 	}
-	return ErrUnauthorized
+
+	otherDevice, err := s.store.DeviceGetByName(ctx, name, tenant)
+	if err != nil {
+		return err		// Not Found
+	}
+
+	if otherDevice != nil {
+		return 			// You already have this device name in the namespace
```
otavio, line 111: return ErrDuplicatedDeviceName

otavio (changes requested):
The commit log could be reworded as:

api: refactor to use early-return pattern

api/deviceadm/service.go
leonardojoao: Could i return a specific error stating that the name is the same?

otavio: I think it does make sense. And who uses it, need to handle this as well.

api/deviceadm/service.go
```diff
@@ -88,24 +88,30 @@ func (s *service) DeleteDevice(ctx context.Context, uid models.UID, tenant, user
 func (s *service) RenameDevice(ctx context.Context, uid models.UID, name, tenant, username string) error {
 	err := s.isNamespaceOnwer(ctx, tenant, username)
 	if err != nil {
-		return err
+		return ErrUnauthorized
+	}
+
+	device, err := s.store.DeviceGetByUID(ctx, uid, tenant)
+	if err != nil {
+		return err		// Not Found
```
leonardojoao, line 96: I could return ErrDeviceNotFound?

api/deviceadm/service.go
leonardojoao: Yes, I will do this

otavio (dismissed):

gustavosbarreto (approved):

PR #899, opened 2021-04-27
cli: fix the error when the provided input is not enough

Asakiz:

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -34,6 +34,7 @@ func main() {
 	rootCmd.AddCommand(&cobra.Command{
 		Use:   "add-user",
 		Short: "Usage: <username> <password> <email>",
+		Args:  cobra.ExactArgs(3),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			username, err := svc.UserCreate(Parameters{
 				Username: args[0],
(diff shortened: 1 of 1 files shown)


cli/main.go
```diff
@@ -53,7 +54,12 @@ func main() {
 		&cobra.Command{
 			Use:   "add-namespace",
 			Short: "Usage: <namespace> <owner>",
+			Args:  cobra.MinimumNArgs(2),
 			RunE: func(cmd *cobra.Command, args []string) error {
+				if len(args) == 2 {
+					args = append(args, "")
+				}
```
otavio, line 61: This looks like a dirty hack. Why?

otavio (changes requested):

cli/main.go
Asakiz: because of the `docker-compose exec cli ./cli add-namespace <namespace> <owner>` command that should not be used unless you know what you are doing, this avoids creating a panic when `tenant_id` is not informed.

otavio: So please add a clear log explaining this.

Asakiz: Sure, i'll do it

cli/main.go
```diff
@@ -53,7 +54,14 @@ func main() {
 		&cobra.Command{
 			Use:   "add-namespace",
 			Short: "Usage: <namespace> <owner>",
+			Args:  cobra.RangeArgs(2, 3),
 			RunE: func(cmd *cobra.Command, args []string) error {
+				// this line was added because of a special case when the command is not running by the script (bin/add-namespace).
+				// set the TenantID as an empty string avoid the panic when he is missing.
```
otavio, line 60:
```go
// Avoid panic when TenandID isn't provided.
```

otavio (changes requested):

otavio (approved):

PR #902, opened 2021-04-27
api: refactor to use early-return pattern

leonardojoao:
This commit apply early-return pattern in deleteDevice and updatePendingStatus.


api/deviceadm/service.go
```diff
@@ -130,37 +131,42 @@ func (s *service) UpdateDeviceStatus(ctx context.Context, uid models.UID, online
 
 func (s *service) UpdatePendingStatus(ctx context.Context, uid models.UID, status, tenant, username string) error {
 	err := s.isNamespaceOnwer(ctx, tenant, username)
+	if err != nil {
```
otavio, line 134:
```go
if err := s.isNamespaceOnwer(ctx, tenant, username); err != nil {
```

otavio (changes requested):

otavio (approved):

otavio (approved):

api/deviceadm/service.go
```diff
@@ -129,38 +127,42 @@ func (s *service) UpdateDeviceStatus(ctx context.Context, uid models.UID, online
...
+			}
+		} else {
+			ns, err := s.store.NamespaceGet(ctx, device.TenantID)
+			if err != nil {
+				return err
+			}
+
+			if ns.MaxDevices > 0 && ns.MaxDevices <= ns.DevicesCount {
+				return ErrMaxDeviceCountReached
 			}
```
otavio, line 163:
```suggestion
		if sameMacDev.UID == device.UID {
			ns, err := s.store.NamespaceGet(ctx, device.TenantID)
			if err != nil {
				return err
			}

			if ns.MaxDevices > 0 && ns.MaxDevices <= ns.DevicesCount {
				return ErrMaxDeviceCountReached
			}

		if err := s.store.SessionUpdateDeviceUID(ctx, models.UID(sameMacDev.UID), models.UID(device.UID)); err != nil {
			return err
		}
		if err := s.store.DeviceDelete(ctx, models.UID(sameMacDev.UID)); err != nil {
			return err
		}
		if err := s.store.DeviceRename(ctx, models.UID(device.UID), sameMacDev.Name); err != nil {
			return err
		}
```

otavio (changes requested):

api/deviceadm/service.go
leonardojoao: I think a "}" is missing. It turns out that I didn't quite understand, the idea is to invert the logic of the if by the else?

otavio: Yes, it was missing a `}`. The question was if we could drop the else and as a plus reduce one level.

leonardojoao: In the last version I sent, it was found that sameMacDev can be nil. Then we couldn’t drop the else.

otavio (approved):

PR #917, opened 2021-05-04
agent: get container id from mountinfo when cgroup is v2

eduardoveiga:


pkg/dockerutils/utils.go
```diff
@@ -8,6 +8,15 @@ import (
 
 // CurrentContainerID returns the current running container ID
 func CurrentContainerID() (string, error) {
+	id, err := currentContainerIDByCgroup()
+	if id == "" {
```
otavio, line 12: err please.

otavio (changes requested):

pkg/dockerutils/utils.go
```diff
@@ -1,29 +1,47 @@
...
 
+var ErrContainerIDNotFound = errors.New("container id not found")
+
 // CurrentContainerID returns the current running container ID
 func CurrentContainerID() (string, error) {
-	f, err := os.Open("/proc/self/cgroup")
+	fCgroup, err := os.Open("/proc/self/cgroup")
+	if err != nil {
+		return "", err
+	}
```
otavio, line 18: defer.

pkg/dockerutils/utils.go
```diff
@@ -1,29 +1,47 @@
...
+
+	fMountInfo, err := os.Open("/proc/self/mountinfo")
 	if err != nil {
 		return "", err
 	}
```
otavio, line 23: defer.

pkg/dockerutils/utils.go
```diff
@@ -1,29 +1,47 @@
...
-	defer f.Close()
 
-	content, err := ioutil.ReadAll(f)
+	defer fCgroup.Close()
+	defer fMountInfo.Close()
+
+	reader := io.MultiReader(fCgroup, fMountInfo)
+
```
otavio, line 29: drop the empty line.

pkg/dockerutils/utils.go
```diff
@@ -1,29 +1,47 @@
...
+	re := regexp.MustCompilePOSIX(`([0-9]+:[a-z_,=]+.*docker[/-]|/var/lib/docker/containers)?([0-9a-f]{64})`)
 	line := re.FindAllSubmatch(content, -1)
-	if len(line) <= 0 || len(line[0]) != 2 {
-		return "", nil
+	if len(line) <= 0 || len(line[0]) != 3 {
+		return "", ErrContainerIDNotFound
 	}
 
-	return string(line[0][1]), nil
+	if string(line[0][2]) == "" {
```
otavio, line 41: use:

```
containerId := string(line[0][2])
...
```

pkg/dockerutils/utils.go
```diff
@@ -1,29 +1,47 @@
...
 	if err != nil {
 		return "", err
 	}
 
-	re := regexp.MustCompilePOSIX(`[0-9]+:[a-z_,=]+.*docker[/-]([0-9a-f]{64})`)
+	re := regexp.MustCompilePOSIX(`([0-9]+:[a-z_,=]+.*docker[/-]|/var/lib/docker/containers)?([0-9a-f]{64})`)
 	line := re.FindAllSubmatch(content, -1)
-	if len(line) <= 0 || len(line[0]) != 2 {
-		return "", nil
+	if len(line) <= 0 || len(line[0]) != 3 {
```
otavio, line 37: This is confusing. Why 3? Why <= 0.

otavio (changes requested):

pkg/dockerutils/utils.go
eduardoveiga: I'm using ```len(line[0]) != 3 ``` because now the regexp has 2 grouping 
```	re := regexp.MustCompilePOSIX(`([0-9]+:[a-z_,=]+.*docker[/-]|/var/lib/docker/containers/)([0-9a-f]{64})`)```
in golang regexp documentation sintax there's a  non-capturing group syntax but does not work here

about len(line) <= 0  I will see it's necessary

otavio: if matches == nil || len(matches) != 3 {
   return NotFound
}

PR #934, opened 2021-05-12
refactor local variable in catch

leonardojoao:


noreng-jg (approved):

ui/src/components/device/Device.vue
```diff
@@ -120,11 +120,18 @@ export default {
...
       this.show = true;
     } catch (error) {
-      if (error.response.status === 403) {
+      switch (true) {
+      case (error.response.status === 403): {
         this.$store.dispatch('snackbar/showSnackbarErrorAssociation');
-      } else {
+        break;
+      }
+      case (error.response.status === 401): {
```
leonardojoao, line 128: You could put the error codes in order.

PR #954, opened 2021-05-18
add support to activate and recover the account

leonardojoao:


api/store/mongo/migrations/migration_23.go
```diff
@@ -0,0 +1,48 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+var migration_23 = migrate.Migration{
+	Version: 23,
+	Up: func(db *mongo.Database) error {
+		logrus.Info("Applying migration 23 - Up")
+		mod := mongo.IndexModel{
```
otavio, line 17: `mod` here is too generic. Please use `index_model` or at least `index` so it is clear what it is doing when used below.

You can keep reusing same name for all collections. No problem.

otavio (changes requested):

api/store/mongo/user_store.go
```diff
@@ -246,6 +246,49 @@ func (s *Store) UserUpdateFromAdmin(ctx context.Context, name, username, email,
...
+	if _, err := s.db.Collection("recover_password").InsertOne(ctx, token); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *Store) UserGetToken(ctx context.Context, ID string) (*models.UserRecoverPassword, error) {
+	token := new(models.UserRecoverPassword)
+
```
otavio, line 263: Drop the empty line as this is a single operation block.

otavio (changes requested):

api/store/mongo/migrations/main.go
```diff
@@ -32,6 +32,7 @@ func GenerateMigrations() []migrate.Migration {
 		migration_20,
 		migration_21,
 		migration_22,
+		migration_23,
```
otavio, line 35: There is already a 23 migration, it'll need to be moved to 24 or 25.

leonardojoao: Yes, I will when I liberate the PR. Because it can go into some other migration, so not to change more than once, I ended up not modifying it yet.

PR #956, opened 2021-05-18
ssh: refactor the code

Asakiz:

### DIFF
--- a/pkg/api/client/client_internal.go
+++ b/pkg/api/client/client_internal.go
@@ -4,6 +4,7 @@ package client
 
 import (
 	"fmt"
+	"net/http"
 
 	"github.com/shellhub-io/shellhub/pkg/models"
 	"go.uber.org/multierr"
@@ -25,6 +26,13 @@ type internalAPI interface {
 	GetPublicKey(fingerprint, tenant string) (*models.PublicKey, error)
 	CreatePrivateKey() (*models.PrivateKey, error)
 	EvaluateKey(fingerprint string, dev *models.Device) (bool, error)
+	DevicesOffline(id string) error
+	FirewallEvaluate(lookup map[string]string) []error
+	PatchSessions(uid string) []error
+	FinishSession(uid string) []error
+	RecordSession(session *models.SessionRecorded, opts interface{})
+	Lookup(lookup map[string]string) (string, []error)
+	DeviceLookup(lookup map[string]string) (*models.Device, []error)
 }
 
 func (c *client) LookupDevice() {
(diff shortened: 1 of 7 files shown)


otavio (dismissed):

pkg/api/client/client.go
```diff
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"path"
 
+	"github.com/gliderlabs/ssh"
```
gustavosbarreto, line 13: This package shouldn't be imported here because it is specific to ssh service.

pkg/api/client/client.go
```diff
@@ -109,3 +110,41 @@ func buildURL(c *client, uri string) string {
 
 	return u.String()
 }
+
+func DevicesOffline(id string) {
```
gustavosbarreto, line 114: Why aren't you using `*client struct` as method receiver?

pkg/api/client/client.go
```diff
@@ -109,3 +110,41 @@ func buildURL(c *client, uri string) string {
...
+	gorequest.New().Post(fmt.Sprintf("http://api:8080/internal/devices/%s/offline", id)).End()
+}
+
+func GetDevices(target string, device *models.Device) (gorequest.Response, []error) {
```
gustavosbarreto, line 118: same here

pkg/api/client/client.go
```diff
@@ -109,3 +110,41 @@ func buildURL(c *client, uri string) string {
...
+	res, _, errs := gorequest.New().Get("http://api:8080/api/devices/" + target).EndStruct(&device)
+
+	return res, errs
+}
+
+func FirewallEvaluate(lookup map[string]string) (gorequest.Response, []error) {
```
gustavosbarreto, line 124: ...

pkg/api/client/client.go
Asakiz: The reason of this import is because of the func RecordSession(uid string, waitingString string, pty ssh.Pty, opts models.ConfigOptions)

otavio (approved):

pkg/api/client/client_internal.go
```diff
@@ -25,6 +26,13 @@ type internalAPI interface {
 	GetPublicKey(fingerprint, tenant string) (*models.PublicKey, error)
 	CreatePrivateKey() (*models.PrivateKey, error)
 	EvaluateKey(fingerprint string, dev *models.Device) (bool, error)
+	DevicesOffline(id string)
```
gustavosbarreto, line 29: return error?

pkg/models/session.go
```diff
@@ -31,3 +31,24 @@ type RecordedSession struct {
...
+	Record string `json:"record"`
+	Height int    `json:"height"`
+	Width  int    `json:"width"`
+}
+
+type Status struct {
+	Authenticated bool `json:"authenticated"`
+}
+
+type ConfigOptions struct {
```
gustavosbarreto, line 45: It seems related to config environment ssh service, should not declared in models.

otavio (approved):

PR #961, opened 2021-05-19
gateway: add automated tests

gustavosbarreto:
We have around 20 location NGINX directives for 5 underlying
microservices. This level of complexity slowed down our dev
and release process because every change had to be verified
manually in a separate test environment.

So, we are introducing automated tests for the API gateway.

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -1,4 +1,4 @@
-FROM openresty/openresty:1.15.8.2-6-alpine
+FROM openresty/openresty:1.15.8.2-6-alpine as base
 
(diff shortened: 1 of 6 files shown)


gustavosbarreto:
The gateway no longer has nginx locations to test. The edge is a Go binary that embeds Caddy and generates its configuration from the environment (#6780).

What this PR asked for exists now in a different shape: `gateway/caddy_test.go` puts 12 environment shapes through Caddy's real config adapter, so a broken directive fails in `go test` instead of at a customer's boot, and `gateway` is a job in the CI matrix (#6778).

Closing, since the files this branch patches are gone.

PR #970, opened 2021-05-24
api: prevent concurrent migrations running from several instances

Asakiz:

### DIFF
--- a/api/store/mongo/migrations.go
+++ b/api/store/mongo/migrations.go
@@ -1,11 +1,40 @@
 package mongo
 
 import (
+	"context"
+
 	"github.com/shellhub-io/shellhub/api/store/mongo/migrations"
+	"github.com/sirupsen/logrus"
+	"github.com/square/mongo-lock"
 	migrate "github.com/xakep666/mongo-migrate"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/writeconcern"
 )
 
 func ApplyMigrations(db *mongo.Database) error {
-	return migrate.NewMigrate(db, migrations.GenerateMigrations()...).Up(migrate.AllAvailable)
+	logrus.Info("Creating lock for the resource migrations")
+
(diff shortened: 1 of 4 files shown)


otavio (approved):

otavio:
@Asakiz, what is still being worked on here? If this is ready, please rebase and send it for review.

Asakiz:
> @Asakiz, what is still being worked on here? If this is ready, please rebase and send it for review.

Done!

otavio (approved):

otavio (changes requested):
So from previous discussion, the new todo items are:

- [x] move the lock to the store
- [ ] avoid the write concern to be global

api/store/mongo/migrations.go
```diff
@@ -1,11 +1,39 @@
...
 	"github.com/shellhub-io/shellhub/api/store/mongo/migrations"
+	"github.com/sirupsen/logrus"
+	"github.com/square/mongo-lock"
 	migrate "github.com/xakep666/mongo-migrate"
 	"go.mongodb.org/mongo-driver/mongo"
 )
 
 func ApplyMigrations(db *mongo.Database) error {
-	return migrate.NewMigrate(db, migrations.GenerateMigrations()...).Up(migrate.AllAvailable)
+
```
otavio, line 14: Don' use empty lines at begin of the method.

api/store/mongo/migrations.go
```diff
@@ -1,11 +1,39 @@
...
+	lockClient := lock.NewClient(db.Collection("migrations"))
+	if err := lockClient.CreateIndexes(context.TODO()); err != nil {
+		logrus.WithError(err).Fatal("Failed to create a lock for the database")
+	}
+
+	logrus.Info("Locking the resource migrations")
+
+	lockId := "0"
+
+	if err := lockClient.XLock(context.TODO(), "migrations", lockId, lock.LockDetails{}); err != nil {
```
otavio, line 26: Do we need to use write consensus here?

Asakiz: Yes, because every process has her own instance of the store. But, they share the same resource which is the database, so we need to establish a write consensus.

otavio (approved):

otavio:
LGTM.

PR #979, opened 2021-05-30
api: rename var namepace for tenantID on namespace services and store

eduardoveiga:


otavio (changes requested):
Please explain on commit log why to do this change. There must be a reason why it should be considered?

otavio (approved):

PR #980, opened 2021-05-30
api: refuse to create a device if tenant is invadid

eduardoveiga:


otavio (approved):

otavio (changes requested):
This requires a migration to remove devices in this situation. We may have leftovers and it is important to remove them here.

api/authsvc/service.go
```diff
@@ -65,6 +65,11 @@ func (s *service) AuthDevice(ctx context.Context, req *models.DeviceAuthRequest)
 		LastSeen:  time.Now(),
 	}
 
+	namespace, err := s.store.NamespaceGet(ctx, device.TenantID)
```
otavio, line 68: Please add following comment here:
```go
# The order here is critical as we don't want to register devices if the tenant id is invalid
```

otavio (commented):
Please add the migration in this PR so we review it  together.

api/store/mongo/migrations/main.go
```diff
@@ -33,6 +33,8 @@ func GenerateMigrations() []migrate.Migration {
 		migration_21,
 		migration_22,
 		migration_23,
+		migration_24,
```
otavio, line 36: here is wrong. This should include just 25.

gustavosbarreto (approved):

api/store/mongo/migrations/main.go
```diff
@@ -34,6 +34,7 @@ func GenerateMigrations() []migrate.Migration {
 		migration_22,
 		migration_23,
 		migration_24,
+		migration_25,
```
otavio, line 37: This belongs to the migration patch.

otavio (changes requested):

gustavosbarreto (changes requested):
Why didn't you remove the devices in migration?

PR #990, opened 2021-06-01
api: add shellhub envs package

noreng-jg:


otavio (commented):
I'd like you to consider the possibility of we offer a small public API for this which would access every variable and allow for future easier conversion and finding where we'd use it.

One benefit is to avoid string keys on the rest of the code so we'd use:
```go
envs.GlobalEnvs.PrivateKey()
```
instead of 
```go
envs.GlobalEnvs.GetEnv("private_key")
```

otavio (changes requested):
The `api/store/mocks` change is not related to this PR. Please split it on another PR as this seem to have been missed when we applied the #979 PR.

otavio:
I applied for the new PR, please rebase this on top of the current master branch.

PR #997, opened 2021-06-04
pkg: complement regexp for btrfs mounts

eduardoveiga:
- [x] fix btrfs mount regexp
- [ ] add unittest so we avoid this to happen again


gustavosbarreto (approved):

otavio (commented):
Please add the unit test here. This is critical and seems to not have test coverage. 

I don't want to hold this fix to the unit test addition but it'd be nice to include it soon.

PR #1000, opened 2021-06-04
Do not allow dot in namespace and hostname

eduardoveiga:
- [x] ensure API do not allow wrong namespace and hostname - @eduardoveiga 
- [x] migrate old namespace and hostname replacing `.` to `-` - @eduardoveiga 
- [x] validate the hostname and namespace on UI to avoid sending wrong data - @noreng-jg

### DIFF
--- a/api/store/mongo/migrations/migration_23.go
+++ b/api/store/mongo/migrations/migration_23.go
@@ -1,0 +1,45 @@
+package migrations
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+	migrate "github.com/xakep666/mongo-migrate"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+var migration_23 = migrate.Migration{
+	Version:     23,
+	Description: "change dot in namespace name and hostname to -",
+	Up: func(db *mongo.Database) error {
+		logrus.Info("Applying migration 23 - Up")
+		if _, err := db.Collection("namespaces").UpdateMany(context.TODO(), bson.D{}, []bson.M{
+			{
+				"$set": bson.M{"name": bson.M{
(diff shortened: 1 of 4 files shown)


api/store/mongo/migrations/migration_23.go
```diff
@@ -0,0 +1,39 @@
...
+var migration_23 = migrate.Migration{
+	Version:     23,
+	Description: "change dot in namespace name and hostname to -",
+	Up: func(db *mongo.Database) error {
+		logrus.Info("Applying migration 23 - Up")
+		result, err := db.Collection("namespaces").UpdateMany(context.TODO(), bson.D{}, []bson.M{
+			{
+
+				"$set": bson.M{"name": bson.M{
+					"$replaceOne": bson.M{"input": "$name", "find": ".", "replacement": "-"},
```
otavio, line 24: What happens if there are multiple `.` on the field?

eduardoveiga: I changed to "replaceAll" and now works with multiple dots

api/store/mongo/migrations/migration_23.go
```diff
@@ -0,0 +1,47 @@
+package migrations
+
+import (
+	"context"
+
+	//"github.com/shellhub-io/shellhub/pkg/models"
```
otavio, line 6: Please remove the dead code from here.

otavio (changes requested):

otavio (approved):

otavio (changes requested):
@eduardoveiga please fix the comments on the migration so we can consider this for merging.

api/store/mongo/migrations/migration_23.go
eduardoveiga: done

otavio (approved):

PR #1021, opened 2021-06-08
Add clock and uuid package

leonardojoao:


otavio (changes requested):
As this is a new pkg we could be a good citizen and add the docstrings?

We could also document the package. So we start doing it for all new code and eventually do it to old code.

leonardojoao:
> As this is a new pkg we could be a good citizen and add the docstrings?
> 
> We could also document the package. So we start doing it for all new code and eventually do it to old code.

What would the idea of ​​documentation be, or would there be an example of which model to follow?

otavio:
@leonardojoao take a look at https://golang.org/src/go/doc/example.go

pkg/clock/clock.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"time"
+)
+
+// Clock is an interface to store values ​​of type time
+type Clock interface {
+	Now() time.Time
+}
+
+// Get value from environment variable
```
otavio, line 12: environment?

otavio (approved):

otavio (commented):
Typo on commit log "phg: use uuid package"

pkg/clock/clock.go
leonardojoao: No, this is the defaultBackend which is used to configure the defaultBackend.

otavio: Ok; good. So it is not related to environment. Good. Closing it.

otavio:
@eduardoveiga @leonardojoao I sent a lint to avoid those packages to ever to be used.

Some errors:

pkg/models/device.go:4:2: `time` is in the blacklist: time access must be done using 'github.com/shellhub-io/shellhub/pkg/clock' (depguard)
        "time"
        ^
pkg/models/license.go:3:8: `time` is in the blacklist: time access must be done using 'github.com/shellhub-io/shellhub/pkg/clock' (depguard)
import "time"
       ^
pkg/models/privatekey.go:3:8: `time` is in the blacklist: time access must be done using 'github.com/shellhub-io/shellhub/pkg/clock' (depguard)
import "time"

PR #1058, opened 2021-06-25
ui: fix web terminal dimensions

noreng-jg:
Updates cols and rows to computed for adapting terminal to container box.

### DIFF
--- a/ui/src/components/terminal/TerminalDialog.vue
+++ b/ui/src/components/terminal/TerminalDialog.vue
@@ -178,6 +178,13 @@ export default {
   },
 
   computed: {
+    webTermDimensions() {
+      return {
+        cols: this.xterm.cols,
+        rows: this.xterm.rows,
+      };
+    },
+
     show: {
       get() {
         return this.$store.getters['modals/terminal'] === this.$props.uid;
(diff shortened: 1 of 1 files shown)


ui/src/components/terminal/TerminalDialog.vue
```diff
@@ -178,6 +178,13 @@ export default {
   },
 
   computed: {
+    xtermDimensions() {
```
otavio, line 181: This is used only on the web terminal, correct? In this case, it might be better to call it as `webTermDimensions`

ui/src/components/terminal/TerminalDialog.vue
```diff
@@ -238,36 +245,22 @@ export default {
...
     connectWithPassword() {
-      const params = Object.entries({
-        user: `${this.username}@${this.$props.uid}`,
-        passwd: encodeURIComponent(this.passwd),
-        cols: this.xterm.cols,
-        rows: this.xterm.rows,
-      })
-        .map(([k, v]) => `${k}=${v}`)
-        .join('&');
+      const passwd = encodeURIComponent(this.passwd);
```
otavio, line 248: Is this `encodeURIComponent` already existent? if so, split the code rework and the fix into two different commits.

otavio (changes requested):

ui/src/components/terminal/TerminalDialog.vue
```diff
@@ -238,36 +245,22 @@ export default {
...
+      this.connect({ passwd });
+    },
 
-      this.connect(params);
+    wsURL(params) {
```
otavio, line 252: `encodeURLparams`

PR #1065, opened 2021-06-28
api: update service tests to table driven

noreng-jg:
Updates coverage of tests in service to handle error cases

### DIFF
--- a/api/nsadm/service_test.go
+++ b/api/nsadm/service_test.go
@@ -52,7 +52,7 @@ func TestGetNamespace(t *testing.T) {
 }
 
 func TestCreateNamespace(t *testing.T) {
-	tests := []struct {
+	cases := []struct {
 		name       string
 		isCloud    bool
 		maxDevices int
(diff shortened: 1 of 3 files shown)


api/sessionmngr/service_test.go
```diff
@@ -18,18 +19,65 @@ func TestListSessions(t *testing.T) {
...
+			},
+			expected: Expected{
+				sessions: sessions,
+				count:    len(sessions),
+				err:      nil,
+			},
+		},
+	}
+
+	for _, test := range tests {
```
otavio, line 74: It seems the usual code style for this is

```go
for _, tc := range cases {
```

This will allow you to add the following to the linter config:

```yaml
issues:
  exclude:
     - Using the variable on range scope `tc` in function literal
```

So we don't need to whitelist those one by one.

otavio (changes requested):

api/nsadm/service_test.go
```diff
@@ -52,7 +52,7 @@ func TestGetNamespace(t *testing.T) {
 }
 
 func TestCreateNamespace(t *testing.T) {
-	tests := []struct {
+	cases := []struct {
```
otavio, line 55: This change belongs to the other patch.

otavio (changes requested):

otavio (approved):

PR #1066, opened 2021-06-28
pkg: fix url in record session

leonardojoao:


Asakiz (approved):

otavio (changes requested):
The change looks good, but the commit log does not explain what the error was. Please extend the commit log, so we have a more detailed explanation.

otavio (approved):

PR #1073, opened 2021-06-29
api: fix permission to delete namespace members

eduardoveiga:


otavio (approved):

otavio:
@eduardoveiga please see if you can add a test case to exercise this error case.

otavio (changes requested):
Please add a test to ensure it doesn't break in the future.

PR #1074, opened 2021-06-29
api: fix namespace return errors

eduardoveiga:


noreng-jg (dismissed):

otavio (changes requested):
The change looks fine. However, I'd also to have a test to avoid this kind of error return. Can you add it?

api/nsadm/service.go
```diff
@@ -45,6 +45,32 @@ func NewService(store store.Store) Service {
 	return &service{store}
 }
 
+func (s *service) isNamespaceOwner(ctx context.Context, tenantID, ownerID string) (*models.Namespace, error) {
```
otavio, line 48: GetNamespaceForOwner

api/nsadm/service.go
```diff
@@ -219,7 +198,15 @@ func (s *service) RemoveNamespaceUser(ctx context.Context, tenantID, username, o
 }
 
 func (s *service) EditSessionRecordStatus(ctx context.Context, sessionRecord bool, tenant, ownerID string) error {
-	if _, err := s.GetNamespaceForOwner(ctx, tenant, ownerID); err != nil {
+	err := utils.IsNamespaceOwner(ctx, s.store, tenant, ownerID)
+	if err != nil {
```
otavio, line 202: This can be checked on a single line; there are other similar cases too.

otavio (changes requested):

api/routes/nsadm.go
```diff
@@ -142,6 +144,9 @@ func EditNamespace(c apicontext.Context) error {
 
 	namespace, err := svc.EditNamespace(c.Ctx(), c.Param("id"), req.Name, id)
 	if err != nil {
+		if err == nsadm.ErrInvalidFormat {
+			return c.NoContent(http.StatusBadRequest)
+		}
```
otavio, line 149: This is a perfect usecase for the switch.

otavio (dismissed):

otavio (approved):

noreng-jg (approved):

PR #1082, opened 2021-07-02
pkg: change recordUrl var name to recordURL

eduardoveiga:


eduardoveiga:
```
Running Golang linter on pkg
api/client/client_internal.go:33:49: "interface method parameter recordUrl should be recordURL" (revive)
	RecordSession(session *models.SessionRecorded, recordUrl string)
	                                               ^
api/client/client_internal.go:116:65: "method parameter recordUrl should be recordURL" (revive)
func (c *client) RecordSession(session *models.SessionRecorded, recordUrl string) {
                                                                ^
```

linter script it's failing on revive, but not linter on ci

otavio (approved):

PR #1098, opened 2021-07-07
api: add stats store interface

noreng-jg:


api/store/mongo/stats_store.go
```diff
@@ -8,6 +8,31 @@ import (
...
+
+	if !cursor.Next(context.TODO()) {
+		return 0, err
+	}
+
+	if err = cursor.Decode(&resp); err != nil {
+		return 0, err
+	}
+
+	return resp.Count, nil
```
otavio, line 33: This can use the `aggregateCount`, right?

otavio (changes requested):

otavio (approved):

otavio (approved):

PR #1126, opened 2021-07-19
gateway: mount /etc/resolve.conf within the container via kickstart

### LINKED ISSUE #593: the agent docker was stuck on dns lookup when there was connectivity  problem
It seem the agent docker was stuck on dns lookup and required restart the agent docker to solved.

`2020/12/29 07:39:52 [DEBUG] GET https://sh.mydomain.net:443/info: retrying in 30s (2147483423 left)
2020/12/29 07:40:22 [ERR] GET https://sh.mydomain.net:443/info request failed: Get https://sh.mydomain.net:443/info: dial tcp: lookup sh.mydomain.net on [::1]:53: read udp [::1]:43645->[::1]:53: read: connection refused`

jcarrus:
This commit mounts the host's `/etc/resolv.conf` file into the shellhub container via the `/gateway/kickstart.sh` file .

This is necessary to handle a situation where the host's `/etc/resolv.conf` file changes AFTER the shellhub container is started. As noted in the docker documentation (https://docs.docker.com/config/containers/container-networking/#dns-services)

> By default, a container inherits the DNS settings of the host, as defined in the /etc/resolv.conf configuration file. Containers that use the default bridge network get a copy of this file, whereas containers that use a custom network use Docker’s embedded DNS server, which forwards external DNS lookups to the DNS servers configured on the host.

The implication of "get a copy of this file" is that future changes to the host's file will not propogate to the container.

Fixes #593

### DIFF
--- a/agent/selfupdater/updater_docker.go
+++ b/agent/selfupdater/updater_docker.go
@@ -112,6 +112,21 @@ func (d *dockerUpdater) CompleteUpdate() error {
 			}
 		}
 
+		// Append /etc/resolv.conf to mount if old container
+		// version is less than v0.7.3 since /etc/resolv.conf
+		// is required inside the container to update host
+		// networking after boot.
+		v0_7_3, _ := semver.NewVersion("v0.7.3")
+		if v.LessThan(v0_7_3) {
+			parent.info.HostConfig.Mounts = []mount.Mount{
+				{
+					Type:   mount.TypeBind,
+					Source: "/etc/resolv.conf",
+					Target: "/etc/resolv.conf",
+				},
+			}
+		}
+
 		_, err = d.updateContainer(parent, container.info.Config.Image, parent.info.Name, false)
 		if err != nil {
 			return err
--- a/agent/selfupdater/updater_docker.go
+++ b/agent/selfupdater/updater_docker.go
@@ -72,65 +72,66 @@ func (d *dockerUpdater) CompleteUpdate() error {
 		return err
 	}
 
-	if parent != nil {
-		if err := d.stopContainer(parent); err != nil {
-			return err
-		}
+	if parent == nil {
+		return nil
+	}
 
-		_, pv := parent.splitImageVersion()
-		v, _ := semver.NewVersion(pv)
-		v0_4_0, _ := semver.NewVersion("v0.4.0")
+	if err := d.stopContainer(parent); err != nil {
+		return err
+	}
 
-		// Append /dev to mount if old container version is less than v0.4.0
-		// since /dev from host is required inside container to mount a pseudo-tty
-		if v.LessThan(v0_4_0) {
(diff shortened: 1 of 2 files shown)


otavio:
We need to take care of adding this extra mount point in the upgrade procedure, inside the agent. Could you take a look at it?

agent/selfupdater/updater_docker.go
```diff
@@ -112,6 +112,21 @@ func (d *dockerUpdater) CompleteUpdate() error {
...
+		if v.LessThan(v0_7_3) {
+			parent.info.HostConfig.Mounts = []mount.Mount{
+				{
+					Type:   mount.TypeBind,
+					Source: "/etc/resolv.conf",
+					Target: "/etc/resolv.conf",
+				},
+			}
+		}
+
```
jcarrus, line 129: @otavio is this what you are referring to in: https://github.com/shellhub-io/shellhub/pull/1126#issuecomment-882846883

otavio: Yes. The commit log needs to have `agent: ` as a prefix to know which component it is changing.

otavio (changes requested):

agent/selfupdater/updater_docker.go
```diff
@@ -112,6 +112,21 @@ func (d *dockerUpdater) CompleteUpdate() error {
 			}
 		}
 
+		// Append /etc/resolv.conf to mount if old container
+		// version is less than v0..0 since /etc/resolv.conf
```
otavio: the comment mentions v0 but in fact you should mention v0.7.3 here.

jcarrus: Good catch, sorry about that. should be fixed.

otavio (commented):
Please squash the last two commits so we avoid one just to adjust the comment.

jcarrus:
> Please squash the last two commits so we avoid one just to adjust the comment.

Done!

agent/selfupdater/updater_docker.go
```diff
@@ -112,6 +112,21 @@ func (d *dockerUpdater) CompleteUpdate() error {
...
+		// is required inside the container to update host
```
otavio: Here the comment is still wrong.

jcarrus: Ahh, been a while since I've worked on a project where we don't just squash and merge PRs. Should be fixed in all commits that include the change.

otavio (approved):

agent/selfupdater/updater_docker.go
otavio: We try to keep the commit history as clean as possible as it helps in the future. Great job.

otavio:
@gustavosbarreto, please do a final review before merging.

PR #1148, opened 2021-07-29
ssh: fix finish session

eduardoveiga:
when a session is created it's inserted on active session and it connection fails, finish it


ssh/server.go
```diff
@@ -60,6 +60,7 @@ func (s *Server) sessionHandler(session sshserver.Session) {
 	}).Info("Handling session request")
 
 	sess, err := NewSession(session.User(), session)
+	sess.finish() // nolint:errcheck
```
otavio: I don't understand what this fixes. This is finishing it prior logging the error.

otavio (approved):

PR #1188, opened 2021-08-10
api/agent/pkg/ssh/cli: update jwt

Asakiz:
Since the https://github.com/dgrijalva/jwt-go is deprecated
and it may be a security breach, the commit changes to the new
version of the package.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -11,8 +11,10 @@ linters-settings:
     include-go-root: true
     packages:
       - github.com/satori/go.uuid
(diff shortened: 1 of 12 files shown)


otavio (changes requested):
@Asakiz please add to the linter the original package to the import blacklist, so there is no risk we ever use it again.

Asakiz:
> @Asakiz please add to the linter the original package to the import blacklist, so there is no risk we ever use it again.

Sure!

otavio (approved):

PR #1204, opened 2021-08-17
api: use structured logging to improve log readability

henrybarreto:


api/store/mongo/migrations/migration_10.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/sirupsen/logrus"
+
```
otavio, line 7: This seems to be an unrelated change. Please drop it.

otavio (changes requested):
@henrybarreto it seems there are some import-related changes. Are those intentional or due to linter?

api/store/mongo/migrations/migration_10.go
henrybarreto: This behavior is due to IntelliJ auto import. I did not note that.
Fixing it now

otavio: We try to keep the changes at a minimum, thanks for reviewing it.

otavio (approved):

api/store/mongo/migrations/migration_1.go
```diff
@@ -10,12 +10,20 @@ var migration1 = migrate.Migration{
 	Version:     1,
 	Description: "Create the database for the system",
 	Up: func(db *mongo.Database) error {
-		logrus.Info("Applying migration 1 - Up")
+		logrus.WithFields(logrus.Fields{
+			"component": "migration",
+			"version":   1,
+			"action":    "Up",
```
otavio, line 16: Looking at the task description @gustavosbarreto did, he used lowercase for _up_ and _down_. This seems better indeed.

otavio (changes requested):

PR #1210, opened 2021-08-18
api: fix to store the last_login on users collection

Asakiz:


otavio (approved):

otavio:
The change itself looks fine however I wonder if it'd be good to add a test so we ensure this doesn't fail in future.

PR #1223, opened 2021-08-25
SSH server record waitingString in line

### LINKED ISSUE #1222: SSH server connect function waitingString is incomplete
In the SSH service, session.go file, connect function, line 204, there is a code that records the command entered by the user and the result returned. But the recorded information is incomplete, as follows：
image
The input is eeeeee, but 6 lines are recorded, which makes it difficult to know what the original command is through the record.

So I plan to submit a PR, and only record a complete line of information, so that the user input can be restored by recording.
like this:
image

Jsharkc:
fix #1222


ssh/session.go
```diff
@@ -201,17 +201,22 @@ func (s *Session) connect(passwd string, key *rsa.PrivateKey, session sshserver.
 		go func() {
 			buf := make([]byte, 1024)
 			n, err := stdout.Read(buf)
-			waitingString := ""
 			if err == nil {
-				waitingString = string(buf[:n])
 				c.RecordSession(&models.SessionRecorded{
 					UID:     s.UID,
-					Message: waitingString,
+					Message: string(buf[:n]),
```
otavio, line 207: Please, rework the commit log so it explains what we are fixing here and the motivation for the change.

It must follow the pattern:

```
scope: short description

long description
```

For the scope, it seems `ssh` is a good fit ;-)

otavio (changes requested):
Thank you for contributing to the project. Please take a look below...

ssh/session.go
Jsharkc: Thanks, I changed it.

otavio:
@Jsharkc, from a PR point of view, it looks Okay. However, we need to check if it will break our screen recording on the enterprise version or not. 

How are you viewing it? Unfortunately, we don't have the full structure for the playback on the Open Source version, so I wonder how are you using it?

PR #1236, opened 2021-08-30
Restrict user access using public keys

henrybarreto:
First, It was just added a new field, `usernames`, that is a `regexp` expression for all allowed usernames.

```http
POST http://localhost:80/api/sshkeys/public-keys

{"name":"PublicKeyName","usernames": ".*","data":"...","hostname":".*"}
```

After that, I have checked if the `username´ who want to connect match.

### DIFF
--- a/ssh/ssh2ws.go
+++ b/ssh/ssh2ws.go
@@ -58,7 +58,7 @@ func HandlerWebsocket(ws *websocket.Conn) {
 			return
 		}
 
-		if ok, err := apiClient.EvaluateKey(fingerprint, device); !ok || err != nil {
+		if ok, err := apiClient.EvaluateKey(fingerprint, device, parts[0]); !ok || err != nil {
 			if err != nil {
 				fmt.Println(err) //nolint:forbidigo
 			}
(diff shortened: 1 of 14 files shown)


ssh/ssh2ws.go
```diff
@@ -50,6 +50,22 @@ func HandlerWebsocket(ws *websocket.Conn) {
 		}
 
 		key, err := apiClient.GetPublicKey(fingerprint, device.TenantID)
+		// TODO This is just a draft
+		username := strings.Split(user, "@")[0] // username without '@'
+		var contains func(item string, items []string) bool
+		contains = func(item string, items []string) bool { // TODO Finding a function to check
```
henrybarreto, line 56: I guess that just a `for` could be more preformat, but I used recursion to avoid state variable.

ssh/ssh2ws.go
```diff
@@ -50,6 +50,22 @@ func HandlerWebsocket(ws *websocket.Conn) {
 		}
 
 		key, err := apiClient.GetPublicKey(fingerprint, device.TenantID)
+		// TODO This is just a draft
+		username := strings.Split(user, "@")[0] // username without '@'
```
henrybarreto, line 54: Here, I thought It could check if `username` is equal to `*`, so all `usernames` would be enabled.

leonardojoao (approved):

gustavosbarreto (approved):

PR #1239, opened 2021-08-31
Setting geoip as a optional feature

henrybarreto:


api/routes/auth.go
```diff
@@ -70,9 +71,11 @@ func (h *Handler) AuthDevice(c apicontext.Context) error {
 		return err
 	}
 
-	err = h.service.SetDevicePosition(c.Ctx(), models.UID(res.UID), ip)
-	if err != nil {
-		return err
+	if feature.IsEnableByEnv("MAXMIND_LICENSE") {
```
henrybarreto, line 74: Is this logic here in router could be a problem?

pkg/geoip/geolite2_null.go
```diff
@@ -0,0 +1,38 @@
+// Package geoip helps in geolocation operations.
+package geoip
+
+import (
+	"io"
+	"net"
+)
+
+// nullGeoLite2 is a structure what stores a geoIp2Reader to a GeoIp2 database.
+type nullGeoLite2 struct{}
```
otavio, line 10: The interface should be `nullGeoIp` as `GeoLite2` refer to the implementation.

I'd have a `NewNullGeoIp()` or `NewNull()` for it.

otavio (changes requested):

PR #1244, opened 2021-09-02
pkg, api: update namespace and user delete

noreng-jg:


pkg/api/client/client_public.go
```diff
@@ -33,6 +33,20 @@ type publicAPI interface {
 	NewReverseListener(token string) (*revdial.Listener, error)
 	AuthPublicKey(req *models.PublicKeyAuthRequest, token string) (*models.PublicKeyAuthResponse, error)
 	ReportUsage(ur *models.UsageRecord, billingURL string) (int, error)
+	ReportDelete(ns *models.Namespace, billingURL string) (int, error)
+}
+
+func (c *client) ReportDelete(ns *models.Namespace, billingURL string) (int, error) {
+	res, _, errs := c.http.Delete(fmt.Sprintf("http://%s:8080/api/billing/namespace-subscription", billingURL)).Send(struct {
```
otavio, line 40: This fixed URL is likely to cause errors. We need to use this as a global setting.

api/store/mongo/user_store.go
```diff
@@ -297,30 +297,37 @@ func (s *Store) UserDelete(ctx context.Context, id string) error {
...
 	cursor, err := s.db.Collection("namespaces").Find(ctx, bson.M{"members": id}, findOptions)
 	if err != nil {
-		return fromMongoError(err)
+		return nil, fromMongoError(err)
 	}
 	defer cursor.Close(ctx)
 
+	namespaces := make(map[string][]*models.Namespace, 2)
+	nssOwner := make([]*models.Namespace, 0)
+	nssMember := make([]*models.Namespace, 0)
```
otavio, line 314:
```suggestion
	namespacesMap := make(map[string][]*models.Namespace, 2)
	ownerList := make([]*models.Namespace, 0)
	memberList := make([]*models.Namespace, 0)
```

api/store/mongo/user_store.go
```diff
@@ -297,30 +297,37 @@ func (s *Store) UserDelete(ctx context.Context, id string) error {
...
 		if namespace.Owner != id {
-			if _, err := s.NamespaceRemoveMember(ctx, namespace.TenantID, id); err != nil {
-				return fromMongoError(err)
-			}
+			nssMember = append(nssMember, namespace)
 		} else {
-			if err := s.NamespaceDelete(ctx, namespace.TenantID); err != nil {
-				return fromMongoError(err)
-			}
+			nssOwner = append(nssOwner, namespace)
```
otavio, line 325: This brings up a topic. Can a namespace to have multiple owners?

otavio (changes requested):

pkg/requests/helpers.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"errors"
+
+	"github.com/shellhub-io/shellhub/pkg/models"
+)
+
+var ErrReport = errors.New("report error")
+
+func HasBillingInstance(ns *models.Namespace) bool {
+	if ns == nil || ns.Billing == nil || !ns.Billing.Active || ns.MaxDevices != -1 {
```
otavio, line 12:
```suggestion
	if ns == nil || ns.Billing == nil || !ns.Billing.Active || ns.MaxDevices == -1 {
```

otavio: Isn't the logic wrong?

otavio (changes requested):

PR #1258, opened 2021-09-10
gateway: update base image to 1.19.9.1-0-alpine

otavio:


otavio:
I will created an unified PR which does more Docker related upgrades.

PR #1259, opened 2021-09-11
Upgrade golang to 1.17.1

otavio:


otavio:
I will created an unified PR which does more Docker related upgrades.

PR #1286, opened 2021-09-21
api: change HTTP response from string to empty body

henrybarreto:


otavio (changes requested):
The commit log might receive more detailed description of what this is fixing.

otavio (changes requested):
The commit log is totally misleading as it doesn't explain WHY this change is need and WHAT it intends to achieve.

otavio (approved):

PR #1287, opened 2021-09-21
Tag settings api

Asakiz:


api/routes/device.go
```diff
@@ -222,23 +222,19 @@ func (h *Handler) DeleteTag(c apicontext.Context) error {
...
+	if err := h.service.RenameTag(c.Ctx(), c.Param("name"), req.Name); err != nil {
 		switch err {
 		case services.ErrUnauthorized:
 			return c.NoContent(http.StatusForbidden)
 		case services.ErrInvalidFormat:
 			return c.NoContent(http.StatusBadRequest)
-		case services.ErrMaxTagReached:
-			return c.NoContent(http.StatusForbidden)
-		case services.ErrNotFound:
-			return c.NoContent(http.StatusNotFound)
```
otavio, line 241: We are running asynchronously so we can have an `ErrNotFound` if someone else renamed at the same time. We don´t know which one wins the race but one of them will receive a not found.

api/store/mongo/device_store.go
```diff
@@ -450,8 +450,8 @@ func (s *Store) DeviceDeleteTag(ctx context.Context, uid models.UID, tag string)
 	return err
 }
 
-func (s *Store) DeviceRenameTag(ctx context.Context, uid models.UID, currentTagName string, newTagName string) error {
-	_, err := s.db.Collection("devices").UpdateOne(ctx, bson.M{"uid": uid, "tags": currentTagName}, bson.M{"$set": bson.M{"tags.$": newTagName}})
+func (s *Store) DeviceRenameTag(ctx context.Context, currentTagName string, newTagName string) error {
+	_, err := s.db.Collection("devices").UpdateMany(ctx, bson.M{"tags": currentTagName}, bson.M{"$set": bson.M{"tags.$": newTagName}})
```
otavio, line 454: We need to guarantee it needs to be atomic. This needs to be done using a transaction.

otavio (changes requested):
The are some small things to be reworked, also the commit log needs to be more detailed as it doesn't explain what it tries to achieve and what was wrong with the previous behavior.

otavio (dismissed):

otavio (dismissed):

otavio (approved):

otavio (approved):

PR #1288, opened 2021-09-21
Topic/organize store tests

Asakiz:


otavio (changes requested):
The change looks great but it seems it should be done in a single commit and a detailed commit log explaining WHY you are doing it.

otavio (approved):

PR #1289, opened 2021-09-22
api: excludes the characters /@&: for create and update tag name

Asakiz:
Previously, was possible to create a name for a tag with /@&:
this commit is to disallow these types of characters when creating
the tag name.

Signed-off-by: Juan <juan.rios@ossystems.com.br>

### DIFF
--- a/api/services/device.go
+++ b/api/services/device.go
@@ -247,8 +247,8 @@ func (s *service) SetDevicePosition(ctx context.Context, uid models.UID, ip stri
 }
 
 func (s *service) CreateTag(ctx context.Context, uid models.UID, name string) error {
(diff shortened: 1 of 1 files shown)


otavio (changes requested):
The change looks good. However, it seems the validation logic is duplicated. You could rework it to use a private method to validate it and avoid duplicating the checks.

api/services/device.go
```diff
@@ -345,3 +345,11 @@ func contains(s []string, name string) bool {
 
 	return false
 }
+
+func validateName(tag string) error {
```
otavio, line 349: validateTagName.

otavio (changes requested):

otavio (approved):

PR #1290, opened 2021-09-22
api: pattern golang's JSON tags to snake case

henrybarreto:


otavio (dismissed):
Please extend the commit log mentioning which fields has been changed.

otavio:
@leonardojoao I believe this has an impact on UI. Could you check that?

otavio (approved):

otavio:
@henrybarreto, from API point of view it looks good. We just need @leonardojoao or @noreng-jg to check the UI so we do both changes on same PR.

leonardojoao (approved):

noreng-jg (approved):

PR #1291, opened 2021-09-22
api: change null JSON's responses to empty body

henrybarreto:
The `echo` returns a `200` HTTP code when the routing function return a nil,
but it is not explicit for the dev, so I have guessed, for the sake of
simplicity, change it to an empty body with the `200` status.

Another point is the responses which return a JSON, but with `null` inside the
body. I believe it could be a simply `no body` too with the original
status.


otavio (approved):
I totally support this change. We just need to double-check if old versions of the agent will keep working.

@gustavosbarreto can you take a look?

PR #1301, opened 2021-09-27
openapi: add missing body to session record and `/api/` in front cloud routes

henrybarreto:


openapi/spec/resources/sessions/record_del.yaml
```diff
@@ -15,12 +15,24 @@ delete:
     '200':
       $ref: ../../shared/responses/OkResponse.yaml
     '400':
-      $ref: ../../shared/responses/BadRequestResponse.yaml
+      description: Bad Request
+      content:
+        application/json:
+          schema:
+            type: string
```
otavio, line 22: Couldn't this to be added to the shared response file?

otavio (changes requested):

openapi/spec/resources/sessions/record_del.yaml
henrybarreto: I guess no, because if I try to make something like snippet below, the code generate is invalid.

```openapi
    '403':
      description: Forbidden
      $ref: ../../shared/responses/JSONStringResponse.yaml
```
Screenshot 2021-09-27 at 16-51-34 Screenshot

```openapi
# JSONStringResponse.yaml
content:
  application/json:
    schema:
      type: string
```

otavio: It should use  a shared definition. @gustavosbarreto what do you think?

PR #1304, opened 2021-09-27
api: split logging middleware to own file

henrybarreto:


henrybarreto:
> Can the package to be called `middleware`?

`apimiddleware` could be valid in that case?

PR #1315, opened 2021-09-30
shellhub: change RSA SSH key generation to AES256

henrybarreto:
Currently, the Shellhub has used RSA SHA1 for SSH connection. This commit
change the hashing algorithm from the default, SHA1, to AES256.

On the update of OpenSSH to version `8.8`, the RSA SHA1 key was deprecated,
making connection from clients from this version needs an extra param to
connect.

```
This release disables RSA signatures using the SHA-1 hash algorithm
by default. This change has been made as the SHA-1 hash algorithm is
cryptographically broken, and it is possible to create chosen-prefix
hash collisions for <USD$50K [1]
```

https://www.openssh.com/txt/release-8.8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ api_public_key:
 
 # Generate required private key for ssh service
 ssh_private_key:
-	@$(KEYGEN) genrsa -out ssh_private_key 2048
+	@$(KEYGEN) genrsa -aes256 -out ssh_private_key 2048
 
 .PHONY: keygen
 # Generate required keys


otavio (approved):

henrybarreto:
As the Change log  from `OpenSHH 8.8 ` says, and it took me a while to understand, `ssh-rsa` using SHA 1 was deprecated at once, but the key compatibilities haven't changed.

> This release disables RSA signatures using the SHA-1 hash algorithm
  by default. This change has been made as the SHA-1 hash algorithm is
  cryptographically broken, and it is possible to create chosen-prefix
  hash collisions for <USD$50K [1]

I guess it means that we need our server to have support for RSA SHA-2, but the package we use, or its decencies, does not implement.

- https://github.com/golang/go/issues/37278
- https://github.com/golang/go/issues/39885

If the system implements an old version of the protocol, extra configuration would be needed for it to work: 
```bash -oHostKeyAlgorithms=+ssh-rsa```

**As an example**
```bash
ssh -oHostKeyAlgorithms=+ssh-rsa henry@henrysnamespace.d0-94-66-b4-25-bc@localhost
```

> Incompatibility is more likely when connecting to older SSH
  implementations that have not been upgraded or have not closely tracked
  improvements in the SSH protocol. **For these cases, it may be necessary
  to selectively re-enable RSA/SHA1** to allow connection and/or user
  authentication via the HostkeyAlgorithms and PubkeyAcceptedAlgorithms
  options. For example, the following stanza in ~/.ssh/config will enable
  RSA/SHA1 for host and user authentication for a single destination host:

PR #1336, opened 2021-10-08
Fix some details by code analysis

henrybarreto:


henrybarreto:
Trying to fix some details on this PR, I found a missed command what causes the migrations tests brake.

https://github.com/shellhub-io/shellhub/blob/b58a00572fefc8ddef258b42b75c93cc4c3b9b7a/pkg/models/device.go#L19

Issue #1203, opened 2021-08-17
[feature] Namespace Member Roles


eduardoveiga:
Today we had owner that can do anything, including add another members, change settings and delete information
and regular members that can only connect on devices

this issue is about to create some roles to members to define new privileges

eduardoveiga:
Today we had owner that can do anything, including add another members, change settings and delete information 
and regular members that can only connect on devices 

this issue is about to create some roles to members to define new privileges

leonardojoao:
-Migration:
- [ ] Modify namespace; modify the member list to support an object list. Ex: ``[{id: user_id, type: owner|admin|operator|observer}...]``.

-API:
- [ ] Modify the add member in the namespace to support the access type;
- [ ] Add middleware;
- [ ] Can only add/remove down.

-UI:
- [ ] Modify the list to show the access type;
- [ ] Add the user related to the type of access;
- [ ] Block actions according to table

gustavosbarreto:
Implemented in #1400

PR #1400, opened 2021-10-25
api, cli, ui: add permission's check

henrybarreto:
Related to: https://github.com/shellhub-io/shellhub/issues/1203

Before this PR, Shellhub's already had an initial check to owner.

This PR adds a new field to namespace, members, what contains a list of member. A member has tree properties: ID, username and type. 

To add a member in the database, only two properties are needing: ID and type. 

The username may not put inside the database because these properties can be changed and could create reference problems. Username is setting in member through another service when requested.

Type or access type, define what a member can do in Shellhub, and it is based on an internal packager called authorizer.

Authorizer's package has functions to check if an action is allowing for a member's type. The package define these members' type too. These are the type that the Shellhub uses.

Guard is another piece of this feature. It is an API's internal package used for the internal permission's check logic.

The permission's check happens before the data reach the service wanted; in the route. If it is an allowed action, the service will run, else it will return an HTTP's forbidden response.

### DIFF
--- a/api/services/auth.go
+++ b/api/services/auth.go
@@ -331,21 +331,16 @@ func (s *service) AuthUserInfo(ctx context.Context, username, tenant, token stri
 		return nil, err
 	}
 
-	namespace, err := s.store.NamespaceGet(ctx, tenant)
-	if err != nil && tenant != "" {
-		if err == store.ErrNoDocuments {
-			return nil, ErrUnauthorized
-		}
+	namespace, _ := s.store.NamespaceGet(ctx, tenant)
 
-		return nil, err
-	}
+	userType := ""
+	if namespace != nil {
+		for _, member := range namespace.Members {
+			if member.ID == user.ID {
+				userType = member.Type
 
-	var userType string
-	for _, member := range namespace.Members {
-		if member.ID == user.ID {
-			userType = member.Type
-
-			break
+				break
+			}
 		}
 	}
 
--- a/api/services/auth_test.go
+++ b/api/services/auth_test.go
@@ -269,16 +269,6 @@ func TestAuthUserInfo(t *testing.T) {
 			expected: Expected{nil, Err},
 		},
 		{
-			description: "Fails to find the namespace",
-			username:    "user",
-			tenantID:    "xxxxx",
-			requiredMocks: func() {
-				mock.On("UserGetByUsername", ctx, "user").Return(user, nil).Once()
-				mock.On("NamespaceGet", ctx, "xxxxx").Return(nil, Err).Once()
-			},
-			expected: Expected{nil, Err},
-		},
-		{
 			description: "Successful auth login",
 			username:    "user",
 			tenantID:    namespace.TenantID,


gustavosbarreto:
@all-contributors please add @henrybarreto as code

ui/src/components/filter/permission.js
```diff
@@ -1,6 +1,10 @@
-export default (accessType, action, permissions) => {
-  if (permissions[accessType].includes(action)) {
-    return true;
-  }
-  return false;
-};
+// import { authorizer } from '../../authorizer';
```
otavio, line 1: Uh?

ui/src/components/filter/permission.js
```diff
@@ -1,6 +1,10 @@
...
+/**
+ *
+ * @param accessType
+ * @param action
+ * @param permissions
+ * @returns {boolean}
+ */
+// eslint-disable-next-line max-len
```
otavio, line 9: Could it to be avoided?

ui/src/components/filter/permission.js
```diff
@@ -1,6 +1,10 @@
...
+export default (accessType, action, permissions) => !!permissions[accessType] && permissions[accessType].includes(action);
```
otavio, line 10: Why double not (`!!`)?

ui/tests/unit/components/filter/Permission.spec.js
```diff
@@ -0,0 +1,100 @@
...
+
+const casesInvalid = [
+  [false, -1, -1],
+  [false, 'valid', -1],
+  [false, 'valid', 26],
+  [false, -1, 'valid'],
+  [false, 5, 'valid'],
+];
+
+describe('Permission', () => {
```
otavio, line 92: Here you test a custom permission list; I fail to see why it uses the `authorizer.permissions` here. For me, it seems it could just validate three cases:
1. valid
2. invalid
3. non-existent

If those works, considering `authorizer.permissions` has the proper format, it'll work.

otavio (changes requested):

ui/src/components/filter/permission.js
henrybarreto: It "converts" a possible undefined in `permissions[accessType]` to false.

ui/src/components/filter/permission.js
henrybarreto: Yes, it does, however…

The ESLInt rule, arrow-body-style, request to the function, when as simple as that, to be in a line, but, when put in line, it breaks another ESLint rule, max-line.

Aiming for the simplicity, I have chosen to disable the max-line instead doing a return true and false block.

ui/src/components/filter/permission.js
henrybarreto: I've forgotten to remove this JSDoc and this import :D

ui/tests/unit/components/filter/Permission.spec.js
henrybarreto: I've thought that if the local permission, "mocked", is not equal permission used by the UI, the tests should break.

otavio: In my understanding, the reason to use a mock is to not depend on external values so I think it could exercise the needed cases only. @gustavosbarreto?

api/pkg/guard/guard.go
```diff
@@ -0,0 +1,123 @@
...
+
+	namespaceUserActive, err := s.NamespaceGet(ctx, tenantID)
+	if err != nil || err == store.ErrNoDocuments {
+		return "", false
+	}
+
+	var userType string
+	for _, member := range namespaceUserActive.Members {
+		if member.ID == user.ID {
+			userType = member.Type
```
otavio, line 27: It seems it is lacking a `break` here.

api/pkg/guard/guard.go
```diff
@@ -0,0 +1,123 @@
...
```
otavio: and here as well.

api/pkg/guard/guard.go
```diff
@@ -0,0 +1,123 @@
...
+
+	namespace, err := s.NamespaceGet(ctx, tenantID)
+	if err != nil || err == store.ErrNoDocuments {
+		return false
+	}
+
+	var userType string
+	for _, member := range namespace.Members {
+		if member.ID == user.ID {
+			userType = member.Type
```
otavio: so does here.

otavio (changes requested):

bin/add-user-namespace
```diff
@@ -1,8 +1,9 @@
...
 
-[ $# -ne 2 ] && echo "Usage: $0 <username> <namespace>" && exit 1
+[ $# -ne 3 ] && echo "Usage: $0 <username> <namespace> <type>" && exit 1
 
 USERNAME=$1
 NAMESPACE=$2
+TYPE=$3
 
-docker-compose exec cli ./cli add-user-namespace $USERNAME $NAMESPACE
+docker-compose exec cli ./cli add-user-namespace $USERNAME $NAMESPACE $TYPE
```
otavio, line 9: Newline here.

otavio (dismissed):

otavio (changes requested):

PR #1442, opened 2021-11-12
ui: improve route errors return

noreng-jg:


otavio (approved):

ui/src/components/app_bar/notification/Notification.vue
```diff
@@ -9,7 +9,7 @@
         :content="showNumberNotifications"
         :value="showNumberNotifications"
         overlap
-        offset-x="20"
+        offset-x="30"
```
leonardojoao, line 12: Don't you think it's better to set this value according to the amount of notifications?

ui/src/components/billing/BillingDialogPaymentMethod.vue
```diff
@@ -208,8 +224,12 @@ export default {
           });
           this.$store.dispatch('snackbar/showSnackbarSuccessAction', this.$success.updateSubscription);
           this.$emit('update');
-        } catch {
+        } catch (error) {
           this.$store.dispatch('snackbar/showSnackbarErrorAction', this.$errors.snackbar.updateSubscription);
+          const { status } = error.response;
```
leonardojoao, line 229: You can insert a blank line. Same way as you follow in the code above.

PR #1484, opened 2021-12-03
api: update store device toplist to a more effective query

noreng-jg:
The previous query related to top usage of devices had some performance issues,
this commit updates the previous store function use to use a faster aggregation.

Signed-off-by: Vagner S. Nornberg <vagner.nornberg@ossystems.com.br>

### DIFF
--- a/api/store/mongo/device_store.go
+++ b/api/store/mongo/device_store.go
@@ -259,54 +259,24 @@ func (s *Store) DeviceUpdateStatus(ctx context.Context, uid models.UID, status s
 	return nil
 }
 
-func (s *Store) DeviceListByUsage(ctx context.Context, tenant string) ([]models.Device, error) {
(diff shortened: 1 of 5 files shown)


api/store/mongo/device_store.go
```diff
@@ -259,78 +259,53 @@ func (s *Store) DeviceUpdateStatus(ctx context.Context, uid models.UID, status s
...
+	cursor, err := s.db.Collection("sessions").Aggregate(ctx, query)
 	if err != nil {
-		return devices, fromMongoError(err)
+		return uids, fromMongoError(err)
 	}
 
 	for cursor.Next(ctx) {
-		device := new(models.Device)
-		err = cursor.Decode(&device)
+		var dev *Qry
```
gustavosbarreto, line 299: There is no need to create a type in this case, you can use a interface. eg:
```go
var dev map[string]interface{}

cursor.Decode(&dev)
....
uids = append(uids, models.UID(dev["_id"].(string)))
.....
```

PR #1488, opened 2021-12-03
api: add migration to set last_login property to created_at

henrybarreto:


api/store/mongo/migrations/migration_38.go
```diff
@@ -0,0 +1,53 @@
...
+				{
+					{"$match", bson.D{
+						{"created_at", zeroTime},
+					}},
+				},
+				{
+					{"$replaceRoot", bson.D{{"newRoot", bson.M{"$mergeObjects": bson.A{"$$ROOT", bson.M{"created_at": "$last_login"}}}}}},
+				},
+				{
+					{"$out", "users"},
```
henrybarreto, line 34: This line terminate the mongo's pipeline saying to write modified document. However, there is another staged, `$merge`, that could be used in the same way, in this case.

https://docs.mongodb.com/manual/reference/operator/aggregation/out/#mongodb-pipeline-pipe.-out

api/store/mongo/migrations/migration_38_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	migrations := GenerateMigrations()[37:38]
+
+	timeNow := time.Now().UTC()
+	convertDate := func(t time.Time) string {
+		return t.Format("2006-01-02 15:04:05.000")
+	}
+
+	userNoCreated := models.User{
+		ID:         "userNoCreatedID",
```
otavio, line 38: `userNoCreatedAt` in all bellow cases; also the other example would be `userWithCreatedAt`.

api/store/mongo/migrations/migration_38_test.go
```diff
@@ -0,0 +1,112 @@
...
+		UserPassword: models.UserPassword{
+			Password: "",
+		},
+	}
+	userWithCreated := models.User{
+		ID:         "userWithCreatedID",
+		Namespaces: 0,
+		Confirmed:  false,
+		CreatedAt:  timeNow,
+		LastLogin:  timeNow,
```
otavio, line 57: This case should use a date in the past so your test ensures it is not changing it.

otavio (changes requested):

api/store/mongo/migrations/migration_38_test.go
```diff
@@ -0,0 +1,115 @@
...
+			description: "Executes migration when user's created_at property is empty",
+			Test: func(t *testing.T) {
+				t.Helper()
+
+				var userMigrated *models.User
+				err = db.Client().Database("test").Collection("users").FindOne(context.TODO(), bson.D{{"username", userNoCreatedAt.Username}}).Decode(&userMigrated)
+				assert.NoError(t, err)
+				assert.Equal(t,
+					Expected{CreatedAt: convertDate(userNoCreatedAt.LastLogin), LastLogin: convertDate(userNoCreatedAt.LastLogin)},
+					Expected{CreatedAt: convertDate(userMigrated.CreatedAt), LastLogin: convertDate(userMigrated.LastLogin)},
```
otavio, line 91: What happens if user didn't login yet?

otavio (changes requested):

api/store/mongo/migrations/migration_38_test.go
henrybarreto: Theoretically, if time is "zero"for both, it copies the zero. 

I'm working on a better query for the aggregation.

api/store/mongo/migrations/migration_38.go
```diff
@@ -0,0 +1,56 @@
...
+		}).Info("Applying migration")
+		zeroTime := time.Time{}.UTC()
+		_, err := db.Collection("users").Aggregate(context.TODO(),
+			mongo.Pipeline{
+				{
+					{"$match", bson.D{
+						{"$and", bson.A{
+							bson.M{"created_at": bson.M{"$eq": zeroTime}},
+							bson.M{"last_login": bson.M{"$ne": zeroTime}},
+						}},
```
henrybarreto, line 30: That seems preciser

PR #1503, opened 2021-12-06
ssh: update message when user can't access device

leonardojoao:
This commit improves the error message on the terminal when the user has more
than 3 devices and is not enabled for billing.

Signed-off-by: Leonardo R.S. Joao <leonardo.joao@ossystems.com.br>

### DIFF
--- a/ssh/session.go
+++ b/ssh/session.go
@@ -24,7 +24,7 @@ import (
 
 var (
 	ErrInvalidSessionTarget = errors.New("invalid session target")
-	ErrBillingBlock         = errors.New("reached the device limit, update to premium or choose up to 3 devices")
+	ErrBillingBlock         = errors.New("you reached the free device limit, please upgrade to the premium plan or choose up to 3 devices")
 )
 
 type Session struct {


ssh/session.go
```diff
@@ -24,7 +24,7 @@ import (
 
 var (
 	ErrInvalidSessionTarget = errors.New("invalid session target")
-	ErrBillingBlock         = errors.New("reached the device limit, update to premium or choose up to 3 devices")
+	ErrBillingBlock         = errors.New("you reached the device limit, you need to update the premium plan or choose up to 3 devices")
```
otavio, line 27:
```suggestion
	ErrBillingBlock         = errors.New("You reached the free device limit, please upgrade to the premium plan or choose up to 3 devices.")
```

otavio (changes requested):

ssh/session.go
leonardojoao: @otavio golang lint doesn't allow you to add capital letters or dot. 
What to do in this case?

otavio (approved):

PR #1589, opened 2022-01-03
api: add worker to delete session records

henrybarreto:


otavio (changes requested):
Where do we define the time window to preserve? For example, to remove sessions older than 90 days.

henrybarreto:
That first implementation was intended to be an env variable, but the next one will empower each namespace to set it own time interval at the settings panel.

otavio:
Sure, it makes sense. But it is not defined in this PR. Is it?

henrybarreto:
> Sure, it makes sense. But it is not defined in this PR. Is it?

Yes, that was I've intended, but it is not well documented at PR.

PR #1592, opened 2022-01-06
api, openapi: fix and improvements to tag system

henrybarreto:


api/store/mongo/utils.go
```diff
@@ -85,6 +85,8 @@ func buildFilterQuery(filters []models.Filter) ([]bson.M, error) {
 				}
 
 				property = bson.M{"$gt": value}
+			case "all":
```
otavio, line 88: As `all` has a special meaning, we need to forbid a tag named _all_ to be added.

otavio (changes requested):

PR #1620, opened 2022-01-11
ui: use vuex to set dark mode

leonardojoao:


ui/src/store/modules/layout.js
```diff
@@ -1,25 +1,36 @@
 import Vue from 'vue';
+// import Vuetify from '../../plugins/vuetify';
```
otavio, line 2: Commented code; please drop.

otavio (changes requested):

PR #1711, opened 2022-02-15
Topic/add tags to public keys

henrybarreto:


pkg/models/publickey.go
```diff
@@ -4,7 +4,7 @@ import (
 	"regexp"
 	"time"
 
-	validator "gopkg.in/go-playground/validator.v9"
+	"gopkg.in/go-playground/validator.v9"
```
otavio, line 7: This is not related to this PR and could be split to allow merging it sooner.

api/services/errors.go
```diff
@@ -4,27 +4,44 @@ import (
 	"errors"
 )
 
+// Public Key's errors
+var (
+	ErrPublicKeyNotFound = errors.New("public key not found")
+)
```
otavio, line 10: This is not related to this PR and could be split to allow merging it sooner.

api/services/device_tags.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/shellhub-io/shellhub/pkg/validator"
 )
 
-type DeviceTags interface {
+type DeviceTagsService interface {
```
otavio, line 10: This is not related to this PR and could be split to allow merging it sooner.

otavio (approved):

PR #1737, opened 2022-02-25
api: fix public key loop with wrong range definition

henrybarreto:


otavio (changes requested):
We need tests :-)

henrybarreto:
> We need tests :-)

Test fixed

otavio (approved):

PR #1834, opened 2022-04-01
api, pkg: refactor guard package

henrybarreto:
guard and authorizer were splitting the same responsibility, help ShellHub
to control the roles.

In a package, authorizer, were all role, actions and permissions, and another,
guard, were all logic to validate if a role has permission to execute an action,
what obligated the programmer change at two district places when he/she wants
to add a permission, action or role.

The old authorizer package was not remove, but marked as "deprecated", so we
can remove its call at pieces.

Some comments on guard were improved too, giving more precise description 
and code usage examples.


api/pkg/guard/guard.go
```diff
@@ -1,56 +1,129 @@
...
 
-// CheckRole checks if a member from a namespace can act over other with a specif role.
-func CheckRole(roleActive, rolePassive string) bool {
-	return authorizer.CheckRole(roleActive, rolePassive)
+// GetRoleCode converts a models.Member's role string to a role code. If the role is not found in Roles, it returns -1.
+func GetRoleCode(role string) int {
+	code, ok := Roles[role]
+	if !ok {
+		// return -1 when member role is not valid.
+		return -1
```
otavio, line 66: As we are reworking this, we might consider using a type, even if it is an int behind the scenes, to have a more meaningful name when checking for it.

otavio (changes requested):

api/pkg/guard/guard.go
henrybarreto: Okay. It is a good an idea.

What do you think about a `InvalidRole` to map to this `-1`?

api/pkg/guard/guard.go
```diff
@@ -62,29 +61,30 @@ func CheckMember(namespace *models.Namespace, id string) (*models.Member, bool)
...
 
-// GetRoleCode converts a models.Member's role string to a role code. If the role is not found in Roles, it returns -1.
+// GetRoleCode converts a models.Member's role string to a role code. If the role is not found in Roles, it returns InvalidRole.
 func GetRoleCode(role string) int {
 	code, ok := Roles[role]
 	if !ok {
-		// return -1 when member role is not valid.
-		return -1
+		// return InvalidRole when member's role is not valid.
+		return InvalidRole
```
henrybarreto, line 69: Using the constant `InvalidRole` to represent the value when the member's role does not exist.

otavio: Seems better

api/pkg/guard/guard.go
```diff
@@ -1,56 +1,133 @@
...
+
+const (
+	RoleObserver      = "observer"
+	RoleOperator      = "operator"
+	RoleAdministrator = "administrator"
+	RoleOwner         = "owner"
+)
+
+const (
+	// RoleObserverCode is a role code for observer.
```
otavio, line 19: Why the invalid isn't defined here with the other codes?

otavio (changes requested):

api/pkg/guard/guard.go
henrybarreto: https://github.com/shellhub-io/shellhub/pull/1834/commits/4ccf4a4e64175d6b9189909d289ccd862652be9b#diff-471917d73591cfd1c57419b9cc8ed82462afe06ae236dc6587c899758fcb8097R15

I've added the code out of iota because of its value: `-1`.

otavio (approved):

PR #1854, opened 2022-04-06
api: set HandleStatusResponse function as public

henrybarreto:


api/services/service.go
```diff
@@ -42,7 +42,7 @@ func NewService(store store.Store, privKey *rsa.PrivateKey, pubKey *rsa.PublicKe
 	return &service{store, privKey, pubKey, cache, c, l}
 }
 
-func handleStatusResponse(status int) error {
+func HandleStatusResponse(status int) error {
```
otavio, line 45: Please add a FIXME comment.

otavio (changes requested):

api/services/service.go
henrybarreto: Okay

api/services/service.go
```diff
@@ -42,7 +42,10 @@ func NewService(store store.Store, privKey *rsa.PrivateKey, pubKey *rsa.PublicKe
 	return &service{store, privKey, pubKey, cache, c, l}
 }
 
-func handleStatusResponse(status int) error {
+// FIXME: private function.
```
henrybarreto, line 45: Done

otavio (approved):

PR #1858, opened 2022-04-08
ssh, pkg: add connection source tracker

noreng-jg:
It enables connection tracking of the terminal origin usage, this will allow us
to a better source identification of new sessions.


ssh/session.go
```diff
@@ -110,6 +111,15 @@ func NewSession(target string, session sshserver.Session) (*Session, error) {
 	s.Target = uid
 	s.Lookup = lookup
 
+	var connectionSource string
+
+	if _, ok := IsWs[uid]; ok {
+		connectionSource = "webterminal"
```
otavio, line 117: please use 'web'

ssh/session.go
```diff
@@ -110,6 +111,15 @@ func NewSession(target string, session sshserver.Session) (*Session, error) {
...
+		delete(IsWs, uid)
+	} else {
+		connectionSource = "ptyclient"
```
otavio, line 120: please use 'pty'

otavio: also, please use `const` for both so it is easier to find it in code and there is no risk of typos.

ssh/ssh2ws.go
```diff
@@ -34,21 +34,21 @@ func HandlerWebsocket(ws *websocket.Conn) {
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(), //nolint:gosec
 	}
 
-	if fingerprint != "" && signature != "" { //nolint:nestif
-		parts := strings.SplitN(user, "@", 2)
-		if len(parts) != 2 {
-			ws.Close()
+	parts := strings.SplitN(user, "@", 2)
```
otavio, line 37: Please add a comment explaining why two parts are required.

otavio (approved):

PR #1868, opened 2022-04-12
github: block dependabot sass minor versions updates

noreng-jg:


otavio (dismissed):
Just confirm if this is upgrade type whitelist or blacklist.

PR #1894, opened 2022-05-06
openapi: improve information about device and tags routes

henrybarreto:
I would like a little help to solve the To-dos commented on the code


openapi/spec/resources/devices/auth.yaml
```diff
@@ -15,47 +15,110 @@ post:
...
+                  type: string
+                  example: latest
+                arch:
+                  description: Device's OS arch
+                  type: string
+                  # TODO: add more compatible arches.
+                  enum:
+                    - x86_64
+                    - amd64
+                    - arm64
```
otavio, line 51: I'd not validate it, at this moment. We support anything which runs Go.

otavio (dismissed):
Auth and device model are quite close ... is this intended?

openapi/spec/resources/devices/auth.yaml
henrybarreto: Perfect!

henrybarreto:
> Auth and device model are quite close ... is this intended?

Yes. That auth is the device's auth used by agent.

PR #1918, opened 2022-05-18
api,pkg: add Swaggon comments and doc to device and stats model

henrybarreto:
That is a proposal to use [swaggon](https://github.com/swaggo/swag) to generate our OpenAPI/Swagger AOS2 documentation from declarative comments.

The tool, basically, follows the code looking for the comments and generating the OpenAPI/Swagger AOS2 into `docs` folder. Unfortunately, the tool only generate AOS2 compatible specification, what limit a bit.

### why

Using the declarative comments, the specification is placed next to the code it documents, helping to avoid errors and facilitating the modification.

### try yourself
Install the tool:
```
go install github.com/swaggo/swag/cmd/swag@latest
```

From the `api` directory, run:
```
swag init -g server.go -d ./,../pkg/models
```
That command will generate or update the `docs` folders.

### preview
To preview the docs, you can use the `@redocly/openapi-cli` to it:
```
openapi preview-docs docs/swagger.json
```

Notice that the `swaggon` get metadata from other tags like `validate` from `go-playground/validator`, what helps the specification.


pkg/models/device.go
```diff
@@ -7,20 +7,31 @@ import (
 )
 
 type Device struct {
```
henrybarreto, line 9: Changes here reflects on the OpenAPI/Swagger final generated file.

PR #1930, opened 2022-05-25
api,pkg: add custom validator to echo and change password validation

henrybarreto:
This is the first step to apply new models to routes and remove the database models from it. With the new models, we'll preserve information about what has been saved on the database and empower local validation for that new models.


henrybarreto:
For that, I had to change the tests on services, removing validation ones and focusing only on the logic. I also created two extra packages, bodies and rules, to define the models to requests' bodies and the rules to validate our Path's parameters.

henrybarreto:
I removed the packaga 'rule' by passing the validation  for path parameters  to 'request' package inside 'pkg/api'.

PR #1950, opened 2022-06-09
 api,pkg: add API models to auth's routes

henrybarreto:


otavio:
We need to check if this works with 0.3.2 agent version so we keep old embedded devices working.

Issue #1959, opened 2022-06-19
agent panics when `!isPty` and `len(session.Command()) == 0`


afriza:
### Panic detail
```
Sun Jun 19 17:36:58 2022 daemon.err sh[5210]: time="2022-06-19T10:36:58Z" level=info msg="Starting ShellHub" mode=multi-user version=v0.9.0
# ...
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: time="2022-06-19T10:41:36Z" level=info msg="Accepted password" user=redacted
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: time="2022-06-19T10:41:36Z" level=info msg="New session request" pty=false user=redacted
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: panic: runtime error: slice bounds out of range [1:0]
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]:
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: goroutine 52 [running]:
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: github.com/shellhub-io/shellhub/agent/sshd.newCmd(0x1480840, {0x0, 0x0}, {0x0, 0x0}, {0x145c2b8, 0x11}, {0x0, 0x0, 0x0})
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: 	/redacted/shellhub/agent/sshd/cmd.go:30 +0x4dc
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: github.com/shellhub-io/shellhub/agent/sshd.(*Server).sessionHandler(0x1450a00, {0x542444, 0x18021c0})
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: 	/redacted/shellhub/agent/sshd/server.go:161 +0xbe8
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: github.com/gliderlabs/ssh.(*session).handleRequests.func1(0x18021c0)
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: 	/redacted/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.3/session.go:262 +0x2c
Sun Jun 19 17:41:36 2022 daemon.err sh[5210]: created by github.com/gliderlabs/ssh.(*session).handleRequests
[...]

PR #1967, opened 2022-06-22
agent: fix panic when client try to use heredoc format

henrybarreto:


henrybarreto:
Mitigate the problem related on isssue https://github.com/shellhub-io/shellhub/issues/1959.

PR #1968, opened 2022-06-23
ssh,agent: add suport to connection with heredoc format

henrybarreto:


ssh/session.go
```diff
@@ -343,7 +345,47 @@ func (s *Session) connect(passwd string, key *rsa.PrivateKey, session sshserver.
 		serverConn.Close()
 		conn.Close()
 		session.Close()
-	} else {
+	case !isPty && requestType == "shell":
+		// When an user try to connect and execute command through heredoc pattern, Pty is set to true, but
```
afriza, line 349: Looking at the location of this comment, shouldn't this be:
```
isPty is set to false
```

henrybarreto: Sorry. You are right. IsPty is set to false when user uses the heredoc format. My falt.

As soon as possible,  I'll correct this comment. 

Thank you.

Issue #1969, opened 2022-06-24
stderr is not propagated


afriza:
via shellhub:
```console
$ ssh user@device@shellhub sh -c " ' echo StandardOutput ; echo StandardError >&2 ' "
StandardOutput
```

direct:
```console
$ ssh user@device sh -c " ' echo StandardOutput ; echo StandardError >&2 ' "
StandardOutput
StandardError
```

I'm still using v0.9.0 ... not sure if recent changes already support this ...

otavio:
It'd be nice if you could test the last RC and see if it helps.

afriza:
do I need to update the server and agent or only agent is sufficient?

otavio:
Yes. The fix requires the server updated.

afriza:
hm.. if we update the server; and some agents are not presently online, there is possibility of version mismatch because we cannot update all agents at the same time.. would old agents be able to connect?

otavio:
I'd prefer if you do a test. Not upgrade your server yet.

gustavosbarreto:
Maybe this has been fixed already by #1968. I'll confirm this.

afriza:
I haven't tried the new code.. but I have a feeling you need to propagate stderr for "exec" request type as well..

https://stackoverflow.com/q/6770206/109747

Issue #1966, opened 2022-06-21
exit code not propagated


afriza:
I can do this if I connect to `user@device` directly
```console
$ ssh user@device notExist ; echo $?
ash: notExist: not found
127
$ ssh user@device true ; echo $?
0
$ ssh user@device false ; echo $?
1
$ ssh user@device exit 333 ; echo $?
77
$ ssh user@device exit 255 ; echo $?
255
$ ssh user@device exit 256 ; echo $?
0


$ ssh user@device ; echo $?
% exit 234
Connection to device closed.
234
```

but if I use `user@device@shellhub` with `!isPty`, I always get `0`;
and I always get `255` if `isPty`.
```console
$ # !isPty
$ ssh user@device@shellhub notExist ; echo $?
0
$ ssh user@device@shellhub exit 255 ; echo $?
0

$ # isPty
$ ssh user@device@shellhub ; echo $?
% exit 234
Connection to device closed by remote host.
Connection to device closed.
255
```

PR #1981, opened 2022-06-27
ssh,agent: progate the exit code from agent to client

henrybarreto:
I've also added a pipe from agent to client with stderr when `requestType` is "exec".


henrybarreto:
Try to solve https://github.com/shellhub-io/shellhub/issues/1966 and https://github.com/shellhub-io/shellhub/issues/1969.

PR #1986, opened 2022-07-01
ssh: refactor new session function

henrybarreto:


ssh/session.go
```diff
@@ -83,105 +83,162 @@ func handlePty(s *Session) {
...
+
+// isSSHID checks if target is a SSHID.
+func (t *Target) isSSHID() bool {
+	return strings.Contains(t.Data, ".")
+}
+
+// isID checks if target is a ID.
+func (t *Target) isID() bool {
+	return !strings.Contains(t.Data, ".")
+}
```
otavio, line 122: I'd drop this method to check if the target isn't an SSHID. So `!isSSHDID()` is equivalent.

ssh/session.go
```diff
@@ -83,105 +83,162 @@ func handlePty(s *Session) {
...
+
+// splitSSHID splits the SSHID target into namespace and hostname into lower strings.
+// Namespace is the device's namespace and hostname is the device's name.
+func (t *Target) splitSSHID() (string, string, error) {
+	if t.isID() {
```
otavio, line 127:
```suggestion
	if !t.isSSHID() {
```

ssh/session.go
```diff
@@ -83,105 +83,162 @@ func handlePty(s *Session) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	handlePty(s)
+	return &Host{Host: host}, nil
+}
+
+// isLocal checks if host address is local.
+func (h *Host) isLocal() bool {
```
otavio, line 156: I'd rename this to `isLocalhost`.

otavio (changes requested):

PR #2039, opened 2022-07-28
ssh: add comments, nolint rules and rename variable on handlers

henrybarreto:
I've added some comments to improve the readability and remove
error variables, what have not used in more than one place on code.

### DIFF
--- a/ssh/server/server.go
+++ b/ssh/server/server.go
@@ -26,13 +26,13 @@ import (
 )
 
 var (
-	ErrPassword   = fmt.Errorf("it could not get the password from context")
-	ErrPublicKey  = fmt.Errorf("it could not get the public key from context")
-	ErrPrivateKey = fmt.Errorf("it could not get the private key")
-	ErrSession    = fmt.Errorf("it could not create the session")
+	ErrPassword   = fmt.Errorf("failed to get the password from context")
+	ErrPublicKey  = fmt.Errorf("failed to get the public key from context")
+	ErrPrivateKey = fmt.Errorf("failed to get the private key")
+	ErrSession    = fmt.Errorf("failed to create the session")
 	ErrWebhook    = fmt.Errorf("the connection was reject by webhook")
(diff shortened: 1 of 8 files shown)


ssh/server/server.go
```diff
@@ -25,16 +25,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var (
```
otavio, line 28: I don't like the idea of dropping those are it makes harder to find which errors we return.

otavio (changes requested):

henrybarreto:
> Those errors should be consts.

We cannot create `const` with a function's return because of that, I have used `var` to list errors.

That is the same pattern of API's errors.

otavio:
Great, so we ought to use `var` with the extra errors as well.

ssh/server/server.go
```diff
@@ -72,29 +73,37 @@ func NewServer(opts *Options, tunnel *httptunnel.Tunnel) *Server {
...
 			"session":  glidersession.Context().Value(ssh.ContextKeySessionID),
 		}).Error("Failed to handler the session")
 
+		// finish close the SSH's session.
 		finish := func(session ssh.Session) {
 			if session != nil {
-				session.Close()
+				err := session.Close()
+				if err != nil {
+					log.WithError(err).Error("Failed to finish the session")
```
otavio, line 97: Here, for example, ought to be in the `var` list.

henrybarreto: I thought, in this case, as this is only an error message and not an error itself, let it hardcoded.

The `var` errors are the only errors what the ShellHub's user receives. The messages inside `log.Error` are exclusively for ShellHub logging and specific for one case, and the user does not have contact with it.

Even though, should we put this messages into variables?

otavio: Consistency. Put it there.

gustavosbarreto (dismissed):

PR #2055, opened 2022-08-01
api: add comments to device tags and remove unneeded validation

henrybarreto:


api/services/device_tags.go
```diff
@@ -4,23 +4,25 @@ import (
...
+	CreateDeviceTag(ctx context.Context, uid models.UID, tag string) error
+	RemoveDeviceTag(ctx context.Context, uid models.UID, tag string) error
 	UpdateDeviceTag(ctx context.Context, uid models.UID, tags []string) error
 }
 
 // DeviceMaxTags is the number of tags that a device can have.
 const DeviceMaxTags = 3
 
-func (s *service) CreateDeviceTag(ctx context.Context, uid models.UID, name string) error {
-	if !validator.ValidateFieldTag(name) {
```
henrybarreto, line 20: That verification isn't more needed because the route already check if the tags are valid.

api/services/device_tags.go
```diff
@@ -62,24 +76,12 @@ func (s *service) UpdateDeviceTag(ctx context.Context, uid models.UID, tags []st
 		}
 
 		return l
-	}
-
-	if len(tags) > DeviceMaxTags {
```
henrybarreto, line 67: Same here.

The route check if the maximum number of allowed tags, and each one is valid.

api/services/device_tags.go
```diff
@@ -30,28 +32,40 @@ func (s *service) CreateDeviceTag(ctx context.Context, uid models.UID, name stri
...
+	if contains(device.Tags, tag) {
+		return NewErrTagDuplicated(tag, nil)
 	}
 
-	return s.store.DeviceCreateTag(ctx, uid, name)
+	return s.store.DeviceCreateTag(ctx, uid, tag)
 }
 
-func (s *service) RemoveDeviceTag(ctx context.Context, uid models.UID, name string) error {
+// RemoveDeviceTag removes a tag from a device. UID is the device's UID and tag is the tag's name.
```
henrybarreto, line 42: This way to document the services is a usual way to do this or it is too much declarative?

otavio: I like it.

otavio (dismissed):

PR #2060, opened 2022-08-05
api: remove billing store implmentation

henrybarreto:


api/store/mongo/store.go
```diff
@@ -16,12 +16,12 @@ var (
 )
 
 type Store struct {
-	db    *mongo.Database
-	cache cache.Cache
+	DB    *mongo.Database
+	Cache cache.Cache
```
otavio, line 20: Maybe use a method so it reduces the number of indirection.

otavio (changes requested):

otavio (dismissed):

gustavosbarreto (approved):

PR #2075, opened 2022-08-10
agent: ensure that selfupdater are running inside a docker container

gustavosbarreto:


agent/selfupdater/updater_native.go
```diff
@@ -3,26 +3,6 @@
...
-
-func (n *nativeUpdater) ApplyUpdate(_ *semver.Version) error {
-	return nil
-}
-
-func (n *nativeUpdater) CompleteUpdate() error {
-	return nil
-}
-
 func NewUpdater(version string) (Updater, error) {
```
otavio, line 6: As this file is kept, might make sense to keep the methods here.

otavio (changes requested):

agent/selfupdater/updater_native.go
gustavosbarreto: > As this file is kept, might make sense to keep the methods here.

See the build tags to understand why the methods were moved to another file.

gustavosbarreto: I mean, looking only at the diff it doesn't appear to do what I had in mind

otavio: Sorry but it isn't clear to me. Please provide the link where I see it.

gustavosbarreto: https://github.com/shellhub-io/shellhub/blob/97881681d4294204c63240eee73f8b01a7c3ff05/agent/selfupdater/updater_docker.go#L1-L4

https://github.com/shellhub-io/shellhub/blob/97881681d4294204c63240eee73f8b01a7c3ff05/agent/selfupdater/updater_docker.go#L223-L234

https://github.com/shellhub-io/shellhub/blob/97881681d4294204c63240eee73f8b01a7c3ff05/agent/selfupdater/updater_native.go#L1-L8

otavio: Ok; sure. But the native struct has been moved to `updater.go` and it seems to belong in `updater_native.go`.

gustavosbarreto: There is a two `NewUpdater` func that are used according to the build tags. That's why I need to keep the `updater_native.go` file.

gustavosbarreto: > Ok; sure. But the native struct has been moved to `updater.go` and it seems to belong in `updater_native.go`.

Yup, the struct has been moved to `updater.go` cause we need them both in `NewUpdater` metods (w/o docker)

PR #2085, opened 2022-08-15
api: remove option to disable cache

henrybarreto:


api/main.go
```diff
@@ -11,15 +11,15 @@ import (
 func main() {
 	rootCmd := &cobra.Command{Use: "api"}
 
-	rootCmd.AddCommand(serverCmd)
+	rootCmd.AddCommand(serverCMD)
```
gustavosbarreto, line 14: Why this variable was renamed?

PR #2099, opened 2022-08-26
ssh, docker: set log level to trace when in development

gessecarneiro:


ssh/main.go
```diff
@@ -22,6 +23,13 @@ const (
 )
 
 func main() {
+	if value, ok := os.LookupEnv("SHELLHUB_ENV"); ok && value == "development" {
+		log.SetLevel(log.TraceLevel)
+		log.Debug("Log level set to Trace")
```
otavio, line 28: as you are enabling the trace log level, this could use Trace.

ssh/main.go
```diff
@@ -22,6 +23,13 @@ const (
 )
 
 func main() {
+	if value, ok := os.LookupEnv("SHELLHUB_ENV"); ok && value == "development" {
+		log.SetLevel(log.TraceLevel)
```
otavio, line 27: One aspect to consider is if we should allow setting the log level directly. Even in production, putting it to trace may help when debugging issues.

otavio (changes requested):

PR #2161, opened 2022-09-27
ssh: fix break line and error messages sent to user

henrybarreto:
I have fixed the missing breaking line when an error message is sent to the user, and I added the one when client cannot authenticate on device/agent.

screenshot 1

### DIFF
--- a/ssh/server/handler/ssh.go
+++ b/ssh/server/handler/ssh.go
@@ -53,6 +53,7 @@ var (
 	ErrPty                = fmt.Errorf("failed to request the pty to agent")
 	ErrShell              = fmt.Errorf("failed to get the shell to agent")
 	ErrTarget             = fmt.Errorf("failed to get client target")
+	ErrAuthentification   = fmt.Errorf("failed to authenticate to device")
 )
 
 // sendAndInformError sends the external error to client and log the internal one to server.
@@ -159,7 +160,7 @@ func SSHHandler(tunnel *httptunnel.Tunnel) gliderssh.Handler {
 
 		err = connectSSH(ctx, client, sess, config, api, opts)
 		if err != nil {
-			sendAndInformError(client, err, ErrConnect)
+			sendAndInformError(client, err, err)
 
 			return
 		}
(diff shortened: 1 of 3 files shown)


ssh/server/handler/ssh.go
```diff
@@ -169,7 +170,7 @@ func SSHHandler(tunnel *httptunnel.Tunnel) gliderssh.Handler {
 func connectSSH(ctx context.Context, client gliderssh.Session, sess *session.Session, config *gossh.ClientConfig, api internalclient.Client, opts ConfigOptions) error {
 	connection, reqs, err := sess.NewClientConnWithDeadline(config)
```
henrybarreto, line 171: `NewClientConnWithDeadline` can return many errors related to the connection between the server/client to device/agent, so I believe that “failed to authenticate to device”  is the most generic, but “helper” error's message in this context.

ssh/session/session.go
```diff
@@ -27,7 +27,7 @@ var (
 	ErrFirewallBlock = fmt.Errorf("you connot connect to this device because a firewall rule block your connection")
 	ErrHost          = fmt.Errorf("failed to get the device address")
 	ErrFindDevice    = fmt.Errorf("failed to find the device")
-	ErrDial          = fmt.Errorf("failed to tunnel the server to agent")
+	ErrDial          = fmt.Errorf("failed to tunnel the server to agent. the device can be offline")
```
henrybarreto, line 30: Something similar happens here: One of the reasons to tunnel connection fail is the device be offline.

otavio, line 30: Please use: failed to connect to device agent, please check the device connection

ssh/server/handler/ssh.go
```diff
@@ -159,7 +160,7 @@ func SSHHandler(tunnel *httptunnel.Tunnel) gliderssh.Handler {
 
 		err = connectSSH(ctx, client, sess, config, api, opts)
 		if err != nil {
-			sendAndInformError(client, err, ErrConnect)
+			sendAndInformError(client, err, err)
```
henrybarreto, line 163: Propagating error from `connectSSH`  to the client.

ssh/session/session.go
henrybarreto: Perfect

henrybarreto: https://github.com/shellhub-io/shellhub/pull/2161/files#diff-c4ddc485bc4aee5bf70bf7c489310bb48e38c6ea908529464013c17ef40df86dR30

otavio: What other cases?

PR #2166, opened 2022-09-30
api, pkg: add business layer

henrybarreto:


api/businesses/namespace.go
```diff
@@ -0,0 +1,157 @@
...
+	maxDevices    int
+
+	err error
+}
+
+// CreateNamespace initializes a new NamespaceBuilder to create a new namespace.
+//
+// Example:
+//
+//  namespace := bussiness
```
henrybarreto, line 29: I'll create a `ExampleCreateNamespace` to it on `namespace_test.go` file.

PR #2182, opened 2022-10-07
ssh: split the web SSH connection into two steps

henrybarreto:
To start a new SSH connection through web terminal, the client
needs to send to the SSH's server the device access credentials as
a JSON POST request. These credentials are the device's UID, a
device's username, either the device's password for that username
or fingerprint plus its signature. This POST request will return a
session's token to be used to directly connect to the device in a
next request. This token remains valid for 30 seconds.

`
{
    "device": <device>,
    "username": <username>,
    "password": <password>
}
`

or

`
{
    "device": <device>,
    "username": <username>,
    "fingerprint" <fingerprint>,
    "signature": <signature>
}
`

The main idea behind this change is no more send the credentials
through GET request that were logged by NGINX, but with POST one,
what can be encrypted.

With the token, the client sends it as a query parameter to the
same route, but through the GET method. Beyond the token, this
route requires the number of columns and of rows from the web
terminal. If any of those parameters are empty or invalid, the
connection will fail, reporting the problem to the user.

Subsequently, if the token is still alive and the data is valid,
the SSH connection should behave like usual, connecting to the
agent on the device indicated.


gustavosbarreto:
The most important thing about this improvement is that now the auth credentials will not be sent by a GET request that were logged by NGINX, now the credentials will be sent by a POST request which could be encrypted when using HTTPS

otavio (approved):

PR #2198, opened 2022-10-12
ui: add autocomplete attribute for some inputs

afriza:
I added [`autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute for `email`, `username`, `current-password`, and `new-password`.

I haven't tested but the changes are pretty straightforward.


afriza:
Assuming the changes are okay, it seems I need to update the tests snapshot as well. Do you prefer force push or new commit?

AntonyRafael (approved):
It seems ok to me, most browsers have this feature by default.

But this will support if some user has a very old browser.

afriza:
Specifically, my motivation to add this attribute is because when I tried to change my password, I didn't get new-password suggestion from Google Chrome, and it is normally there on other sites.

I planned to use the randomly generated password from Google Chrome & save it with Chrome's builtin password manager. With these changes, the new-password suggestion should appear.

PR #2200, opened 2022-10-14
pkg: add MaxNamespaces property to user's model

henrybarreto:


api/services/namespace.go
```diff
@@ -109,6 +109,8 @@ func (s *service) CreateNamespace(ctx context.Context, namespace request.Namespa
 		return nil, NewErrNamespaceCreateStore(err)
 	}
 
+	s.store.UserUpdateNamespacesCounter(ctx, userID, 1) // nolint: errcheck
```
otavio, line 112: Must run in a transaction, no?

henrybarreto: Yes! You're right!

I don't even need these store methods; I just need to put it together with namespace's creation method.

Thank you.

api/services/namespace.go
```diff
@@ -143,11 +143,16 @@ func (s *service) DeleteNamespace(ctx context.Context, tenantID string) error {
 	if err != nil {
 		return NewErrNamespaceNotFound(tenantID, err)
 	}
+
 	if err := deleteReportUsage(s.client.(req.Client), ns); err != nil {
 		return err
 	}
 
-	return s.store.NamespaceDelete(ctx, tenantID)
+	if err := s.store.NamespaceDelete(ctx, tenantID); err != nil {
```
otavio, line 151: Transaction as well.

api/store/mongo/device_store_tags_test.go
```diff
@@ -13,7 +13,7 @@ import (
 func TestDeviceCreateTag(t *testing.T) {
 	data := initData()
 
-	db := dbtest.DBServer{}
+	db := dbtest.DBServer{Replicaset: true}
```
otavio, line 16: In the past, we had support for use with and without replicaset. However, since a few releases ago, we have used replicaset in all deployments, and as a consequence, we should use it by default and avoid touching all test code.

otavio (changes requested):

api/store/mongo/migrations/migration_49.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/sirupsen/logrus"
+	migrate "github.com/xakep666/mongo-migrate"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+var migration49 = migrate.Migration{
+	Version:     49,
+	Description: "",
```
gustavosbarreto, line 14: missing description

otavio (approved):

PR #2246, opened 2022-11-09
api: add appropriated error when device not found

henrybarreto:


otavio:
I believe it needs:

- [ ] addition on OpenAPI
- [ ] fix of unit tests

otavio (approved):

PR #2321, opened 2022-12-20
Feature/report try device connection

henrybarreto:


henrybarreto:
I think that an approach with workers is better than create a "reporter" for each needed event

PR #2345, opened 2023-01-06
env: remove empty quotes to represent empty values 

henrybarreto:


otavio (approved):

henrybarreto:
## The issue

#2341 

> I have successfully followed the setup instructions for deploying the self hosted shellhub service (v0.10.9) locally on my Ubuntu desktop. I have created my first user and namespace, but every time I try and login to the service (with my credentials) I get a small dialog popping up saying "The request has failed. Please try again". It does this even if I type in rubbish into the username and password.

## The response

> [@drumwhizz](https://github.com/drumwhizz), the problem seems to be with the *docker-compose* version what you are using.
> 
> > I'm assuming you are using the Ubuntu 20.04, sorry.
> 
> Looking for the *docker-compose* package on your distro, I've checked that this version, **[1.25.0](https://packages.ubuntu.com/focal/docker-compose)**, doesn't support empty (`""`) strings as null on `.env`.
> 
> To work around that issue, you can run that script. It'll remove all `""` from the `.env`, allowing the API server to run normally.
> 
> `sed -i 's,"",,g' .env`
> 

## The explanation

When using docker-compose in version > 1.25.0, Sentry's environment variable is not empty when its value on the `.env` is `SHELLHUB_SENTRY_DSN=""`, but contains the double quotes as a string itself.

Screenshot from 2023-01-06 11-10-54

On the < 1.26.0, the docker-compose has delegated to `python-dotenv` the processing of `.env` file, what becoming the `SHELLHUB_SENTRY_DSN=""` valid as empty environmental variable on the container.

Screenshot from 2023-01-06 11-02-12

## The PR

To avoid this problem again, it was requested to remove the attribution of empty string as `VALUE=""` on `.env` file.

otavio (approved):

PR #2529, opened 2023-02-27
ui: improve typescript types on api

AntonyRafael:


ui/src/store/modules/devices.ts
```diff
@@ -304,6 +304,7 @@ export const devices: Module<DevicesState, State> = {
 
     updateDeviceTag: async (context, data) => {
       try {
+        console.log(data);
```
otavio, line 307: It seems like a test change that has been forgotten here.

PR #2536, opened 2023-03-03
github: update go version from QA workflow to 1.20

henrybarreto:


otavio:
If moving, please replace all 1.17 entries. Many places refer to it.

PR #2537, opened 2023-03-03
fix/command output loss

gustavosbarreto:
- agent: fix exec command output loss
- ssh: Wait for stdout pipe to finish reading bytes from remote client.


henrybarreto:
🙏🏼

gustavosbarreto:
@benasse, would you be able to run a test using this branch? We have tested it internally and were no longer able to reproduce the issue. I would be very grateful if you could test it one more time.

benasse:
@gustavosbarreto, no problem to test.
I'm trying to do this during the day.

benasse:
I tried to update on the server side, but it would make my life easier if you provide me with a docker image with the patch for shellhubio/ssh.
I don't have a proper dev environment to build and test it ...

If it's complicated for you I should manage to come up with something but it will take me a bit more time.

gustavosbarreto:
> I tried to update on the server side, but it would make my life easier if you provide me with a docker image with the patch for shellhubio/ssh. I don't have a proper dev environment to build and test it ...
> 
> If it's complicated for you I should manage to come up with something but it will take me a bit more time.

@benasse I have made a release candidate (`v0.11.3-rc.2`)

benasse:
I have updated and I confirm that I can no longer reproduce the problem!
Awsome! 😃 😃 😃

Issue #2562, opened 2023-03-09
Longitude and Latitude Duplicated in `service.CreateSession`


gustavosbarreto:
### Description

While testing the `CreateSession` function in the service package, I have identified an issue with the assignment of the longitude and latitude values in the `models.SessionPosition` struct. See the following code snippet:

https://github.com/shellhub-io/shellhub/blob/d0502b7f38c5fa73eb2c4131295ac40d2a3531a7/api/services/session.go#L40-L43

I observed that both the `Longitude` and `Latitude` values are assigned with the value of `position.Longitude`, resulting in duplicate and incorrect coordinates.

### Edition

Community

### Version

master

PR #2563, opened 2023-03-09
api: fix usage of longitude on latitude field while session creation

henrybarreto:


henrybarreto:
Related to: https://github.com/shellhub-io/shellhub/issues/2562

otavio (approved):

PR #2568, opened 2023-03-10
env: Add SHELLHUB_PUBLIC_URL_DOMAIN env

gustavosbarreto:
This commit adds the `SHELLHUB_PUBLIC_URL_DOMAIN` environment variable to the `.env` file, which is used to define the public URL domain for accessing devices via HTTP in a ShellHub instance.

The value of this variable is used in the `gateway/shellhub.conf` file to generate the public URL for HTTP access.

The `docker-compose.yml` file is also updated to pass this variable to the gateway container.

NOTE: If `SHELLHUB_PUBLIC_URL_DOMAIN` is not defined, the value of `SHELLHUB_DOMAIN` is used as the default value.

### DIFF
--- a/gateway/shellhub.conf
+++ b/gateway/shellhub.conf
@@ -392,12 +392,13 @@ server {
     }
 }
 
+{{- $PUBLIC_URL_DOMAIN := or (env.Getenv "SHELLHUB_PUBLIC_URL_DOMAIN") (env.Getenv "SHELLHUB_DOMAIN") }}
 server {
    listen 80;
-   server_name ~^(?<namespace>.+)\.(?<device>.+)\..+$;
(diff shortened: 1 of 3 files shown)


gateway/shellhub.conf
```diff
@@ -392,12 +392,13 @@ server {
     }
 }
 
+{{- PUBLIC_URL_DOMAIN := or (env.Getenv "SHELLHUB_PUBLIC_URL_DOMAIN") (env.Getenv "SHELLHUB_DOMAIN") }}
 server {
    listen 80;
-   server_name ~^(?<namespace>.+)\.(?<device>.+)\..+$;
+   server_name ~^(?<namespace>.+)\.(?<device>.+)\.{{ $PUBLIC_URL_DOMAIN }}$;
```
otavio, line 398: Is the same condition as using `SHELLHUB_DOMAIN` as default isn't required here?

PR #2629, opened 2023-03-27
api,pkg: add slot strategy to block device switching

henrybarreto:
Here I introduce the concept of "slot", which is a way to control the number of
devices that namespace can have in a given time and when there is a device
limitation. The "slot" is a document that is created when a device is
"accepted" and its status change to "removed" when the device is deleted.
A "slot" with the status "removed" expires after a period of time, allowing a
new device to be accepted at its place.

### DIFF
--- a/api/services/device.go
+++ b/api/services/device.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	"github.com/shellhub-io/shellhub/api/store"
 	req "github.com/shellhub-io/shellhub/pkg/api/internalclient"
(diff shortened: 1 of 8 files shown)


api/services/device.go
```diff
@@ -149,6 +164,102 @@ func (s *service) UpdatePendingStatus(ctx context.Context, uid models.UID, statu
...
+					// We set the variable to avoid creating a new slot for the device.
+					set = true
+
+					break
+				}
+			}
+
+			// If the device is not into the slot, but there is a device's slot with its status as removed, and it is
+			// expired, we need to delete the slot and create a new one for the device.
+			// TODO: What could be the best way to get the slot's duration? A environment variable? A value hardcoded?
```
henrybarreto, line 231: Maybe It could be an env variable?

otavio: I prefer to encapsulate the slot handling into a specific package and reduce the coupling between the device and the slot management. Also, let's consider having this in a cloud project to avoid adding this code in the community version.

What do you think?

otavio (approved):

henrybarreto:
Closed in favor of https://github.com/shellhub-io/shellhub/pull/2630

PR #2630, opened 2023-03-29
api,pkg: add system to avoid device switching.

henrybarreto:


api/services/device.go
```diff
@@ -149,10 +164,39 @@ func (s *service) UpdatePendingStatus(ctx context.Context, uid models.UID, statu
...
+
+		if slot == nil {
+			removed, err := s.store.DeviceRemovedList(ctx, tenant)
+			if err != nil {
+				return NewErrDeviceRemovedList(err)
+			}
+
+			if (ns.DevicesCount + len(removed)) >= ns.MaxDevices {
+				return NewErrDeviceRemovedFull(ns.MaxDevices, nil)
+			}
```
otavio, line 192: Making the list of removed devices while a simple count using the tenant as a filter would do the work.

api/services/device.go
```diff
@@ -149,10 +164,39 @@ func (s *service) UpdatePendingStatus(ctx context.Context, uid models.UID, statu
...
 
+	// NOTICE: The logic below is only executed when the new status is "accepted".
+
+	if ns.MaxDevices > 0 {
+		slot, err := s.store.DeviceRemovedGet(ctx, tenant, uid)
+		if err != nil && err != store.ErrNoDocuments {
+			return NewErrDeviceRemovedGet(err)
+		}
+
+		if slot == nil {
```
otavio, line 184: I'd handle the `if slot != nil` case first. However, adding a comment explaining it is good as well.

api/store/mongo/device_store.go
```diff
@@ -471,3 +471,50 @@ func (s *Store) DeviceUpdate(ctx context.Context, uid models.UID, name *string,
 
 	return FromMongoError(err)
 }
+
+func (s *Store) DeviceRemovedList(ctx context.Context, tenant string) ([]models.DeviceRemoved, error) {
```
otavio, line 475: A `DeviceRemovedCount` method seems like a better approach.

otavio (changes requested):

PR #2631, opened 2023-03-29
api: fix wrong key when deleting a device from cache

henrybarreto:


otavio (changes requested):
The commit log needs to be more accurate. For example, it implies you removed the cache removal, but in fact, it changed the order and logic. Please rework the commit log to explain what issue it fixes.

gustavosbarreto:
> The commit log needs to be more accurate. For example, it implies you removed the cache removal, but in fact, it changed the order and logic. Please rework the commit log to explain what issue it fixes.

I guess this change should only fix the incorrect name of the key used in the cache, keeping the logic unchanged.

PR #2644, opened 2023-04-03
api: add acceptable attribute to device and when listing it

henrybarreto:


api/store/mongo/device_store.go
```diff
@@ -69,6 +71,46 @@ func (s *Store) DeviceList(ctx context.Context, pagination paginator.Query, filt
...
+			"$addFields": bson.M{
+				"acceptable": false,
+			},
+		})
+	}
+
+	// When status is "pending" or "rejected", the user should be able to accept the device only if the maximum number
+	// of devices has not been reached.
+	if status == "pending" || status == "rejected" {
+		if mode&store.DeviceListModeMaxDeviceReached != 0 {
```
henrybarreto, line 86: Is there a simple/better way to check if mode "is" a store.DeviceListModeMaxDeviceReached.

PR #2726, opened 2023-05-05
cli,pkg: add lowcase validation tag on namespace model

danyecls:


gustavosbarreto (changes requested):
Please provide a test case for this

cli/services/namespaces.go
```diff
@@ -25,7 +26,7 @@ func (s *service) NamespaceCreate(namespace, username, tenant string) (*models.N
 	}
 
 	ns := &models.Namespace{
-		Name:     namespace,
+		Name:     strings.ToLower(namespace),
```
otavio, line 29: It would be great if this task could be completed within the inner layer to ensure that the logic is centralized in one location.

gustavosbarreto: There is no other way to create a namespace in the CLI. I think the validator package, which is used in both the CLI and the API, should validate the namespace name correctly, ensuring that it is always in lowercase

PR #2758, opened 2023-05-11
fix(api): avoid internal error when generic error occour

henrybarreto:


api/pkg/echo/handlers/errors.go
```diff
@@ -78,9 +78,10 @@ func NewErrors(reporter *sentry.Client) func(error, echo.Context) {
 			}
 
 			maybeReport(reporter, converted, ctx.Request())
+		} else if he, ok := err.(*echo.HTTPError); ok {
```
otavio, line 81: `herr` or `http_err`, please.

henrybarreto: `herr` sounds better than just `he`. Thanks.

PR #2782, opened 2023-05-17
fix(ssh): force exit when exec command is run on agent less than 0.9.3

henrybarreto:
Agent version less than 0.9.3 does not support SSH exec command, what makes some request for it hang foverever. Version great than that, closes correctly and return the status code from the ran command.

### DIFF
--- a/ssh/server/handler/ssh.go
+++ b/ssh/server/handler/ssh.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/Masterminds/semver"
 	gliderssh "github.com/gliderlabs/ssh"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/shellhub-io/shellhub/pkg/api/internalclient"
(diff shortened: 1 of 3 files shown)


ssh/server/handler/ssh.go
```diff
@@ -398,6 +399,24 @@ func exec(api internalclient.Client, uid string, agent *gossh.Session, client gl
 		return err
 	}
 
+	dev, err := api.GetDevice("234ab181155c14c1383e2d1cd1fdfe3c770b54ed5aea2343bd9f77451908b60a")
```
otavio, line 402: fixed device?

otavio (changes requested):

ssh/server/handler/ssh.go
henrybarreto: sorry, I forgot it. Fixing.

otavio (approved):

PR #2834, opened 2023-05-25
feat(ui): add global css style

gustavosbarreto:
The commit adds global CSS styles to the UI, improving the overall
consistency of the application and avoiding duplicated code

The changes include:

* Added global.css file to the assets folder
* Removed the .full-height class from various components

### DIFF
--- a/ui/src/views/ForgotPassword.vue
+++ b/ui/src/views/ForgotPassword.vue
@@ -140,10 +140,6 @@ export default defineComponent({
 </script>
 
 <style>
-.full-height {
-  height: 100vh;
-}
-
 .v-field__append-inner {
   cursor: pointer;
 }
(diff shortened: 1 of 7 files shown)


ui/src/views/ForgotPassword.vue
```diff
@@ -140,10 +140,6 @@ export default defineComponent({
 </script>
 
 <style>
-.full-height {
-  height: 100vh;
-}
-
 .v-field__append-inner {
   cursor: pointer;
 }
```
otavio, line 145: Should it be removed as well?

otavio (changes requested):

ui/src/views/ForgotPassword.vue
gustavosbarreto: Yes, it should be removed. It seems to have been included to fix a specific issue (vuetify alpha?). However, it no longer serves any practical purpose

PR #2890, opened 2023-06-13
improvement/refactor services test

danyecls:


otavio (approved):

api/services/session_test.go
```diff
@@ -44,41 +32,56 @@ func TestListSessions(t *testing.T) {
 	cases := []struct {
 		name          string
 		pagination    paginator.Query
-		requiredMocks func()
+		requiredMocks func(query paginator.Query)
```
henrybarreto, line 35: Should we instantiate the `query` inside the `requiredMocks` as in other tests?

I think I'll enforce the test and make it broken if `query` changes in some way. Am I right?

api/services/namespace_test.go
```diff
@@ -155,59 +218,58 @@ func TestListNamespaces(t *testing.T) {
...
 
 	cases := []struct {
-		name          string
+		description   string
+		user          *models.User
+		namespace     *models.Namespace
 		ctx           context.Context
-		tenantID      string
-		requiredMocks func()
+		requiredMocks func(namespace *models.Namespace, user *models.User)
```
henrybarreto, line 235: As `query`, here it could be the same?

api/services/namespace_test.go
```diff
@@ -296,63 +358,74 @@ func TestCreateNamespace(t *testing.T) {
...
-		},
-		Settings: &models.NamespaceSettings{SessionRecord: true},
-		TenantID: "xxxxx",
-	}
-
 	cases := []struct {
-		name          string
-		requiredMocks func()
+		description   string
+		requiredMocks func(namespace requests.NamespaceCreate)
```
henrybarreto, line 363: here too

api/services/namespace_test.go
```diff
@@ -1075,78 +1512,223 @@ func TestEditNamespaceUser(t *testing.T) {
...
 			},
-			Expected: NewErrNamespaceMemberNotFound(passiveMember.ID, nil),
+			Expected: NewErrNamespaceMemberNotFound("passiveMemberID", nil),
 		},
 		{
-			Name:          "EditNamespaceUser fails when could not find active member inside namespace",
-			TenantID:      namespacePassiveObserver.TenantID,
-			UserID:        activeMember.ID,
-			MemberID:      passiveMember.ID,
+			description:   "EditNamespaceUser fails when could not find active member inside namespace",
```
henrybarreto, line 1659: Should `description` be capitalized like the other fields?

henrybarreto: I think you could check to see if there is this inconsistency somewhere else.

api/services/session_test.go
danyecls: corrected in other cases

api/services/namespace_test.go
danyecls: corrected in other cases

api/services/namespace_test.go
danyecls: fixed in all cases

api/services/namespace_test.go
danyecls: corrected in other cases

henrybarreto (approved):

otavio (approved):

PR #2915, opened 2023-06-23
fix(docker-compose): Use bash as shell

otavio:


gustavosbarreto:
I don't think that's the best approach as it makes bash mandatory. Wouldn't it be better to make the script compatible with POSIX?

otavio:
After running both `checkbashisms` and `shellcheck`, I did not find any issues. Therefore, I suggested testing to confirm if it is the root cause.

Issue #2931, opened 2023-06-29
Incorrect `token` field format in `GET /api/auth/user` response


gustavosbarreto:
### Description

The `GET /api/auth/user` route is returning a JSON response that includes a field called token. However, the token field is currently in the incorrect format. It includes the prefix "Bearer" before the actual token value, which is not expected.

Current JSON response:
```json
{
  "token": "Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiZ3VzdGF2byIsImFkbWluIjp0cnVlLCJ0ZW5hbnQiOiIwMDAwMDAwMC0wMDAwLTQwMDAtMDAwMC0wMDAwMDAwMDAwMDAiLCJpZCI6IjY0MTIxMWI2OTFhZmFmZjI0OTk4MzkzZCIsInJvbGUiOiJvd25lciIsImNsYWltcyI6InVzZXIiLCJleHAiOjE2ODgzMjc0OTZ9.kY0LFuvZghzHuq-wL6me_TZcO3T4MMWvKlKLa4sTPqkYaWR9_teRGz-Byn1PaEjJ9hAJQmpbHP1ivrGYHAPXGcXsTI3i-Nqroh382WvBIbRrZREFkODBvSZeEArW2pYLBMLcokN6j2v4K6iblZ3ye7azt-HBK59oScnM21uN_Ez6zbUpmEbSXS4RtIjISFfiF2ipVfoaMJne3Fkh6GD9d8BR3LUy8C5EawAo0wDtNLZnf31C7excugQWkAXiLyABtKb77Bab6fGxPjkv1qSTNP1AYuJAS7FLsIokRFOcuNmVG75m7kqiHoeEnut6L8FsfV6rfYj2lK6VH5eK_N8YNA",
  "user": "gustavo",
  "name": "gustavo",
  "id": "641211b691afaff24998393d",
  "tenant": "00000000-0000-4000-0000-000000000000",
  "role": "owner",
  "email": "gustavo@gmail.com"
}
```

Expected JSON response:
```json
{
[...]

PR #2938, opened 2023-07-03
api(services): correct token

danyecls:


api/store/mongo/migrations.go
```diff
@@ -17,7 +17,7 @@ import (
 func ApplyMigrations(db *mongo.Database) error {
 	logrus.Info("Creating lock for the resource migrations")
 
-	lockClient := lock.NewClient(db.Collection("locks", options.Collection().SetWriteConcern(writeconcern.New(writeconcern.WMajority()))))
+	lockClient := lock.NewClient(db.Collection("locks", options.Collection().SetWriteConcern(writeconcern.New(writeconcern.WMajority())))) //nolint:staticcheck
```
gustavosbarreto, line 20: Why this is necessary?

gustavosbarreto (changes requested):
I'm not sure if this fixes #2931.

Does the expected JSON match the description in the issue?

api/store/mongo/migrations.go
danyecls: Added a nolint comment because github was failing the API validation with the  error: "SA1019: writeconcern.New is deprecated..."

danyecls:
> I'm not sure if this fixes #2931.
> 
> Does the expected JSON match the issue description?

Previously changed route was incorrect, corrected changes to equivalent route GET auth/user

henrybarreto (approved):

henrybarreto (approved):

otavio (approved):

PR #2953, opened 2023-07-10
fix/correct device version

danyecls:


ssh/server/handler/ssh.go
```diff
@@ -433,8 +424,21 @@ func exec(api internalclient.Client, uid string, device *models.Device, agent *g
...
+	if dev.Info.Version != "latest" {
+		ver, err = semver.NewVersion(dev.Info.Version)
+		if err != nil {
+			log.WithError(err).WithFields(log.Fields{
+				"client": uid,
+			}).Error("failed to parse device version")
+
+			return err
+		}
+	}
```
otavio, line 438: I've been considering ways to avoid multiple comparisons for the latest version. One potential solution is to label it as `9999.9999.9999`. I'm confident that we won't need to release that many versions in our lifetime.

PR #2969, opened 2023-07-19
fix(api,pkg): return from validation with wrong name and without rule

henrybarreto:


henrybarreto:
Screenshot from 2023-07-19 12-26-52

PR #2988, opened 2023-07-28
refactor(agent): convert "agent" binary into a package 

### LINKED ISSUE #2930: Convert "agent" binary into a package
### Description

Currently, we have a binary that is generated from the source code located in the "agent" directory. However, we would like to convert this agent into a package so that it can be used outside the binary as well. This will enable us to create an instance of the agent without the need to execute a binary.

## Task

* [x] Refactor the "agent" code to remove any dependencies on the binary execution.
* [x] Organize the "agent" code into a Go package structure.
* [x] Update the necessary import statements and dependencies to ensure the package can be utilized independently.
* [ ] Test the converted package to ensure it functions correctly when used outside the binary.
* [x] Document the steps required to instantiate and use the agent package separately from the binary.

### Edition

Community

### Version

master

henrybarreto:
Closes #2930


henrybarreto:
## Tasks:
Some tasks I believe we need to perform before merging this PR.

- [ ] Test agent connections
  - [ ] SSH
  - [ ] Web
  - [ ] Exec
  - [ ] SCP
  - [ ] SFTP
  - [ ] "HEREDOC"
- [ ] Test agent update

I would like your help, @gustavosbarreto, here.

PR #3023, opened 2023-08-08
feat(gateway): enable IPv6 in each listen directive

gustavosbarreto:


otavio (approved):

gustavosbarreto:
`gateway/conf.d/shellhub.conf` no longer exists. The edge embeds Caddy and generates its configuration from the environment (#6780).

`listen [::]:...` was an nginx quirk: nginx binds IPv4 only unless told otherwise. Caddy binds dual-stack by default, and on the current image the listener is `:::80`.

Whether the edge is reachable over IPv6 in a Compose deployment is a separate question, decided by the published port on the host rather than by the listener.

Closing.

PR #3050, opened 2023-08-22
Fix/username validator

heiytor:


pkg/models/namespace.go
```diff
@@ -37,6 +37,6 @@ type NamespaceSettings struct {
 
 type Member struct {
 	ID       string `json:"id,omitempty" bson:"id,omitempty"`
-	Username string `json:"username,omitempty" bson:"username,omitempty" validate:"min=3,max=30,alphanum,ascii"`
+	Username string `json:"username,omitempty" bson:"username,omitempty" validate:"required"`
```
henrybarreto, line 40: I think we could set the `required` and `username` tags.

PR #3063, opened 2023-08-28
Cli/refactor cmd

heiytor:


README.md
```diff
@@ -128,13 +129,13 @@ $ make start
...
 ```
-$ ./bin/add-user <username> <password> <email>
+$ ./bin/cli user create <username> <password> <email>
 ```
 
 Create a namespace for grouping devices and set user as owner:
 
 ```
-$ TENANT_ID=00000000-0000-4000-0000-000000000000 ./bin/add-namespace <namespace> <owner>
+$ TENANT_ID=00000000-0000-4000-0000-000000000000 ./bin/cli namespace create <namespace> <owner>
```
henrybarreto, line 138: The right usage of this command is something like:

`cli namespace create <namespace> <owner> [tenant] [flags]`

The [tenant] parameter is optional, but must be set as the third parameter in the command.

henrybarreto: This is a thing I've noticed in a first glance. I'll provide a full review when it is marked as ready.

heiytor: Since this is how the command was called, I only changed the executable name, not its usage. So, is the correct way `cli namespace create <namespace> <owner> [tenant] [flags]`?

henrybarreto: The script `./bin/cli` performs commands directly on the `cli` binary. Check it out.

henrybarreto (dismissed):

PR #3078, opened 2023-09-04
feat(api): implementing MFA two-factor authentication

danyecls:


api/routes/auth_test.go
```diff
@@ -185,7 +185,7 @@ func TestAuthUser(t *testing.T) {
 				Password: "testpassword",
 			},
 			requiredMocks: func() {
-				mock.On("AuthUser", gomock.Anything, gomock.Anything).Return(&models.UserAuthResponse{}, nil).Once()
+				mock.On("AuthUser", gomock.Anything, gomock.Anything, gomock.Anything).Return(&models.UserAuthResponse{}, nil).Once()
```
henrybarreto, line 188: According to [testify doc](https://pkg.go.dev/github.com/stretchr/testify/mock#Anything), the `gomock.Anything` must be used when the argument isn't relevant to the test made. I believe could be a good take to check if the arguments sent to the functions tested are really not necessary because it can cause a false positive on it.

henrybarreto: I mean, not only on this function, in general on tests modified.

henrybarreto: You could develop this idea when building the tests here: https://github.com/shellhub-io/shellhub/pull/3177/files#r1389292023

pkg/models/user.go
```diff
@@ -24,6 +24,9 @@ type User struct {
 	CreatedAt      time.Time `json:"created_at" bson:"created_at"`
 	LastLogin      time.Time `json:"last_login" bson:"last_login"`
 	EmailMarketing bool      `json:"email_marketing" bson:"email_marketing"`
+	MFA            bool      `json:"status_mfa" bson:"status_mfa"`
+	Secret         string    `json:"secret" bson:"secret"`
+	Codes          []string  `json:"codes" bson:"codes"`
```
henrybarreto, line 29: I think we will need a migration here to avoid parsing problems.

danyecls: a migration was added to include the necessary fields in 'user'

PR #3088, opened 2023-09-08
fix(api): prevent resource initialization from being blocked by a previous one

henrybarreto:


henrybarreto:
Closed in favor of https://github.com/shellhub-io/shellhub/pull/3176.

The new pull request, cited above, is simpler and, essentially, does the same.

PR #3103, opened 2023-09-15
refactor store's tests

heiytor:
Since we don't have any method to mock certain things, I can't test all return possibilities. For example in this method:
```go
func (s *Store) DeviceCreate(ctx context.Context, d models.Device, hostname string) error {
	if hostname == "" {
		hostname = strings.ReplaceAll(d.Identity.MAC, ":", "-")
	}

	var dev *models.Device
	if err := s.cache.Get(ctx, strings.Join([]string{"device", d.UID}, "/"), &dev); err != nil {
		logrus.Error(err) // <------
	}

	q := bson.M{
		"$setOnInsert": bson.M{
			"name":              hostname,
			"status":            "pending",
			"status_updated_at": time.Now(),
			"created_at":        clock.Now(),
			"tags":              []string{},
		},
		"$set": d,
	}
	opts := options.Update().SetUpsert(true)
	_, err := s.db.Collection("devices").UpdateOne(ctx, bson.M{"uid": d.UID}, q, opts)

	return FromMongoError(err) // <------
}
```
I can't test all the possible returns from `FromMongoError` because I can't control the return value from `s.db.Collection().UpdateOne()`.

However, there are a few observations:
1. In the `TestDeviceListByUsage` test, if we provide a non-existent tenant, should the expected output be an error or an empty list?
2. The `DeviceChoose` method currently updates all devices that have an *accepted* status in a given tenant, barring the ones listed in the *notIn* array. I propose that this method might be more informative if it returns the count of updated devices rather than just signaling an error. As of now, no part of our code utilizes this method.
3. I observed that certain update methods, when they don't modify any documents, the older version of the code doesn't yield an `ErrNoDocuments`. I've made adjustments to address this. You can see these changes in methods like `DeviceRename` and `DeviceCreateTag`.
4. While the `DeviceGet` method provides information about the device tenant, the methods `DeviceGetByName`, `DeviceGetByUID`, `DeviceGetByMac`, and `DeviceLookup` don't seem to do so. Is this correct?

### DIFF
--- a/ui/src/envVariables.ts
+++ b/ui/src/envVariables.ts
@@ -24,5 +24,5 @@ export const envVariables = {
   announcementsEnable: (window.env || process.env).VUE_APP_SHELLHUB_ANNOUNCEMENTS === "true",
   stripeKey: (window.env || process.env).VUE_APP_SHELLHUB_STRIPE_PUBLISHABLE_KEY,
   sentryDsn: (window.env || process.env).VUE_APP_SHELLHUB_SENTRY_DSN_VERSION,
-  googleAnalyticsID: (window.env || process.env).VUE_APP_GOOGLE_ANALYTICS_ID,
+  googleAnalyticsID: (window.env || process.env).VUE_APP_SHELLHUB_GOOGLE_ANALYTICS_ID,
 };


api/pkg/dbtest/dbserver.go
```diff
@@ -67,7 +67,7 @@ type DBServer struct {
 	client  *mongo.Client
 	output  bytes.Buffer
 	server  *exec.Cmd
-	host    string
+	Host    string
```
henrybarreto, line 70: Why `host` must be public here?

PR #3146, opened 2023-10-05
wait for goroutines to finish before exiting the client

heiytor:
sometimes when executing a command, SSH closes the session before writing the command's output to the client's stdout.


otavio (changes requested):
The comment inside the code looks great but it should also be included in commit log.

PR #3190, opened 2023-10-23
refactor(client): remove build URL function and increase, simplify reverse listener creation and package strucuture

henrybarreto:
In addition to that, we had added an extra validation on the request status on each route called by the client, returning a specified error for some HTTP's status mapped, and used context to control the verse listener


pkg/api/client/client.go
```diff
@@ -1,112 +1,231 @@
...
-	httpClient.SetRetryCount(math.MaxInt32)
-	httpClient.AddRetryCondition(func(r *resty.Response, err error) bool {
+type client struct {
+	http   *resty.Client
+	logger *log.Logger
+}
+
+// NewClient creates a new ShellHub HTTP client through the server address.
+//
+// Server address must contain the scheme, the host and the port. For instance: `https://cloud.shellhub.io:80/`.
```
otavio, line 51:
```suggestion
// Server address must contain the scheme, the host and the port. For instance: `https://cloud.shellhub.io:443/`.
```

PR #3201, opened 2023-10-26
fix(api): caching user token

danyecls:
when trying to make login using 'user actions' through cloud, the token generated to login is invalid because is not save in cache, this correction is to save token in cache in the AuthGetToken function, this providing a valid token on the user actions screen

### DIFF
--- a/api/routes/auth_test.go
+++ b/api/routes/auth_test.go
@@ -1,0 +1,543 @@
+package routes
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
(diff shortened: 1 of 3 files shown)


heiytor (commented):
What do you think about making the commit message title more descriptive, maybe something similar to 'fix(api): caching user token'?

danyecls:
> What do you think about making the commit message title more descriptive, maybe something similar to 'fix(api): caching user token'?

I think it's a good idea so I changed the title to the suggested message

otavio (changes requested):
```
fix(api): Ensure token is cached when getting it from service

When trying to log in through the cloud using 'user actions', the
token generated for login is invalid because it is not saved in the
cache. To resolve this issue, we need to save the token in the 
`AuthGetToken` function, so that a valid token is provided on the
user actions screen.
```

otavio (approved):

PR #3241, opened 2023-11-02
fix(agent): avoid agent panic when dial error has no status

henrybarreto:


pkg/api/client/client_public.go
```diff
@@ -96,6 +96,10 @@ func (c *client) Endpoints() (*models.Endpoints, error) {
 func Dial(url string, header http.Header) (*websocket.Conn, *http.Response, error) {
 	conn, res, err := websocket.DefaultDialer.Dial(url, header)
 	if err != nil {
+		if res == nil {
+			return nil, nil, err
```
otavio, line 100: We should add a proper log message here.

PR #3244, opened 2023-11-03
improvement(api): improvement logs to api 

danyecls:
To facilitate debugging and system analysis,
logging was implemented in error cases at the api and pkg shellhub

### DIFF
--- a/api/pkg/dbtest/dbserver.go
+++ b/api/pkg/dbtest/dbserver.go
@@ -36,6 +36,7 @@ import (
 	"time"
 
 	"github.com/shellhub-io/shellhub/pkg/dockerutils"
+	log "github.com/sirupsen/logrus"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
@@ -78,15 +79,16 @@ func (dbs *DBServer) SetTimeout(timeout int) {
 
 func (dbs *DBServer) start() {
 	if dbs.server != nil {
-		panic("DBServer already started")
+		log.Panic("DBServer already started")
 	}
 	l, err := net.Listen("tcp", "127.0.0.1:0")
 	if err != nil {
-		panic("unable to listen on a local address: " + err.Error())
+		log.WithError(err).Panic("unable to listen on a local address")
 	}
+
 	addr, ok := l.Addr().(*net.TCPAddr)
 	if !ok {
-		panic("Type assertion failed")
+		log.Panic("Type assertion failed")
 	}
 
 	l.Close()
@@ -97,7 +99,7 @@ func (dbs *DBServer) start() {
 	if dockerutils.IsRunningInDocker() {
 		containerID, err := dockerutils.CurrentContainerID()
 		if err != nil {
-			panic("failed to get current container id: " + err.Error())
+			log.Panic("failed to get current container id: " + err.Error())
 		}
 
 		if containerID != "" {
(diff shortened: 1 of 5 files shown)


heiytor (commented):
In the code, we import `logrus` as `log`. However, in this commit, different names like `logrus` and `logger` are used in different files. Maybe it would be a good idea to standardize this?

heiytor (changes requested):
import `logrus` as `log`

heiytor (dismissed):

api/pkg/dbtest/dbserver.go
```diff
@@ -122,33 +124,37 @@ func (dbs *DBServer) start() {
...
 		if err := cmd.Run(); err != nil {
+			log.WithError(err).Warning("Failed to list running mongo containers")
+
 			return err
 		}
 
-		fmt.Fprintf(os.Stderr, "----------------------------------------\n")
-
-		panic("mongod container died unexpectedly")
+		log.Info("----------------------------------------")
```
gustavosbarreto, line 156: `log.Error`

api/pkg/dbtest/dbserver.go
```diff
@@ -122,33 +124,37 @@ func (dbs *DBServer) start() {
...
 	if _, err := dbs.server.Process.Wait(); err != nil {
+		log.WithError(err).Warning("mongod container process wait error")
+
 		return err
 	}
 
 	if dbs.tomb.Alive() {
 		// Present some debugging information.
-		fmt.Fprintf(os.Stderr, "---- mongod container died unexpectedly:\n")
+		log.Warning("---- mongod container died unexpectedly ----")
```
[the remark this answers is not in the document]
danyecls: changed to log.Error

api/pkg/dbtest/dbserver.go
```diff
@@ -122,33 +124,37 @@ func (dbs *DBServer) start() {
...
 		fmt.Fprintf(os.Stderr, "%s", dbs.output.Bytes())
-		fmt.Fprintf(os.Stderr, "---- mongod containers running right now:\n")
+		log.Info("---- mongod containers running right now ----")
```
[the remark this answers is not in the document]
danyecls: changed to log.Error

api/pkg/dbtest/dbserver.go
danyecls: changed to log.Error

api/workers/cleaner.go
```diff
@@ -22,15 +21,15 @@ import (
 func StartCleaner(_ context.Context, store store.Store) error {
 	envs, err := getEnvs()
 	if err != nil {
-		return fmt.Errorf("failed to get the envs: %w", err)
+		return err
```
henrybarreto, line 24: I believe we could maintain the errors built, what provider a better context on what's happened

api/workers/cleaner.go
```diff
@@ -22,15 +21,15 @@ import (
...
 	}
 
 	if envs.SessionRecordCleanupRetention == 0 {
 		return nil
 	}
 
 	if envs.SessionRecordCleanupRetention < 0 {
-		return fmt.Errorf("invalid time interval: %w", fmt.Errorf("%d is not a valid time interval", envs.SessionRecordCleanupRetention))
+		return err
```
henrybarreto, line 32: In this case, the `err` does not exist, as nil, or the value does not correspond to the current problem.

I think we can maintain the `fmt.Errorf`.

api/workers/heartbeat.go
```diff
@@ -18,12 +18,12 @@ import (
 func StartHeartBeat(_ context.Context, store store.Store) error {
 	envs, err := getEnvs()
 	if err != nil {
-		return fmt.Errorf("failed to get the envs: %w", err)
+		return err
```
henrybarreto, line 21: Same here

henrybarreto (changes requested):

api/workers/cleaner.go
danyecls: fixed to return with fmt.

henrybarreto (approved):

heiytor (approved):

PR #3263, opened 2023-11-08
fix(agent): set home fallback when user home does not exist

henrybarreto:


pkg/agent/server/command/command_native.go
```diff
@@ -34,7 +34,12 @@ func NewCmd(u *osauth.User, shell, term, host string, command ...string) *exec.C
 		"SHELL=" + shell,
 		"SHELLHUB_HOST=" + host,
 	}
-	cmd.Dir = u.HomeDir
+
+	if _, err := os.Stat(u.HomeDir); err != nil {
+		cmd.Dir = "/"
```
otavio, line 39: We must log so the user knows what is happening.

otavio (changes requested):

pkg/agent/server/command/command_native.go
henrybarreto: Done!

This is an example of message sent:
```WARN[0006] failed to open user's home dir                dir=/home/test error="stat /home/test: no such file or directory"```

File: pkg/agent/server/command/command_native.go [hunk 2]
```diff
@@ -34,7 +35,14 @@ func NewCmd(u *osauth.User, shell, term, host string, command ...string) *exec.C
 		"SHELL=" + shell,
 		"SHELLHUB_HOST=" + host,
 	}
-	cmd.Dir = u.HomeDir
+
+	if _, err := os.Stat(u.HomeDir); err != nil {
+		log.WithError(err).WithField("dir", u.HomeDir).Warn("failed to open user's home dir")
```
otavio, line 40: explain it is falling back to /

otavio (approved):

PR #3269, opened 2023-11-10
[ssh test] Add unit tests for session authentication handlers and enhance logging

heiytor:


ssh/server/auth/publickey_test.go
```diff
@@ -0,0 +1,386 @@
...
+							*ctx = s.Context()
+						},
+					},
+					&gossh.ClientConfig{
+						User: "user@namespace.00-00-00-00-00-00",
+						Auth: []gossh.AuthMethod{
+							gossh.Password("secret"),
+						},
+						HostKeyCallback: gossh.InsecureIgnoreHostKey(),
+					},
```
henrybarreto, line 57: How is possible to test the PublicKeyHandler using only Password on `gossh.AuthMethod`?

heiytor: The test does not test the connection itself. Instead, it passes the session's context directly to the `PublicKeyHandler()` as shown below:
```go
func TestPublicKeyHandler(t *testing.T) {
    // [...]
    for _, tc := range cases {
        t.Run(tc.description, func(t *testing.T) {
            var ctx gliderssh.Context

            srv := tc.setup(&ctx)
            defer srv.Teardown()

            srv.Start()
            assert.NoError(t, srv.Agent.Run("cmd"))

            tc.mocks(ctx)

            result := PublicKeyHandler(ctx, generateTestPubKey(t)) // <----
            assert.Equal(t, tc.expected, result) // <---
        })
    }
}
```
This approach allows us to test the handler as a unit and retrieve the result for use in `assert.Equal()`.

heiytor: It is similar to the way the `PasswordHandler()` is tested:
```go
func TestPasswordHandler(t *testing.T) {
    // [...]
    for _, tc := range cases {
        t.Run(tc.description, func(t *testing.T) {
            var ctx gliderssh.Context

            srv := tc.setup(&ctx)
            defer srv.Teardown()

            srv.Start()
            assert.NoError(t, srv.Agent.Run("cmd"))

            tc.mocks(ctx)

            result := PasswordHandler(ctx, tc.password) // <---
            assert.Equal(t, tc.expected, result) // <---
        })
    }
}
```
This consistency in testing methods helps maintain a standardized approach across different handlers, making the testing process more organized.

PR #3316, opened 2023-11-27
refactor(connman): restructure ConnectionManager and introduce SyncSliceMap

gustavosbarreto:
This commit refactors the `ConnectionManager` by replacing the `sync.Map`
for dialers with a `SyncSliceMap`.

The `SyncSliceMap` struct provides methods for loading, storing, and
deleting values in a synchronized map with slices.

The use of `SyncSliceMap` improves the management of reverse dialers and
handles multiple connections for the same identifier.

This change potentially addresses the SSH connection issue where multiple
agents using the same identifier prevented SSH connection, even when the
web interface showed them as online.

### DIFF
--- a/pkg/connman/connman.go
+++ b/pkg/connman/connman.go
@@ -4,21 +4,22 @@ import (
 	"context"
 	"errors"
 	"net"
-	"sync"
 
 	"github.com/shellhub-io/shellhub/pkg/revdial"
+	"github.com/sirupsen/logrus"
 )
 
 var ErrNoConnection = errors.New("no connection")
 
 type ConnectionManager struct {
-	dialers                 sync.Map
+	dialers                 *SyncSliceMap
 	DialerDoneCallback      func(string, *revdial.Dialer)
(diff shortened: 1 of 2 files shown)


pkg/connman/connman.go
```diff
@@ -27,7 +28,12 @@ func New() *ConnectionManager {
 func (m *ConnectionManager) Set(key string, conn net.Conn) {
 	dialer := revdial.NewDialer(conn, "/ssh/revdial")
 
-	m.dialers.Store(key, dialer)
+	if size := m.dialers.Store(key, dialer); size > 1 {
+		logrus.WithFields(logrus.Fields{
+			"key":  key,
+			"size": size,
+		}).Warning("Multiple connections stored for the same identifier.")
```
otavio, line 35: Other important fields to add to the log?

gustavosbarreto: In this context, there are no additional fields that can be added to the log

PR #3343, opened 2023-12-06
chore(api,cli,pkg): remove deprecated function on validator package

henrybarreto:


pkg/models/user.go
```diff
@@ -13,7 +15,45 @@ type UserData struct {
 }
 
 type UserPassword struct {
-	Password string `json:"password" bson:",omitempty" validate:"required,min=5,max=30"`
+	// Plain contains the plain text password.
+	Plain string `json:"password" bson:"-" validate:"required,password"`
+	// Password contains the hashed pasword from plain text.
+	Password string `json:"-" bson:"password"`
```
heiytor, line 21: Isn't it a good idea to call this 'Hash' instead of 'Password'?

otavio: `HashedPassword`?

henrybarreto: Done. It is now `PlainPassword` and `HashedPassword`.

api/pkg/echo/handlers/validator.go
```diff
@@ -18,13 +18,8 @@ func NewValidator() *Validator {
...
 func (v *Validator) Validate(structure interface{}) error {
 	// Use the ShellHub package validator to validate the request body.
-	if ok, err := v.validator.Struct(structure); !ok {
-		fields, err := validator.GetInvalidFieldsFromErr(structure, err)
-		if err != nil {
-			return err
-		}
-
-		return errors.NewErrInvalidEntity(fields)
+	if ok, err := v.validator.Struct(structure); !ok || err != nil {
```
otavio, line 21: The word 'deprecated' has been mistyped.

PR #3403, opened 2024-01-03
refactor update device status functionality in the service

heiytor:


api/services/device.go
```diff
@@ -183,11 +183,6 @@ func (s *service) OffineDevice(ctx context.Context, uid models.UID, online bool)
 
 // UpdateDeviceStatus updates the device status.
 func (s *service) UpdateDeviceStatus(ctx context.Context, tenant string, uid models.UID, status models.DeviceStatus) error {
-	namespace, err := s.store.NamespaceGet(ctx, tenant)
```
henrybarreto, line 186: I think that verifying the namespace existence before trying to do operations is a better approach.

heiytor: In my view, retaining the namespace for an extended period without use requires greater cognitive effort. It may lead to questioning how the namespace will eventually be used, and by the time its is used, we may have already forgotten where and how it was originally defined. I also believe that this practice introduces additional complexity to the code, making it more challenging to read and understand.

henrybarreto: We fetch the namespace first to confirm it exists before dealing with the device itself.

heiytor: Ah, my mistake. I didn't understand it. I will roll back the change.

otavio: As it was unclear, add a comment.

PR #3419, opened 2024-01-09
fix(api): return the right error when no resource is matched

henrybarreto:


api/store/mongo/user.go
```diff
@@ -213,7 +213,7 @@ func (s *Store) UserUpdatePassword(ctx context.Context, newPassword string, id s
 		return FromMongoError(err)
 	}
 
-	if user.ModifiedCount < 1 {
+	if user.MatchedCount < 1 {
```
otavio, line 216: Add tests please.

PR #3427, opened 2024-01-11
fix(api): add verification if the user with active MFA has the MFA va…

danyecls:
…lidated


pkg/models/user.go
```diff
@@ -100,7 +100,8 @@ type UserAuthResponse struct {
 	Tenant string `json:"tenant"`
 	Role   string `json:"role"`
 	Email  string `json:"email"`
-	MFA    bool   `json:"mfa" bson:"mfa"`
+	// MFA    bool   `json:"mfa" bson:"mfa"`
```
gustavosbarreto, line 103: Please remove this comment

PR #3449, opened 2024-01-16
fix(api): correct binding of query parameters

heiytor:
This commit addresses two issues:
1. Ensures proper percent encoding in query parameters.
2. Enables the API to retrieve query parameters in POST, PUT, and PATCH routes.


otavio (changes requested):
Add tests please.

PR #3501, opened 2024-01-31
fix device update

### LINKED ISSUE #3500: Renaming a device is inconsistent
### Description

I noticed that renaming a device using `PUT /api/devices/UID` isn't consistent.

Sometimes, the device will not be renamed. Sometimes a 409 will be thrown when you try to rename a device with a name already used, but not always.

To reproduce, just have a shellhub instance with one instance on it and spam a rename with a new name on it. The name will not change directly, and once it finally does, then it will not always throw a 409

### Edition

Community

### Version

0.14.1

heiytor:
fixes #3500

### DIFF
--- a/api/store/mongo/device.go
+++ b/api/store/mongo/device.go
@@ -539,28 +539,19 @@ func (s *Store) DeviceChooser(ctx context.Context, tenantID string, chosen []str
 // DeviceChooser updates devices with "accepted" status to "pending" for a given tenantID,
 // excluding devices with UIDs present in the "notIn" list.
 func (s *Store) DeviceUpdate(ctx context.Context, tenant string, uid models.UID, name *string, publicURL *bool) error {
-	session, err := s.db.Client().StartSession()
-	if err != nil {
-		return err
+	changes := bson.M{}
+
+	if name != nil {
+		changes["name"] = *name
 	}
 
-	defer session.EndSession(ctx)
+	if publicURL != nil {
+		changes["public_url"] = *publicURL
+	}
 
-	err = mongo.WithSession(ctx, session, func(sessionContext mongo.SessionContext) error {
-		if name != nil {
-			if _, err := s.db.Collection("devices").UpdateOne(sessionContext, bson.M{"tenant_id": tenant, "uid": uid}, bson.M{"$set": bson.M{"name": *name}}); err != nil {
-				return err
-			}
-		}
-
-		if publicURL != nil {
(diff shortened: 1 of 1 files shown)


api/routes/device.go
```diff
@@ -322,9 +322,10 @@ func (h *Handler) UpdateDevice(c gateway.Context) error {
 		tenant = c.Tenant().ID
 	}
 
-	if err := guard.EvaluatePermission(c.Role(), guard.Actions.Device.Update, func() error {
+	err := guard.EvaluatePermission(c.Role(), guard.Actions.Device.Update, func() error {
 		return h.service.UpdateDevice(c.Ctx(), tenant, models.UID(req.UID), req.Name, req.PublicURL)
-	}); err != nil {
+	})
+	if err != nil {
```
henrybarreto, line 328: Why do you have preferred changed how the if was structured?

heiytor: I strongly believe that the syntax of `if err := ...; err != nil {}` becomes highly illegible when the `[..]; err != nil {}` statement is placed on another line or when the function call is too long. In my opinion, the eyes get lost when trying to understand the flow of the code.

heiytor: This has been my experience, and it is why I consider it a good idea to change this.

otavio: I believe binding it to the `if` statement also reduces the variable's scope. I prefer the original code (even if it's uglier).

henrybarreto: > I believe binding it to the `if` statement also reduces the variable's scope. I prefer the original code (even if it's uglier).

I agree. This change could fit perfectly if we needed the error in the following, but we don't.

api/store/mongo/device.go
```diff
@@ -536,31 +536,26 @@ func (s *Store) DeviceChooser(ctx context.Context, tenantID string, chosen []str
...
-// excluding devices with UIDs present in the "notIn" list.
 func (s *Store) DeviceUpdate(ctx context.Context, tenant string, uid models.UID, name *string, publicURL *bool) error {
-	session, err := s.db.Client().StartSession()
-	if err != nil {
-		return err
+	// Not deleting the device from the cache may cause issues when trying to retrieve the device after the update.
+	// TODO: Maybe we can standardize the key creation?
+	if err := s.cache.Delete(ctx, strings.Join([]string{"device", string(uid)}, "/")); err != nil {
+		logrus.Error(err)
 	}
```
henrybarreto, line 544: Delete the cache is an excellent idea, but I think that we should do it after a successful update only

api/routes/device.go
heiytor: I understand your concerns; I'm rolling back this commit.

henrybarreto (approved):

PR #3509, opened 2024-02-01
fix(agent): don't send trash to SFTP when agent is running on host

henrybarreto:


pkg/agent/server/modes/host/sessioner.go
```diff
@@ -406,12 +407,20 @@ func (s *Sessioner) SFTP(session gliderssh.Session) error {
...
+					"user": session.Context().User(),
+				}).Error("failed when reading the error output from sftp process")
 
-			return
+				return
+			}
+
+			log.WithFields(log.Fields{
+				"user": session.Context().User(),
+			}).Info(msgs.Text())
```
gustavosbarreto, line 423: Could we assume that everything coming from stderr is an .Error?

PR #3517, opened 2024-02-05
tests(agent,ssh): add integration tests to SSH connection

henrybarreto:
In this PR, we are adding the initial integration testes between Agent and ShellHub instance.

- [x] Connect to the server.
- [x] Connect to server with custom identity
- [x] Reconnect to it after losing its connection.
- [x] Authentication
	- [x] Password.
		- [x] With custom identity
	- [x] Public key.
- [x] SSH connections
	- [x] Pipeing shell and exec request on same connection
	- [x] Interactive shell.
		- [x] Connection
		- [x] With Pty
		- [x] Without Pty
	- [x] Command execution.
		- [x] Exit status.
		- [x] Stdin, Stdout, Stderror
			- [x] With custom identity
	- [x] SFTP.
		- [x] Lagacy mode (SCP).
			- [x] From Agent to client.
			  Error when download the file; seems to be some package internal.
			- [x] From client to Agent.
		- [x] From Agent to client.
		- [x] From client to Agent.
		
## Running it

To run tests, I use this command:

```
go test -v --failfast -shuffle on -timeout 1h ./tests -run "TestAgent"
```

Timeout is set to avoid the default one that is `600` second, `--failfast` to help to spare time and `-shuffle` to guarantee false positive due to a before test, and `-parallel` to define how many tests can run in the same time.


tests/agent_test.go
```diff
@@ -0,0 +1,235 @@
...
+func assertLogsContainsString(t *testing.T, logs io.Reader, expected string) {
+	t.Helper()
+
+	// FIXME: Is there a better strategy to read the logs?
+	buf := new(bytes.Buffer)
+
+	_, err := buf.ReadFrom(logs)
+	assert.NoError(t, err)
+
+	text := buf.String()
```
heiytor, line 27: I think you can use a `strings.Builder` instead of `bytes.Buffer`, which can be more efficient and performant.

1. https://golangnote.com/golang/golang-stringsbuilder-vs-bytesbuffer
2. https://medium.com/@felipedutratine/string-concatenation-in-golang-since-1-10-bytes-buffer-vs-strings-builder-2b3081848c45

PR #3518, opened 2024-02-05
feature(pkg): add log format to loglevel package

henrybarreto:


henrybarreto:
It is worth saying that this PR doesn't change the current behavior of the logs, only adds the possibility to change it through environmental variables.

otavio (approved):

pkg/loglevel/loglevel.go
```diff
@@ -7,27 +7,30 @@ import (
 
 // UseEnvs set the logger level to the specified in SHELLHUB_LOG_LEVEL and the log format for SHELLHUB_LOG_FORMAT.
 func UseEnvs() {
-	SetLogLevel()
 	SetLogFormat()
+	SetLogLevel()
```
otavio, line 11: Part of other commit.

PR #3519, opened 2024-02-06
refactor namespace update route

heiytor:
Introducing a new method, `NamespaceEdit`, within the store to
facilitate the updating of a namespace's attributes. This new method
replaces the deprecated `NamespaceRename` and `NamespaceSetSessionRecord`.

Accompanying this change is the introduction of a new struct,
`NamespaceChanges`, designed to be a parameter for the `NamespaceEdit`
method. This struct consolidates all possible attributes that users can
modify. The use of the tag `bson:"omitempty"` helps avoid unnecessary
and undesirable `if` statements by handling empty values appropriately.

In the service layer, the method `EditNamespace` now calls the store
method with the specified changes. This method now takes the entire
request and determines its course of action (in this case, creating the
`changes` struct). The validation process within the service is removed
to centralize it in the router, streamlining code complexity.

As part of deprecating `NamespaceRename`, we're also renaming the
permission to `Namespace.Update` and `NamespaceUpdate`.


pkg/api/requests/namespace.go
```diff
@@ -34,7 +34,10 @@ type NamespaceDelete struct {
 // NamespaceEdit is the structure to represent the request data for edit namespace endpoint.
 type NamespaceEdit struct {
 	TenantParam
-	Name string `json:"name"  validate:"required,hostname_rfc1123,excludes=."`
+	Name     string `json:"name"  validate:"omitempty,hostname_rfc1123,excludes=."`
```
henrybarreto, line 37: Is `omitempty` a valid `validate` tag?

heiytor: Yes, you read about [here](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Omit_Empty).

api/services/namespace.go
```diff
@@ -180,28 +180,18 @@ func (s *service) fillMembersData(ctx context.Context, members []models.Member)
...
+		SessionRecord: req.Settings.SessionRecord,
 	}
 
-	name = strings.ToLower(name)
-	if ok, err := s.validator.Struct(&models.Namespace{Name: name}); !ok || err != nil {
-		return nil, NewErrNamespaceInvalid(err)
-	}
-
-	if namespace.Name == name {
-		return nil, NewErrNamespaceDuplicated(nil)
```
henrybarreto, line 201: I think we should check if already exist a namespace with the desired name.

api/services/namespace.go
```diff
@@ -180,28 +180,18 @@ func (s *service) fillMembersData(ctx context.Context, members []models.Member)
...
-// EditNamespace edits the namespace name.
-//
-// It receives a context, used to "control" the request flow,  tenant ID from models.Namespace and the new name to
-// namespace. Name is set to lowercase.
-//
-// EditNamespace returns a models.Namespace and an error. When error is not nil, the models.Namespace is nil.
-func (s *service) EditNamespace(ctx context.Context, tenantID, name string) (*models.Namespace, error) {
-	namespace, err := s.store.NamespaceGet(ctx, tenantID)
-	if err != nil {
-		return nil, NewErrNamespaceNotFound(tenantID, err)
```
henrybarreto, line 192: I believe keeping the namespace get to guarantee the namespaces existence facilitates the reading process

api/services/namespace.go
heiytor: In this context, this namespace was retrieved with the tenantId, which implies (to my understanding) that this snippet is not checking for duplications, but only verifying if the desired name matches the current name.

I understand and agree that the code needs to ensure a unique name. However, since the current state of the code does not accomplish that, I would prefer to simply remove this "useless" snippet, as MongoDB already handles situations where the name is the same.

api/services/namespace.go
heiytor: Due to the fact that this method returns the updated namespace, I believe it is unnecessary to search for the namespace twice, leading to the following (excessive in my opinion) flow:

1. Retrieve the namespace solely to check for its existence.
2. Update the namespace.
3. Retrieve the updated namespace.

There may also be a third find operation if searching for a duplicated name is necessary.

I also believe it's not overly complicated to understand that updating a namespace assumes it already exists, potentially returning an error if it does not. Additionally, I think that we can leverage the power of MongoDB, which can handle this with only one operation.

henrybarreto (approved):

PR #3529, opened 2024-02-09
feature(agent,pkg): validate required env variable when it is empty 

henrybarreto:


pkg/envs/backend_env.go
```diff
@@ -2,11 +2,14 @@ package envs
 
 import (
 	"context"
+	"errors"
 	"os"
 
 	"github.com/sethvargo/go-envconfig"
 )
 
+var ErrParseRequiredEmpty = errors.New("environmental variable set cannot be empty")
```
otavio, line 11:
```suggestion
var ErrParseRequiredEmpty = errors.New("environmental variable cannot be empty")
```

otavio (dismissed):

agent/main.go
```diff
@@ -57,6 +58,10 @@ func main() {
 				log.Fatal(err)
 			}
 
+			if ok, fields, err := validator.New().StructWithFields(cfg); err != nil || !ok {
+				log.WithError(err).WithFields(fields).Fatal("failed to validate the environmental variables")
+			}
```
otavio, line 63: Please add tests.

otavio (changes requested):

PR #3559, opened 2024-02-19
refactoring internal client package

heiytor:
The internalclient package now features improved organization with
clearer distinctions between files and interfaces. A full documentation
are also provided


pkg/api/internalclient/client.go
```diff
@@ -94,72 +101,29 @@ func NewClientWithAsynq(uri string) Client {
...
+// It accepts a format string for the URI where placeholders can be replaced by the provided arguments.
+// For example, both of the following calls are valid:
+//
+//	client.buildURL(fmt.Sprintf("/hello/%s", "world"))
+//	client.buildURL("/hello/%s", "world")
+//
+// In both cases, the resulting URL will be constructed as follows:
+//
+//	{scheme}://{host}:{port}/hello/world
+func (c *client) buildURL(format string, args ...any) string {
```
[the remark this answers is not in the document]
henrybarreto: The external client uses something like it here:
https://github.com/shellhub-io/shellhub/blob/master/pkg/api/client/client.go#L62

pkg/api/internalclient/billing.go
```diff
@@ -0,0 +1,46 @@
...
+	// It returns the billing evaluation result, the HTTP status code of the response, and an error, if any.
+	BillingEvaluate(tenantID string) (*models.BillingEvaluation, int, error)
+}
+
+func (c *client) BillingReport(tenant string, action string) (int, error) {
+	res, err := c.http.
+		R().
+		SetHeader("X-Tenant-ID", tenant).
+		SetQueryParam("action", action).
+		Post("http://billing-api:8080/internal/billing/report")
```
henrybarreto, line 25: Would you mind confirming if the request built won't be something like `http://api:8080/http://billing-api:8080/internal/billing/report` due to usage of the same client what has `http://api:8080` as URL?

heiytor: According to the [official documentation](https://pkg.go.dev/github.com/go-resty/resty/v2#Client.SetBaseURL), SetBaseURL is only intended for relative URLs.

pkg/api/internalclient/firewall.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (c *client) FirewallEvaluate(lookup map[string]string) error {
+	local := resty.New()
+	local.AddRetryCondition(func(r *resty.Response, err error) bool {
+		if _, ok := err.(net.Error); ok {
+			return true
+		}
+
+		return r.StatusCode() >= http.StatusInternalServerError && r.StatusCode() != http.StatusNotImplemented
+	})
```
henrybarreto, line 31: I believe we should do something like this on billing requests, no?

heiytor: I haven't made any changes to the implementation within the existing methods. The way `BillingReport` and `BillingEvaluate` function remains the same as before these commits.

pkg/api/internalclient/billing.go
henrybarreto: Exactly, but you are using the same client as you have set the URL, so It could end up with something like what I have mentioned.

pkg/api/internalclient/firewall.go
henrybarreto: You are right, but the usage of base URL could cause a small issue in URL on `billing` routes. Firewall ones are immune to it because it creates it own client.

pkg/api/internalclient/billing.go
heiytor: I understand, but even when using the same client, the BaseURL should only work with relative URLs:
> It will be used with request raised from this client **with** relative URL

heiytor: ust to confirm, I've conducted some tests, and yes, it does not use the base URL when the request is something like `http://billing-api:8080/internal/billing/report`~.

pkg/api/internalclient/firewall.go
heiytor: Since the base URL won't cause this issue, I don't believe creating a new client here is a good approach. It would result in duplicating some code snippets in four places, which could lead to minimal bugs that are difficult to identify if changes are made in one place but not in another. Additionally, I also believe that the firewall should utilize the client's HTTP instead of creating a new one.

pkg/api/internalclient/billing.go
henrybarreto: I have checked and the problem I have mention seems does not happen. Perfect!

pkg/api/internalclient/sshkey.go
```diff
@@ -0,0 +1,70 @@
...
+		R().
+		SetResult(&pubKey).
+		Get(fmt.Sprintf("/internal/sshkeys/public-keys/%s/%s", fingerprint, tenant))
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode() == 404 {
+		return nil, ErrNotFound
+	}
```
henrybarreto, line 34: Are your idea standardizing the errors from the internal client?

We have something like this too.

https://github.com/shellhub-io/shellhub/blob/master/pkg/api/client/errors.go#L40

Issue #3639, opened 2024-03-23
The gateway container keeps timing out waiting for the ACME server on first start


merlinran:
### Description

It's when `SHELLHUB_AUTO_SSL` is set. In my case, this line
https://github.com/shellhub-io/shellhub/blob/12e51169339893368c58fd156210d5951f55a5ea/gateway/entrypoint.sh#L71
 starts the server on `0.0.0.0`, while `wait_for_acme_webserver` https://github.com/shellhub-io/shellhub/blob/12e51169339893368c58fd156210d5951f55a5ea/gateway/entrypoint.sh#L33  in my case, gets `localhost` resolved to `::1` and never tries the IPv4 one.

I guess the simplest fix here could be replacing `nc` with `curl -sSf http://localhost > /dev/null`, which tries all resolved IPs.

### Edition

Community

### Version

v0.14.2

merlinran:
Alternatively maybe replace NGINX with Caddy or something with ACME support, to simplify things. https://www.eff.org/deeplinks/2024/03/should-caddy-and-traefik-replace-certbot

gustavosbarreto:
> ### Description
> It's when `SHELLHUB_AUTO_SSL` is set. In my case, this line
> 
> https://github.com/shellhub-io/shellhub/blob/12e51169339893368c58fd156210d5951f55a5ea/gateway/entrypoint.sh#L71
> 
> 
> starts the server on `0.0.0.0`, while `wait_for_acme_webserver`
> https://github.com/shellhub-io/shellhub/blob/12e51169339893368c58fd156210d5951f55a5ea/gateway/entrypoint.sh#L33
> 
> in my case, gets `localhost` resolved to `::1` and never tries the IPv4 one.
> I guess the simplest fix here could be replacing `nc` with `curl -sSf http://localhost > /dev/null`, which tries all resolved IPs.
> 
> ### Edition
> Community
> 
> ### Version
> v0.14.2

@merlinran Could you create a pull request with the proposed fix?  Since replacing NGINX with Caddy or any other API gateway isn't on the roadmap right now due to the significant effort involved.

merlinran:
> ... replacing NGINX with Caddy or any other API gateway isn't on the roadmap right now due to the significant effort involved.

Is this change desirable? Is there any downsides or caveats of doing this? I can work on that if I get a :+1: Do you have any preference on Caddy vs Traefik vs (any other options)?

gustavosbarreto:
[...]

PR #3647, opened 2024-03-25
use curl to test availability of ACME server, which tries all resolved IPs

merlinran:
Fixes https://github.com/shellhub-io/shellhub/issues/3639

Tested in my `gateway` container.
```
docker compose exec -ti gateway /bin/sh
...
/app # nc -z localhost 80; echo $?
1
/app # nc -z 127.0.0.1 80; echo $?
0
/app # nc -z ::1 80; echo $?
1
/app # curl -sSf http://localhost > /dev/null; echo $?
0
```

Even with this fix, it may still fail to get the certificate, if the host has only IPv6 address, because this line listens only on IPv4 somehow.
https://github.com/shellhub-io/shellhub/blob/12e51169339893368c58fd156210d5951f55a5ea/gateway/entrypoint.sh#L71

Even though [the Python3 doc](https://docs.python.org/3/library/http.server.html) says
> By default, the server binds itself to all interfaces. The option -b/--bind specifies a specific address to which it should bind. Both IPv4 and IPv6 addresses are supported.


gustavosbarreto:
Merged! Thanks @merlinran

PR #3658, opened 2024-03-28
fix(ssh): allow multiplexed connection

### LINKED ISSUE #3656: Multiplexed Connections Rejected with 'Shared Connection Closed'
### Description

# Description

ShellHub is rejecting connections when the client is configured to use [multiplexed connections](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Multiplexing).

# Steps to Reproduce

1. Enable multiplexed connections.
2. Authenticate to a device.
3. Attempt to reauthenticate to the same device with the previous connection still open.

> Note: Connections using `-f` will succeed in connecting the next attempt, but subsequent connections attempts will fail.

# Expected Behavior

The client should reuse the existing "master" connection, bypassing, for example, the authentication step.

# Actual Behavior

The server rejects subsequent connections with a `Shared connection to localhost closed.` error.

___

See also  #3653

### Edition

Community

### Version

v0.14.3

henrybarreto:
A Multiplexing connection can send more than one channel request over the same socket, making possible reuse the authentication method used by the master in the derivate ones.

To support it, we should only allow the GliderSSH's server to dial with the connection on the agent's side, removing our intervention on the protocol flow, that enforces the channel limitation, just sending the device's response to the client.

Closes #3656

### DIFF
--- a/ssh/server/channels/session.go
+++ b/ssh/server/channels/session.go
@@ -175,21 +175,6 @@ func DefaultSessionHandler(opts DefaultSessionHandlerOptions) gliderssh.ChannelH
 
 				switch req.Type {
 				case ShellRequestType, ExecRequestType, SubsystemRequestType:
-					// Once the session has been set up, a program is started at the remote end.  The program can be a
-					// shell, an application program, or a subsystem with a host-independent name.  **Only one of these
-					// requests can succeed per channel.**
-					//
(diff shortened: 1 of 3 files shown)


heiytor (approved):

henrybarreto:
Delayed for another moment

PR #3759, opened 2024-04-23
switch to bulk writes in DeviceSetOnline

heiytor:
The `DeviceClaims` now includes a `Tenant` attribute. The `AuthDevice` will utilize these attribute to populate the `X-Tenant-ID`. Furthermore, the `/ssh/connection` endpoint has been updated to recognize and utilize these headers for identifying the device's tenant.

Additionaly `DeviceSetOnline` method will now use bulk writes instead of single operations. This change should improve performance and reduce MongoDB I/O needs since the method is exclusively used in the `api:heartbeat` queue, which processes a list of devices to set as online. As this method's sole purpose is to set devices as online, it will no longer accept a boolean parameter to set a device as offline. Instead, a new method called `DeviceSetOffline` has been added.


api/services/device.go
```diff
@@ -176,13 +176,17 @@ func (s *service) LookupDevice(ctx context.Context, namespace, name string) (*mo
...
-	if err == store.ErrNoDocuments {
-		return NewErrDeviceNotFound(uid, err)
+func (s *service) OffineDevice(ctx context.Context, uid models.UID) error {
+	if err := s.store.DeviceSetOffline(ctx, string(uid)); err != nil {
+		switch {
+		case errors.Is(err, store.ErrNoDocuments):
+			return NewErrDeviceNotFound(uid, err)
+		default:
+			return err
+		}
```
otavio, line 186:
```suggestion
func (s *service) OffineDevice(ctx context.Context, uid models.UID) error {
  err := s.store.DeviceSetOffline(ctx, string(uid))
	if errors.Is(err, store.ErrNoDocuments) {
			return NewErrDeviceNotFound(uid, err)
	}
	
	return err
```

otavio (changes requested):

otavio (approved):

PR #3760, opened 2024-04-23
fix(api): use device name custom rule for field validation

### LINKED ISSUE #3291: Containers with underscore (_) in the name not appearing in ShellHub
### Description

It seems that the hostname of the `SSHID` does not support the underscore (_) character.

### Edition

Community

### Version

master

henrybarreto:
We have been using the `hostname_rfc1123` for validating the device name, which isn't compatible with [Docker's daemon naming convention](https://github.com/moby/moby/blob/master/daemon/names/names.go#L6), causing containers with `underscore` to be denied by the server for invalid entity format.

Closes #3291

### DIFF
--- a/pkg/api/requests/device.go
+++ b/pkg/api/requests/device.go
@@ -79,7 +79,7 @@ type DeviceInfo struct {
 type DeviceAuth struct {
 	Info      *DeviceInfo     `json:"info" validate:"required"`
 	Sessions  []string        `json:"sessions,omitempty"`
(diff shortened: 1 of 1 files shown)


henrybarreto:
The `device_name` means it on our code base can be found here:

https://github.com/shellhub-io/shellhub/blob/master/pkg/validator/validator.go#L72-L78

henrybarreto:
Our `device_name` rule can be found here:

https://github.com/shellhub-io/shellhub/blob/master/pkg/validator/validator.go#L72-L78

PR #3887, opened 2024-06-14
tests(api): add tests to api gateway

haller33:
little tests added

### DIFF
--- a/api/pkg/gateway/context_test.go
+++ b/api/pkg/gateway/context_test.go
@@ -1,0 +1,416 @@
+package gateway
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/labstack/echo/v4"
+	"github.com/shellhub-io/shellhub/pkg/api/authorizer"
+	"github.com/shellhub-io/shellhub/pkg/models"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRole(t *testing.T) {
+	cases := []struct {
+		description string
+		headers     map[string]string
+		expected    authorizer.Role
+	}{
+		{
+			description: "verify if given value returns from header",
+			headers: map[string]string{
+				"X-Role": authorizer.RoleOwner.String(),
+			},
+			expected: authorizer.Role("owner"),
+		}, {
+			description: "validate empty headers, for fail function",
+			headers:     map[string]string{},
+			expected:    authorizer.RoleInvalid,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.description, func(t *testing.T) {
+			e := echo.New()
+			req := httptest.NewRequest(http.MethodPost, "/", nil)
+
+			for k, v := range tc.headers {
+				req.Header.Set(k, v)
+			}
+
+			rec := httptest.NewRecorder()
+			ctx := e.NewContext(req, rec)
+
+			ctxNew := Context{
+				nil,
+				ctx,
+			}
+
+			role := ctxNew.Role()
+
+			require.Equal(t, tc.expected, role)
+		})
+	}
+}
+
+func TestTenant(t *testing.T) {
+	cases := []struct {
+		description string
+		headers     map[string]string
+		expected    *models.Tenant
+	}{
+		{
+			description: "verify if given value returns from header",
+			headers: map[string]string{
+				"X-Tenant-ID": "00000000-0000-4000-0000-000000000000",
+			},
+			expected: &models.Tenant{
+				ID: "00000000-0000-4000-0000-000000000000",
+			},
+		}, {
+			description: "validate empty headers, for fail function",
+			headers:     map[string]string{},
+			expected:    nil,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.description, func(t *testing.T) {
+			e := echo.New()
+			req := httptest.NewRequest(http.MethodPost, "/", nil)
+
+			req.Header.Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
+
+			for k, v := range tc.headers {
+				req.Header.Set(k, v)
+			}
+
+			rec := httptest.NewRecorder()
+			ctx := e.NewContext(req, rec)
+
+			ctxNew := Context{
+				nil,
+				ctx,
+			}
+
+			tenant := ctxNew.Tenant()
+
+			require.Equal(t, tc.expected, tenant)
+		})
+	}
+}
+
+func TestUsername(t *testing.T) {
+	cases := []struct {
+		description string
+		headers     map[string]string
+		expected    *models.Username
+	}{
+		{
+			description: "verify if given value returns from header",
+			headers: map[string]string{
+				"X-Username": "someone",
+			},
+			expected: &models.Username{
+				ID: "someone",
+			},
+		}, {
+			description: "validate empty headers, for fail function",
+			headers:     map[string]string{},
+			expected:    nil,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.description, func(t *testing.T) {
+			e := echo.New()
+			req := httptest.NewRequest(http.MethodPost, "/", nil)
+
+			req.Header.Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
+
+			for k, v := range tc.headers {
+				req.Header.Set(k, v)
+			}
+
+			rec := httptest.NewRecorder()
+			ctx := e.NewContext(req, rec)
+
+			ctxNew := Context{
+				nil,
+				ctx,
+			}
+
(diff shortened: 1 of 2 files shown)


api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,599 @@
+package gateway
+
+import (
+	//	"fmt"
```
gustavosbarreto, line 4: Please remove the commented code

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,599 @@
...
+	"testing"
+
+	"github.com/labstack/echo/v4"
+
+	"github.com/shellhub-io/shellhub/pkg/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRole(t *testing.T){
+
```
gustavosbarreto, line 16: Unnecessary new line

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,599 @@
...
+			retRoleCtx := gatewayCtxT.Role()
+
+			assert.Equal(t, tc.expected.retValue, retRoleCtx)
+		})
+	}
+}
+
+func TestTenant(t *testing.T){
+
```
gustavosbarreto, line 81: Unnecessary new line

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,599 @@
...
+			retTenantCtx := gatewayCtxT.Tenant()
+
+			assert.Equal(t, tc.expected.retValue, retTenantCtx)
+		})
+	}
+}
+
+func TestUsername(t *testing.T){
+
```
gustavosbarreto, line 148: Unnecessary new line

henrybarreto (dismissed):

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,541 @@
...
+			gatewayCtxT := Context{
+				nil,
+				ctxEchoT,
+			}
+
+			retStringRole, retBoolRoleValidate := gatewayCtxT.GetRole()
+
+			require.Equal(t, tc.expected.role, retStringRole)
+			require.Equal(t, tc.expected.ok, retBoolRoleValidate)
```
heiytor, line 538: You can compare the entire struct instead of its individual attributes. Change this to:
```go
require.Equal(t, tc.expected, Expected{retStringRole, retBoolRoleValidate})
```

api/pkg/gateway/gateway_test.go
```diff
@@ -0,0 +1,214 @@
...
+			retTenantfctx := TenantFromContext(ctx)
+
+			require.Equal(t, tc.expected.tenantID, retTenantfctx)
+		})
+	}
+}
+
+func TestUsernameFromContext(t *testing.T) {
+	type Expected struct {
+		userNameTenant *models.Username
```
heiytor, line 84: Why `userNameTenant`?

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,541 @@
...
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.description, func(t *testing.T) {
+			e := echo.New()
+			req := httptest.NewRequest(http.MethodPost, "/", nil)
+
+			req.Header.Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
```
heiytor, line 57: I believe that since you're not sending a body, there's no need to set the `Content-Type` header.

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,541 @@
...
+
+			rec := httptest.NewRecorder()
+			ctxEchoT := e.NewContext(req, rec)
+
+			gatewayCtxT := Context{
+				nil,
+				ctxEchoT,
+			}
+
+			retGetTenantFst, retGetTenantSnd := gatewayCtxT.GetTennat()
```
heiytor, line 401: What are `fst` and `snd`?

heiytor:
I also noticed that you use `ret` as a prefix for many return values. As far as I know, this is not a common Go pattern, so I believe it would be a good idea to remove it. Additionally, you frequently use `T` as a postfix. Could you explain what this means?

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,541 @@
...
+
+			rec := httptest.NewRecorder()
+			ctxEchoT := e.NewContext(req, rec)
+
+			gatewayCtxT := Context{
+				nil,
+				ctxEchoT,
+			}
+
+			retTenantID := gatewayCtxT.Tenant()
```
henrybarreto, line 136: As mentioned by @heiytor, the `ret` prefix isn't necessary and doesn't follow the Golang pattern for variables in general. I believe it could be simple `tenant` or `tenantID` without cognitive prejudice.

You could check these pattern here: https://go.dev/doc/effective_go#variables

henrybarreto (commented):
Beyond what already said, I recommend the [Effective Go](https://go.dev/doc/effective_go) to give a better overview of the Golang patterns and recommendations.

api/pkg/gateway/gateway_test.go
haller33: I'll be changed, it has been because of an old design decision

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,541 @@
...
+	"github.com/labstack/echo/v4"
+
+	"github.com/shellhub-io/shellhub/pkg/api/authorizer"
+	"github.com/shellhub-io/shellhub/pkg/models"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRole(t *testing.T) {
+	type Expected struct {
+		roleString authorizer.Role
```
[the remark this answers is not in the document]
haller33: sure! understood and agree

api/pkg/gateway/context_test.go
haller33: Understood

haller33:
> I also noticed that you use `ret` as a prefix for many return values. As far as I know, this is not a common Go pattern, so I believe it would be a good idea to remove it. Additionally, you frequently use `T` as a postfix. Could you explain what this means?

this has been a tag name for an abbreviation of 'Tuple' an old intermediate struct created by myself that has been used and later on, removed, and this T has been as a forgotten code, that is now removed

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,541 @@
...
+		headers     map[string]string
+		expected    Expected
+	}{
+		{
+			description: "verify if given value returns from header",
+			headers: map[string]string{
+				"X-Role": "owner",
+			},
+			expected: Expected{
+				roleString: authorizer.Role("owner"),
```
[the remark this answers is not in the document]
haller33: Understood

api/pkg/gateway/context_test.go
haller33: and a very good one

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,477 @@
...
+			description: "validate empty ID string, for ID behaviour",
+			headers: map[string]string{
+				"X-Role": "",
+			},
+			expected: authorizer.RoleInvalid,
+		}, {
+			description: "validate empty headers, for fail function",
+			headers:     map[string]string{},
+			expected:    authorizer.RoleInvalid,
+		},
```
heiytor, line 37: These two cases are redundant. `authorizer.RoleFromString` checks whether `X-Role` corresponds to a valid role. If it doesn't, it returns `authorizer.RoleInvalid`.

Since `echo` assigns an empty string to a non-existent header, both scenarios will produce the same result. If you prefer to maintain distinct cases, you can handle one case without the header (similar to the second one), and another with an invalid role (such as the string `invalid`).

api/pkg/gateway/context_test.go
```diff
@@ -0,0 +1,477 @@
...
+			description: "validate empty ID string, for ID behaviour",
+			headers: map[string]string{
+				"X-Tenant-ID": "",
+			},
+			expected: nil,
+		}, {
+			description: "validate empty headers, for fail function",
+			headers:     map[string]string{},
+			expected:    nil,
+		},
```
heiytor, line 88: The same applies here. You can safely remove the first case.

Also, you can do the same in the other tests.

api/pkg/gateway/context_test.go
haller33: Understood

heiytor (approved):

otavio (approved):

henrybarreto (approved):

PR #3939, opened 2024-07-05
perf(api): cache auth device response to agent

henrybarreto:


henrybarreto:
In fact, this route already has a caching approach, what makes this PR unnecessary.

PR #3984, opened 2024-07-23
feat(namespace,api,cli): adding default value to announcement on namespace

haller33:
I am adding a default value as a migration to the announcement if no value is set on the announcement.

### DIFF
--- a/api/store/mongo/migrations/migration_74.go
+++ b/api/store/mongo/migrations/migration_74.go
@@ -3,33 +3,12 @@ package migrations
 import (
 	"context"
 
+	"github.com/shellhub-io/shellhub/pkg/models"
 	"github.com/sirupsen/logrus"
 	migrate "github.com/xakep666/mongo-migrate"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 )
-
-const defaultAnnouncementMessage = `
-******************************************************************
-*                                                                *
-*             Welcome to ShellHub Community Edition!             *
-*                                                                *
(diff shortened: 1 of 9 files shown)


api/store/mongo/migrations/migration_74.go
```diff
@@ -0,0 +1,83 @@
...
+		}).Info("Reverting migration")
+
+		filter := bson.M{
+			"settings.connection_announcement": defaultAnnouncementMessage,
+		}
+
+		update := bson.M{
+			"$unset": bson.M{
+				"settings.connection_announcement": "",
+			},
```
heiytor, line 74: Switch the `$unset` to `$set`. The `$unset` expression will remove the `settings.connection_announcement` field instead of setting it to an empty string.

heiytor (changes requested):

api/store/mongo/migrations/migration_74.go
haller33: thanks for the correction @heiytor , I will change now to correct that

haller33: I have added new tests to support the implementation

heiytor (dismissed):

haller33:
adding a new commit, to ensure that the default message will happen on the creation of namespace via cli and api

heiytor:
I think you can squash these two commits into one. Since they depend on each other, combining them would make the history more concise and descriptive, in my opinion.

haller33:
Thanks for the advice @heiytor, I'll maintain this code with two commits since they are related but the scope with each other touches is different and I prefer to keep it this way as separate commits.

PR #3999, opened 2024-07-26
fix(namespace,api,cli): fixing default announcements

haller33:
Fixing default announcements so they can only not be triggered on the enterprise and cloud edition

this PR is a change to the features added on https://github.com/shellhub-io/shellhub/pull/3984

### DIFF
--- a/api/services/namespace.go
+++ b/api/services/namespace.go
@@ -90,9 +90,13 @@ func (s *service) CreateNamespace(ctx context.Context, namespace requests.Namesp
 		},
 		Settings: &models.NamespaceSettings{
 			SessionRecord:          true,
-			ConnectionAnnouncement: models.DefaultAnnouncementMessage,
+			ConnectionAnnouncement: "",
 		},
 		TenantID: namespace.TenantID,
+	}
+
+	if envs.IsCommunity() {
+		ns.Settings.ConnectionAnnouncement = models.DefaultAnnouncementMessage
 	}
 
 	if ok, err := s.validator.Struct(ns); !ok || err != nil {
--- a/api/services/namespace_test.go
+++ b/api/services/namespace_test.go
@@ -493,6 +493,8 @@ func TestCreateNamespace(t *testing.T) {
 				}
 
 				mock.On("UserGetByID", ctx, user.ID, false).Return(user, 0, nil).Once()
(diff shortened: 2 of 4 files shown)


api/services/namespace.go
```diff
@@ -90,7 +90,7 @@ func (s *service) CreateNamespace(ctx context.Context, namespace requests.Namesp
 		},
 		Settings: &models.NamespaceSettings{
 			SessionRecord:          true,
-			ConnectionAnnouncement: models.DefaultAnnouncementMessage,
+			ConnectionAnnouncement: models.DefaultEnterprise(models.DefaultAnnouncementMessage),
 		},
```
heiytor, line 94: Personally, I think that this function adds a lot of indirection. It's probably more readable to use an if statement than a `DefaultEnterprise`.

```go
if !envs.IsCommunity() {
    ns.Settings.ConnectionAnnouncement = models.DefaultAnnouncementMessage
}
```

pkg/models/namespace.go
```diff
@@ -80,3 +84,10 @@ const DefaultAnnouncementMessage = `
 *                                                                *
 ******************************************************************
 `
+
+func DefaultEnterprise(msg string) string {
+	if envs.IsEnterprise() {
+		return ""
+	}
+	return msg
+}
```
heiytor, line 93: This function will create the namespace with the community message even in cloud instances. You must use either `!envs.IsCommunity()` or `envs.IsEnterprise() && envs.IsCloud()`.

heiytor (changes requested):

api/services/namespace.go
haller33: Well, thanks for the advice, I have followed the other advice from an answer from StackOverflow in which they explain that it is indeed if-else is the 'Golang way', but if your code is repetitive (which I can agree, it is not the case), you can use a procedure to replace a ternary operator, but I can change the code to follow an if-else statement, as you suggested!

Down below is the answer from the stack overflow 
 
https://stackoverflow.com/questions/19979178/what-is-the-idiomatic-go-equivalent-of-cs-ternary-operator#answer-19979829

I will change the code as you suggested!

haller33: Since I have a better understanding of how to apply this change to the code, I have implemented the suggestion, as intended, thanks for the good advice!

PR #4021, opened 2024-08-01
feature(agent): add VPN capability to agent

henrybarreto:
It requires the #4022.


henrybarreto:
We could use https://github.com/WireGuard/wireguard-go as it is a more mature and well-tested protocol.

PR #4022, opened 2024-08-02
feature(agent,pkg): remove timed authentication

henrybarreto:


henrybarreto:
Authentication request is critical for the device presence on ShellHub, what makes its remotion difficult to do. 

One of the main limitations is the addition of devices to a namespace. When it is initialized, it authenticates on the namespace, and starts to listen for connections. If the device, after being accepted, is removed, the Agent won't notice it until it loses the connection to the server, showing it as offline in the system.

To remove timed authentication, we should also be able to identify this deletion on the tunnel, or refactor core components for device to server communication. JWT invalidation is also a problem, as the connection could also remain open while the tunnel is established.

Issue #4035, opened 2024-08-06
bug: Fail to connect without password through browser (Cloud)


Ossanes:
### Description

When you try to connect to a device without password (user without password) you get the following message:

```
failed to get auth data from key
Connection ended
```

### Steps to Reproduce

- Open ShellHub Cloud
- Try to connect to a user without password
- Wait for the error message

### Expected Behavior

Is expected that you could connect to the device normally. Through dedicated terminal the error doesn't occur. Also if you set a user password the error doesn't occur too.

### Edition

Cloud

### Version

v0.16.0

### Related Logs

```shell
failed to get auth data from key
Connection ended
```

### Additional Information

I'm not using any Private Key or MFA setting. For these tests was used Brave and Firefox.

haller33:
created PR to fix this issue at #4067

haller33:
this solution will be a **breaking change**, since it allows password empty on login, as per reviewer's

PR #4067, opened 2024-08-12
agent: Fix login issue when password is not provided

haller33:
some fix of the bug from issue [#4035]


haller33:
Add support for empty pass logins on FreeBSD devices and the web interface

henrybarreto (approved):
I have tested with FreeBSD, and worked.

However, I think we should consider it as a breaking change, as, now, empty password are authorized.

haller33:
Add breaking change disclosure as per reviewer's recommendation

PR #4093, opened 2024-08-21
feat: new install script

haller33:
change default install script of the shellhub agent for the project


install.sh
```diff
@@ -5,7 +5,7 @@
 docker_install() {
     echo "🐳 Installing ShellHub using docker method..."
 
-    KEEPALIVE_INTERVAL_ARG="-e SHELLHUB_KEEPALIVE_INTERVAL=$KEEPALIVE_INTERVAL"
+    KEEPALIVE_INTERVAL_ARG="-e SHELLHUB_KEEPALIVE_INTERVAL=${KEEPALIVE_INTERVAL:-30}"
```
otavio, line 8: Please make sure that this works using every POSIX shell. Currently, your installation script wants to `/bin/sh`, so there is no guarantee that it will point to `bash`.

otavio (changes requested):

install.sh
gustavosbarreto: Assigning default values to variables is indeed compatible with POSIX standards.

https://bs2manuals.ts.fujitsu.com/psPOSIXV110en/posix-commands-user-guide-20896/working-with-the-posix-shell-posix-commands-2022-06-9/entering-commands-from-the-posix-shell-posix-commands-2022-06-12/posix-shell-variables-and-parameter-substitution-posix-commands-2022-06-26

PR #4100, opened 2024-08-25
feat(agent,ssh): add webdav server to agent

henrybarreto:


henrybarreto:
Adding a WebDAV on the Agent's isn't necessary, as the SFTP could be used to file access to it.

PR #4137, opened 2024-09-05
fix: Add tests to validate cloud user creation fix

haller33:
tests for validate user creation on cloud


haller33:
Remove tests that are no longer needed

PR #4174, opened 2024-09-16
feat(ssh): add logs on ssh connection dialer

henrybarreto:


otavio (changes requested):
Please take a look in few places where you are using uppercase and change to use lowercase consistently.

henrybarreto:
> Please take a look in few places where you are using uppercase and change to use lowercase consistently.

That is a thing I have noticed too, but it's nice to highlight, mainly on the old logs. Before merging it, I will review and fix these inconsistencies.

PR #4194, opened 2024-10-02
fix: namespace creation not permitted for user

haller33:
create a new limit for user namespace creation, so that now, if MaxNamespaces is a negative value, the user will have no namespace creation limit, if it has a value of zero, the user will not be able to create a new namespace and if it is greater than zero, the behavior will be the same as before

### DIFF
--- a/api/services/errors.go
+++ b/api/services/errors.go
@@ -117,6 +117,7 @@ var (
 	ErrAuthInvalid                  = errors.New("auth invalid", ErrLayer, ErrCodeInvalid)
 	ErrAuthUnathorized              = errors.New("auth unauthorized", ErrLayer, ErrCodeUnauthorized)
 	ErrNamespaceLimitReached        = errors.New("namespace limit reached", ErrLayer, ErrCodeLimit)
+	ErrNamespaceCreationIsForbidden = errors.New("namespace creation not permitted for user", ErrLayer, ErrCodeForbidden)
 	ErrDeviceRemovedCount           = errors.New("device removed count", ErrLayer, ErrCodeNotFound)
 	ErrDeviceRemovedInsert          = errors.New("device removed insert", ErrLayer, ErrCodeStore)
 	ErrDeviceRemovedFull            = errors.New("device removed full", ErrLayer, ErrCodePayment)
@@ -430,6 +431,11 @@ func NewErrBadRequest(err error) error {
 // NewErrNamespaceLimitReached a error to be used when the user namespace limit number is reached.
 func NewErrNamespaceLimitReached(limit int, err error) error {
 	return NewErrLimit(ErrNamespaceLimitReached, limit, err)
+}
+
+// NewErrNamespaceCreationIsForbidden a error, since user have no permition to add a new namespace
+func NewErrNamespaceCreationIsForbidden(limit int, err error) error {
+	return NewErrLimit(ErrNamespaceCreationIsForbidden, limit, err)
 }
 
 func NewErrDeviceRemovedCount(next error) error {
--- a/api/services/namespace.go
+++ b/api/services/namespace.go
@@ -77,9 +77,11 @@ func (s *service) CreateNamespace(ctx context.Context, req *requests.NamespaceCr
 		return nil, NewErrUserNotFound(req.UserID, err)
 	}
 
-	// When MaxNamespaces is less than or equal to zero, it means that the user has no limit
-	// of namespaces.
-	if user.MaxNamespaces > 0 {
+	// When MaxNamespaces is less than zero, it means that the user has no limit
+	// of namespaces. If the value is zero, it means he has no right to create a new namespace
+	if user.MaxNamespaces == 0 {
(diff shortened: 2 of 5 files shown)


api/services/namespace.go
```diff
@@ -77,16 +77,19 @@ func (s *service) CreateNamespace(ctx context.Context, req *requests.NamespaceCr
 		return nil, NewErrUserNotFound(req.UserID, err)
 	}
 
-	// When MaxNamespaces is less than or equal to zero, it means that the user has no limit
-	// of namespaces.
-	if user.MaxNamespaces > 0 {
+	// When MaxNamespaces is less than zero, it means that the user has no limit
+	// of namespaces. If the value is zero, it means he has no right to create a new namespace
+	if user.MaxNamespaces < 0 {
```
heiytor, line 82: You don't need to change these lines.

api/services/namespace.go
```diff
@@ -77,16 +77,19 @@ func (s *service) CreateNamespace(ctx context.Context, req *requests.NamespaceCr
...
 		switch {
 		case err != nil:
 			return nil, err
 		case len(info.OwnedNamespaces) >= user.MaxNamespaces:
 			return nil, NewErrNamespaceLimitReached(user.MaxNamespaces, nil)
 		}
+	} else if user.MaxNamespaces == 0 {
+
+		return nil, NewErrNamespaceLimitReached(user.MaxNamespaces, nil)
 	}
```
heiytor, line 93: Refactor this to use an early return. Move the `if` statement above the standard validation.

```go
if user.MaxNamespaces == 0 {
    return nil, NewErrNamespaceLimitReached(user.MaxNamespaces, nil)
}
```

heiytor (changes requested):

heiytor:
IMHO, you could create methods within the `models.User` struct to encapsulate this validation logic. For example:

```go
func (u *User) HasOwnedNamespacesLimit() int {
    // Returns 0 if the user cannot create namespaces, -1 for unlimited, and 1 for limited.
}
```

heiytor (changes requested):
Add a test case for when the `user.MaxNamespace == 0`. It should be similar to the test case that fails when the user reaches the maximum number of namespaces (*api/services/namespace_test.go#406*).

api/services/namespace.go
```diff
@@ -77,9 +77,11 @@ func (s *service) CreateNamespace(ctx context.Context, req *requests.NamespaceCr
 		return nil, NewErrUserNotFound(req.UserID, err)
 	}
 
-	// When MaxNamespaces is less than or equal to zero, it means that the user has no limit
-	// of namespaces.
-	if user.MaxNamespaces > 0 {
+	// When MaxNamespaces is less than zero, it means that the user has no limit
+	// of namespaces. If the value is zero, it means he has no right to create a new namespace
+	if user.MaxNamespaces == 0 {
+		return nil, NewErrNamespaceZeroReached(user.MaxNamespaces, nil)
```
heiytor, line 83: Why a new error and not the `ErrNamespaceLimitReached`?

haller33: Well, since zero value, is a new type of limit, is not a limit, but a condition, if is zero, is now not able to create (on purpose, not by accident) a new namespace, but with values with has a limit, for instance '3', so this implies that the plan needs to be upgraded and is not, a premeditated intention, with that in mind, I believed that adding a new error message, could resolve significant less trouble shooting with this new logic on the already integer of `MaxNamespaces`, as well as clarify that the reason for any zero value error, is doe to zero value, and not a limit, as is not actually the case

haller33: so this clarifies well the idea that zero value is on propose and not by accident

otavio: I would even use a more specific error mentioning that the user is not allowed to create namespaces. Being zero is just an implementation detail, but the error should be meaningful about the purpose of it.

api/services/errors.go
```diff
@@ -117,6 +117,7 @@ var (
 	ErrAuthInvalid                  = errors.New("auth invalid", ErrLayer, ErrCodeInvalid)
 	ErrAuthUnathorized              = errors.New("auth unauthorized", ErrLayer, ErrCodeUnauthorized)
 	ErrNamespaceLimitReached        = errors.New("namespace limit reached", ErrLayer, ErrCodeLimit)
+	ErrNamespaceZeroReached         = errors.New("namespace creation not permitted for user", ErrLayer, ErrCodeLimit)
```
otavio, line 120:
```suggestion
	ErrNamespaceCreationIsForbidden         = errors.New("namespace creation not permitted for user", ErrLayer, ErrCodeLimit)
```

otavio (changes requested):

api/services/errors.go
heiytor: In addition to this error, I believe that when it occurs, the route can return a `403` status, allowing the frontend to handle this case appropriately.

heiytor (approved):

otavio (approved):

PR #4231, opened 2024-10-14
fix(ssh): add ip address to web session

henrybarreto:


pwall2222:
This is a bug in `v16.0.0`, right?

Since I am self-hosting that version, and I couldn't get web connection ips logged to the `Sessions` panel, and decided to take a look if it was a problem with my reverse proxy or ShellHub itself.
Where I came to find that the environment variable `IP_ADDRESS` was showing up correctly in the agent/ssh terminal, but it wasn't getting correctly logged.
So I went looking in the `v16.0.0` source code to find that the session code was just grabbing the `RemoteAddr`.
https://github.com/shellhub-io/shellhub/blob/0de043ff23d322e9d580c534d07049a522c2527d/ssh/session/session.go#L122

gustavosbarreto:
> This is a bug in `v16.0.0`, right?
> 
> Since I am self-hosting that version, and I couldn't get web connection ips logged to the `Sessions` panel, and decided to take a look if it was a problem with my reverse proxy or ShellHub itself. Where I came to find that the environment variable `IP_ADDRESS` was showing up correctly in the agent/ssh terminal, but it wasn't getting correctly logged. So I went looking in the `v16.0.0` source code to find that the session code was just grabbing the `RemoteAddr`.

@pwall2222  Yes, this is a bug in `v0.16.0`. Version `v0.17.0` is expected to be released in a few days, which should fix this issue.

PR #4237, opened 2024-10-17
feat(cli,api): add field for namespace type

haller33:
Implement new field: APIs and CLI components (ref https://github.com/shellhub-io/cloud/issues/2025)

- Add APIs to support the new field
- Develop CLI components for field interaction
- Follow design outlined in issue https://github.com/shellhub-io/cloud/issues/2025

### DIFF
--- a/api/services/namespace.go
+++ b/api/services/namespace.go
@@ -64,10 +64,17 @@ func (s *service) CreateNamespace(ctx context.Context, req *requests.NamespaceCr
 			ConnectionAnnouncement: "",
 		},
 		TenantID: req.TenantID,
+		Type:     models.NewDefaultType(),
 	}
 
 	if envs.IsCommunity() {
 		ns.Settings.ConnectionAnnouncement = models.DefaultAnnouncementMessage
+	}
+
+	if models.IsTypeTeam(req.Type) {
+		ns.Type = models.TypeTeam
+	} else if models.IsTypePersonal(req.Type) {
+		ns.Type = models.TypePersonal
 	}
 
 	if req.TenantID == "" {
--- a/pkg/models/type.go
+++ b/pkg/models/type.go
@@ -1,0 +1,26 @@
+package models
+
+import env "github.com/shellhub-io/shellhub/pkg/envs"
+
+type Type string
+
+const (
+	TypePersonal Type = "personal"
+	TypeTeam     Type = "team"
+)
+
+func NewDefaultType() Type {
+	if env.IsCloud() {
+		return TypeTeam
+	}
+
+	return TypePersonal
+}
+
+func IsTypeTeam(typeNamespace string) bool {
+	return Type(typeNamespace) == TypeTeam
+}
+
+func IsTypePersonal(typeNamespace string) bool {
+	return Type(typeNamespace) == TypePersonal
+}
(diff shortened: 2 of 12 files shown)


pkg/models/type.go
```diff
@@ -0,0 +1,16 @@
+package models
+
+type Type struct {
+	Personal bool
+	Team     bool
+}
```
henrybarreto, line 6: I think that `Type` shouldn't to be a structure with two fields. It would be a lot simpler being a type to a string or a value, and create possible values from this type.

For example:


```go
type Type string

const TypePersonal Type = "personal"

const TypeTeam Type = "team"
```

haller33: In our Namespace structure, we consistently use pointer indirection for 
foreign types/structures. However, `'Type'` field was a raw string, breaking 
this pattern.

``` go

type Namespace struct {
	Name         string             `json:"name"  validate:"required,hostname_rfc1123,excludes=.,lowercase"`
	Owner        string             `json:"owner"`
	TenantID     string             `json:"tenant_id" bson:"tenant_id,omitempty"`
	Members      []Member           `json:"members" bson:"members"`
	Settings     *NamespaceSettings `json:"settings"`
	Devices      int                `json:"-" bson:"devices,omitempty"`
	Sessions     int                `json:"-" bson:"sessions,omitempty"`
	MaxDevices   int                `json:"max_devices" bson:"max_devices"`
	DevicesCount int                `json:"devices_count" bson:"devices_count,omitempty"`
	CreatedAt    time.Time          `json:"created_at" bson:"created_at"`
	Billing      *Billing           `json:"billing" bson:"billing,omitempty"`
	Type         *Type              `json:"type" bson:"type,omitempty"`
}
```

By wrapping `'Type'` in a struct:
- Maintains consistent pointer indirection across codebase
- Makes architectural pattern explicit and uniform 
- Simplifies future refactoring by following same memory model
- Reduces cognitive load when working with the codebase

should `Type` be a struct of a single string, for keeping this pattern, what you think?

henrybarreto: With the suggestion I gave, the Type wouldn't be a raw string, but a type "over" a string, what means that a variable/parameter from this "Type" would only accept values from this type, how I have constructed in the example.

About the points you make about the structure, this is my view:

> Maintains consistent pointer indirection across codebase

The "Type" only stores a value, so a structure doesn't make sense to exist for this value.

> Makes architectural pattern explicit and uniform

It isn't an architecture pattern enforced, it just appears from the necessities.

> Simplifies future refactoring by following same memory model

It cannot simplify something that doesn't exist, or maybe will never exist yet.

> Reduces cognitive load when working with the codebase

Yeah, a simple "string" could reduce the cognitive load of a structure with one value.


Check the `/pkg/models/member.go` to see an existent example.

pkg/models/type.go
```diff
@@ -0,0 +1,15 @@
+package models
+
+type Type string
+
+const TypePersonal Type = "personal"
```
gustavosbarreto, line 5: Instead of declaring the constants separately, you can group them together to make the code more concise and readable:

```
const (
    TypePersonal Type = "personal"
    TypeTeam     Type = "team"
)
```

api/services/namespace.go
```diff
@@ -64,12 +64,19 @@ func (s *service) CreateNamespace(ctx context.Context, req *requests.NamespaceCr
...
 		},
 		TenantID: req.TenantID,
+		Type:     models.NewDefaultType(),
 	}
 
 	if envs.IsCommunity() {
 		ns.Settings.ConnectionAnnouncement = models.DefaultAnnouncementMessage
 	}
 
+	if models.IsTypeTeam(models.Type(req.Type)) {
```
henrybarreto, line 74: I think that this helper functions could receive `req.Type` directly without the convention.

henrybarreto (approved):

PR #4272, opened 2024-10-31
fix(pkg): ensures that geoip dbPath exists before fetch

heiytor:


gustavosbarreto:
We thought, "WWUBD?" (What Would Uncle Bob Do?) and decided it was time for some tough love on redundant comments.

PR #4324, opened 2024-11-14
feat(ssh): simplify session, improve record, and record any PTY

henrybarreto:
In previous refactoring, we have removed a bunch of operations from the server, letting both the parties communicate more freely. However, the session record kept doing the same blocking operations.

The SSH server acts as an intermediary between the Agent and Client, so reducing communication overhead can significantly improve session performance and overall user experience.

In a previous refactor, we reduced several blocking operations on the server, allowing for more efficient communication between the Agent and Client. However, the session record still relied on blocking operations, which limited performance.

The core idea of this improvement was removing more operations over the communication between Agent and Client communication, and turning session record frame asynchronous using channels.

### DIFF
--- a/ssh/server/channels/session.go
+++ b/ssh/server/channels/session.go
@@ -161,6 +161,12 @@ func DefaultSessionHandler() gliderssh.ChannelHandler {
 
 						return
 					}
+				default:
+					if req.WantReply {
+						if err := req.Reply(false, nil); err != nil {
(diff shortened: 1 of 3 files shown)


henrybarreto:
Beyond that, the record works also over exec operations with PTY.

henrybarreto:
@gustavosbarreto, do you think it is worth?

Issue #4350, opened 2024-11-22
Implement backend support for tag colors 


gustavosbarreto:
This feature will enable storing and retrieving color information associated with tags. Although there is no UI implementation required at this stage, the backend should be ready to handle color attributes for tags.

## Database Changes

* Add a new field (e.g., `color`) to the tags schema in the database to store color information.
* Ensure the field supports common color formats (e.g., HEX).

## API Changes

* Update the API endpoints for creating and editing tags to accept a color attribute.
* Update the API endpoints for retrieving tags to include the color attribute in the response.

### Validation

Validate the color field to ensure it adheres to a valid format (e.g., HEX codes like #FFFFFF).

### Backward Compatibility

Ensure existing tags without color information remain functional and default to empty (null value).

### Testing

Add unit and integration tests to verify the proper handling of the color attribute in all relevant operations.

haller33:
Since peers review, there is new directions for the solution required.: 

feat: tags collection redesign
- Add primary key to `tags` to include tenantID/namespace
- Make tenant the primary granularity level for the `tags` collection (has previously are) 
- Implement cascading operations (rename, delete) across all references including the new `tags` collection 
- Update the reference of the use of `tags` to use new tags collection as main source 
- Return array of objects instead of previous array of strings on tags default route

PR #4362, opened 2024-11-26
feat: implement backend support for tag colors

haller33:
feat(devices): implement backend support for tag colors

- Add color field to device schema and update API endpoints to handle color 
- Attribute with HEX validation. Existing tags default to empty.
- related issue https://github.com/shellhub-io/shellhub/issues/4350


api/store/mongo/device_tags.go
```diff
@@ -35,19 +70,59 @@ func (s *Store) DevicePullTag(ctx context.Context, uid models.UID, tag string) e
...
+
+		res, err := s.TagsBulkDeleteTag(sessCtx, tenant, currentTag)
+		if err != nil {
+			return res, FromMongoError(err)
+		}
+
+		tag.Name = newTag
+
+		_, err2 := s.db.Collection("tags", options.Collection().SetWriteConcern(writeconcern.Majority())).
+			InsertOne(sessCtx, tag)
```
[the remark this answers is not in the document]
haller33: This has to do with a way to simplify behavior in development as an atomic way and the old logic that no longer applies, since your observation, it really is better now to just update, thank you

api/store/mongo/tags.go
```diff
@@ -122,11 +102,91 @@ func (s *Store) TagsDelete(ctx context.Context, tenantID string, tag string) (in
...
+			return false, FromMongoError(err)
+		}
+
+		if tag.Name == "" {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
```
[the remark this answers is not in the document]
haller33: Well, you can keep your opinion, but I must say that it's not 'overly complicated', since actually mixing an error with the intended information of a method is not a good pattern/idea (there is a clear separation of what is data, a boolean value, and what is an actual error). indeed, I also will keep the code on this way, if you wish so, since the behavior is the same, and you are just giving an opinion, but again, I'm not criticizing your opinion, just talking about the best interface and logic that fits `VerifyTagExist`.

api/store/mongo/tags.go
```diff
@@ -122,11 +102,86 @@ func (s *Store) TagsDelete(ctx context.Context, tenantID string, tag string) (in
...
+	tag := new(models.Tags)
+
+	cursor, err := s.db.Collection("tags").Find(ctx,
+		bson.M{"name": tagName, "tenant_id": tenant})
+	if err != nil {
+		return false, FromMongoError(err)
+	}
+
+	for cursor.Next(ctx) {
+		if err != nil {
```
gustavosbarreto, line 129: The `err` variable is not updated within the loop

api/store/mongo/tags.go
```diff
@@ -122,11 +102,86 @@ func (s *Store) TagsDelete(ctx context.Context, tenantID string, tag string) (in
...
 	})
 	if err != nil {
 		return int64(0), FromMongoError(err)
 	}
 
 	return count.(int64), nil
 }
+
+func (s *Store) VerifyTagExist(ctx context.Context, tagName, tenant string) (bool, error) {
+	tag := new(models.Tags)
```
gustavosbarreto, line 120: The Decode method is never called to populate the tag variable with the data retrieved from db.

api/store/mongo/tags.go
```diff
@@ -122,11 +102,86 @@ func (s *Store) TagsDelete(ctx context.Context, tenantID string, tag string) (in
...
+	if err != nil {
+		return false, FromMongoError(err)
+	}
+
+	for cursor.Next(ctx) {
+		if err != nil {
+			return false, FromMongoError(err)
+		}
+
+		if tag.Name == "" {
```
gustavosbarreto, line 133: This check is unnecessary since the presence of a document inherently confirms the tag's existence.

api/store/mongo/tags.go
```diff
@@ -122,11 +102,86 @@ func (s *Store) TagsDelete(ctx context.Context, tenantID string, tag string) (in
...
 		return int64(0), FromMongoError(err)
 	}
 
 	return count.(int64), nil
 }
+
+func (s *Store) VerifyTagExist(ctx context.Context, tagName, tenant string) (bool, error) {
+	tag := new(models.Tags)
+
+	cursor, err := s.db.Collection("tags").Find(ctx,
```
gustavosbarreto, line 122: The cursor is not explicitly closed, which could lead to resource leaks.

api/store/mongo/device_tags.go
```diff
@@ -35,19 +70,59 @@ func (s *Store) DevicePullTag(ctx context.Context, uid models.UID, tag string) e
 }
 
 func (s *Store) DeviceSetTags(ctx context.Context, uid models.UID, tags []string) (int64, int64, error) {
-	tag, err := s.db.Collection("devices").UpdateOne(ctx, bson.M{"uid": uid}, bson.M{"$set": bson.M{"tags": tags}})
+	tag, err := s.db.Collection("devices").
+		UpdateOne(ctx, bson.M{"uid": uid}, bson.M{"$set": bson.M{"tags": tags}})
```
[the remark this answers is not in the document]
haller33: I have checked, this not use of the `MatchedCount` has not been made by myself, this could led for minimal changes in method behavior, they will remain consistent as to the rest of the other methods return value.

PR #4441, opened 2024-12-16
feat(api,pkg): create generic server package for services

henrybarreto:


henrybarreto:
Closed in favor of a simpler approach

PR #4554, opened 2025-02-06
fix(enterprise): pass saml envs to enterprise containers

heiytor:


.env
```diff
@@ -146,3 +146,6 @@ SHELLHUB_PAYWALL=true
 
 # Enable Connector features on UI.
 SHELLHUB_CONNECTOR=true
+
+
```
otavio, line 150:
```suggestion
# ?????????
```

PR #4600, opened 2025-02-19
fix(agent): don't initialize agent when hostname and identity are empty

henrybarreto:
```
A MAC address can be empty when the network interface used to 
communicate with the external world isn't a physical one. In this case,
we should be able to define a custom value for MAC's field using the
[PREFERRED_IDENTITY] variable. If the hostname is also empty,
[PREFERRED_HOSTNAME] could be defined to provide a fallback identifier
for the device. This ensures that even if both the MAC address and
hostname is missing, we have a way to identify the device uniquely.
When it occurs, and no variable was defined, the agent should fail to
initialize.

```

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -343,9 +343,11 @@ func (a *Agent) probeServerInfo() error {
 	return err
 }
 
+var ErrNoIdentityAndHostname = errors.New("the device doesn't have a valid hostname and identity. Set PREFERRED_IDENTITY or PREFERRED_HOSTNAME to specify the device's name and identity")
+
 // authorize send auth request to the server with device information in order to register it in the namespace.
 func (a *Agent) authorize() error {
-	data, err := a.cli.AuthDevice(&models.DeviceAuthRequest{
+	req := &models.DeviceAuthRequest{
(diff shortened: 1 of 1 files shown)


pkg/agent/agent.go
```diff
@@ -343,17 +343,31 @@ func (a *Agent) probeServerInfo() error {
 	return err
 }
 
+var ErrNoIdentityHostname = errors.New("the device doesn't have a valid hostname or identity. Set PREFERRED_IDENTITY or PREFERRED_HOSTNAME to specify the device's name and identity")
```
otavio, line 346: This error is not only related to the hostname, but to the identity as well. So the variable name should refer to that.

henrybarreto: Something like `ErrNoIdentityAndHostname` ?

otavio: I believe so. Also, you should adjust the comment inside of the code to mention that both information are missing.

otavio (approved):

henrybarreto:
@otavio, I've pushed again just to fix the commit message. The code remains the same as approved.

PR #4604, opened 2025-02-21
feat(gateway): optimize DNS resolution and hostname handling

gustavosbarreto:


otavio (approved):

otavio:
The code seems to be not properly aligned. Besides that everything looks fine.

PR #4623, opened 2025-03-05
refactor(ssh): use structures to represent agent and client on session

henrybarreto:


ssh/server/channels/session.go
```diff
@@ -112,32 +112,32 @@ func DefaultSessionHandler() gliderssh.ChannelHandler {
 		logger.Info("session channel started")
 		defer logger.Info("session channel done")
 
-		client, clientReqs, err := newChan.Accept()
+		seat, err := sess.NewSeat()
 		if err != nil {
-			reject(err, "failed to accept the channel opening")
+			reject(err, "failed to create a new set on the SSH session")
```
otavio, line 117: seat

ssh/session/session.go
```diff
@@ -204,6 +257,38 @@ func NewSession(ctx gliderssh.Context, tunnel *httptunnel.Tunnel, cache cache.Ca
...
+	if err != nil {
+		return nil, err
+	}
+
+	c := &ClientChannel{
+		Channel:  channel,
+		Requests: requests,
+	}
+
+	s.Client.Channels[seat] = c
```
otavio, line 271: What happens if I call that twice for the same seat ID?

ssh/session/session.go
```diff
@@ -204,6 +257,38 @@ func NewSession(ctx gliderssh.Context, tunnel *httptunnel.Tunnel, cache cache.Ca
...
+	if err != nil {
+		return nil, err
+	}
+
+	a := &AgentChannel{
+		Channel:  channel,
+		Requests: requests,
+	}
+
+	s.Agent.Channels[seat] = a
```
otavio, line 287: What happens if I call that twice for the same seat ID?

henrybarreto: For these cases, I've added an error to handle it.

PR #4624, opened 2025-03-05
refactor(api,ssh,pkg): use web socket endpoint to save events

henrybarreto:


henrybarreto:
Although the main change of this PR will be done on the current endpoint, this PR has more changes than expected. It changes how a session is got and the endpoint for recording a session and the events. I'm splitting into small pieces to facility the integration in our code base, with the first step being the #4633.

PR #4626, opened 2025-03-08
feat: postgres store

heiytor:


api/services/user.go
```diff
@@ -37,24 +41,33 @@ func (s *service) UpdateUser(ctx context.Context, req *requests.UpdateUser) ([]s
...
+	if req.Username != "" {
+		user.Username = strings.ToLower(req.Username)
+	}
+
+	if req.Email != "" {
+		user.Email = strings.ToLower(req.Email)
+	}
+
+	if req.RecoveryEmail != "" {
+		user.Preferences.SecurityEmail = strings.ToLower(req.RecoveryEmail)
```
otavio, line 57: I think here we can use the previous strategy, which is always to the lowercase function call. The checking seems to be not worth and it complicates the code without a real, probably, good performance gain.

api/services/setup.go
```diff
@@ -88,8 +79,9 @@ func (s *service) Setup(ctx context.Context, req requests.Setup) error {
 		},
 	}
 
+	// TODO: use a transaction here
```
otavio, line 82: I understand that you want to do the user creation and the namespace creation using a database transaction and avoid the need to create and delete the user just by logic. However, I think that this to-do is likely to be left for the future because it's going to demand some more complicated changes in the datastore layer. So my suggestion is for you to describe it better in the description so we can leave it for the next step of the work.

api/services/user.go
```diff
@@ -65,21 +78,20 @@ func (s *service) UpdateUser(ctx context.Context, req *requests.UpdateUser) ([]s
 //
 // Deprecated, use [Service.UpdateUser] instead.
 func (s *service) UpdatePasswordUser(ctx context.Context, id, currentPassword, newPassword string) error {
-	user, _, err := s.store.UserGetByID(ctx, id, false)
+	user, err := s.store.UserGet(ctx, store.UserIdentID, id)
```
otavio, line 81: I believe that this duplicated function could die before we do the change. So we have a restore reduction that needs to be supported for the next version.

tests/environment/configurator.go
```diff
@@ -97,7 +99,7 @@ func (dcc *DockerComposeConfigurator) Up(ctx context.Context) *DockerCompose {
 	}
 
 	tcDc, err := compose.NewDockerComposeWith(
-		compose.WithStackFiles("../docker-compose.yml", "../docker-compose.test.yml"),
+		compose.WithStackFiles("../docker-compose.yml", "../docker-compose.mongo.yml", "../docker-compose.test.yml"),
```
otavio: We must be able to run the tests in both databases.

otavio (changes requested):

PR #4632, opened 2025-03-11
refactor(ui): handle new tags objects

luannmoreira:
This pull request refactor all tags object with new handling capabilities, using
pagination on lists and a new refactored TagFormUpdate on the device and
container associated tags.


ui/src/components/Tags/TagList.vue
```diff
@@ -1,19 +1,22 @@
...
+    :headers
+    :items="tags"
+    :nextPage="next"
+    :previousPage="prev"
+    :loading
+    :totalCount="numberTags"
+    :itemsPerPageOptions="[10, 20, 50, 100]"
+    @changeItemsPerPage="changeItemsPerPage"
+    @clickNextPage="next"
+    @clickPreviousPage="prev"
```
luizhf42: `nextPage`, `previousPage`, `@changeItemsPerPage`, and the two `@click` handlers are no longer necessary since #4890.

ui/src/components/Tags/TagList.vue
```diff
@@ -57,55 +60,89 @@
...
+const itemsPerPage = ref(10);
+const page = ref<number>(1);
+const tags = computed(() => tagsStore.list);
+const tenant = computed(() => localStorage.getItem("tenant"));
+const numberTags = computed<number>(
+  () => tagsStore.getNumberTags,
+);
 
-const canEditTag = hasPermission("tag:edit");
+const getTags = async (perPage: number, page: number): Promise<void> => {
```
luizhf42: I think there's no need to pass `page` and `perPage` arguments since they are variables accessible in the method's scope that are updated by the `v-model`s from `DataTable`.

ui/src/components/Tags/TagList.vue
```diff
@@ -57,55 +60,89 @@
...
+  if (page.value > 1) await getTags(itemsPerPage.value, page.value--);
+};
+
+const changeItemsPerPage = async (newItemsPerPage: number) => {
+  itemsPerPage.value = newItemsPerPage;
+};
+
+watch(itemsPerPage, async (newItemsPerPage) => {
+  await getTags(newItemsPerPage, page.value);
+});
```
luizhf42:
```suggestion
watch([page, itemsPerPage], async () => { await getTags(); });
```

ui/src/components/Tags/TagList.vue
luannmoreira: Indeed, this behavior was changed in the #4890, I've removed that boilerplate part of the code as it is no longer needed.

ui/src/components/Tags/TagList.vue
luannmoreira: You're right, that piece of code doesn't need callbacks as it is available on the `v-model` from the `DataTable` component, I've solved this in my latest push.

heiytor (approved):

luizhf42 (approved):

PR #4645, opened 2025-03-14
feat(ui): add disabled MFA switch for Community accounts

luizhf42:
This PR adds a disabled MFA switch with an explanatory tooltip for Community accounts. Previously, the MFA option on the profile settings page wasn't visible for Community. Now, it appears as a disabled switch with a tooltip stating "Only available for Cloud or Enterprise accounts!" to improve user understanding. The responsiveness of the MFA card was improved too, limiting the width of the text in smaller screens, preventing it from "hiding" the switch button. I've added a basic unit test for this switch and updated the snapshot. 

Screenshot

Additionally, I've refactored the code to replace direct references to `envVariables` in the template with the destructured values `isCloud` and `isCommunity`. I also removed some computed properties for these values as they weren't reactive and therefore served no purpose.


luizhf42:
Closing the PR to separate it into two smaller ones, one for the feature (UI improvement) and the other for the refactorings.

PR #4678, opened 2025-03-20
fix: Installation don't stop, error but the service is failing.

slackjeff:
Service normally starts, but after a few seconds it fails. The installer did not cause an error


henrybarreto:
Hi, @slackjeff 

Thank you for the contribution! It makes totally sense to check if the service is up before continue it.

I did notice that the commit message doesn’t quite follow our usual pattern. We like to stick to the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) approach.

Could you please update it when you get a chance? Thanks again.

gustavosbarreto:
Thanks for the PR! This has been addressed in a separate commit, so I’m closing this one. Appreciate the effort!

PR #4735, opened 2025-04-14
fix(ssh,pkg): synchronize events registering and SSH requests

henrybarreto:
Depends on: https://github.com/shellhub-io/shellhub/pull/4753


henrybarreto:
Depends on: https://github.com/shellhub-io/shellhub/pull/4753

PR #4747, opened 2025-04-18
My changes

Sh4der:
Allow public key only connections. Disable password connection

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -39,13 +39,13 @@ func main() {
 				log.WithError(err).WithFields(fields).Fatal("Failed to load de configuration from the environmental variables")
 			}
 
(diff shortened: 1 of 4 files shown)


otavio:
Why did you close the pull request? I think that we would need to improve it a little bit, but you are on the right direction, I think.

Sh4der:
This was just a small change for use on my private devices. As I only have a short time until I finish my master's thesis, I was unable to implement this properly. It's functional as far as I can tell, but the changes still have to be implemented in the UI. Currently, only the password requests are rejected. However, the UI still asks for the password and user name. As soon as I have more time again, I will get back to the subject. The pull request was just a mistake, I actually wanted to wait until it was completely finished. But I'm glad that there is interest :)

PR #4765, opened 2025-04-28
fix(cloud): set user's username to nil when creating with SAML origin

heiytor:


otavio:
Please do a rebase of this change.

PR #4768, opened 2025-04-29
fix(ui): fix player error when playing session recorded on local terminal

luizhf42:
This PR fixes the `getSessionDimensions` method in the player, which previously relied on an initial resize event—behavior specific to web terminal sessions. For sessions recorded via local terminals (e.g., using [ssh](https://docs.shellhub.io/user-guides/devices/connecting#command-line-ssh-client)), the absence of this event caused WASM errors and broke player initialization.

The method now gracefully handles missing resize events by falling back to default dimensions (80x24), ensuring proper playback in these scenarios.


luizhf42:
Closing to merge the two player-related branches and make a solo PR.

PR #4776, opened 2025-05-05
fix(api): avoid nil identity when authenticating a device

henrybarreto:


otavio:
Please check the CI report.

otavio:
Did you try to use this fix with old versions of the agent? My only concern is if it breaks any old users.

henrybarreto:
> Did you try to use this fix with old versions of the agent? My only concern is if it breaks any old users.

Yes, I did. 

I've confirmed that. Versions like `v0.3.3`, `v0.3.9` and `v0.4.0` works normally with this change without a custom identity.

PR #4780, opened 2025-05-06
chore(docker): update golang base image to 1.23

henrybarreto:


agent/Dockerfile.arm32v6
```diff
@@ -1,6 +1,6 @@
 # docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
 
-FROM arm32v6/golang:1.22.6-alpine3.19
+FROM arm32v6/golang:1.23-alpine3.19
```
otavio, line 3: AS ...

henrybarreto: I understand that `AS` here could enforce the pattern that we have in others Dockerfile. However, I think we could do it in another pull request, as the goal of this one is only updating the version.

What do you think?

PR #4781, opened 2025-05-06
refactor(ui): refactor new Snackbar component

luizhf42:
This PR refactors the Snackbar system in the Community edition to, in the future, replace the previous setup, built around four components and a dedicated store, with a simplified, plugin-based approach using a single component.

This sets the foundation for migrating existing components and views to the new system. That migration, along with test updates, will follow in subsequent PRs after this one is approved.

### DIFF
--- a/ui/src/plugins/snackbar.ts
+++ b/ui/src/plugins/snackbar.ts
@@ -1,35 +1,56 @@
-import { App, Plugin } from "vue";
+import { App, Plugin, reactive } from "vue";
 
-const InjectionKey = "snackbar";
+const InjectionKey = Symbol("snackbar");
 
-interface PluginInterface {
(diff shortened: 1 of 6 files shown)


ui/src/plugins/snackbar.ts
```diff
@@ -1,35 +1,56 @@
...
+  showWarning: (msg: string) => void;
+  getMessage: () => string;
+  getType: () => SnackbarType;
+  getShow: () => boolean;
 }
 
+const state = reactive<SnackbarState>({
+  message: "",
+  type: "info",
+  show: false,
```
luizhf42, line 26: Maintained the `show` name because other components of the codebase also use this name for conditional rendering.

PR #4783, opened 2025-05-07
feature(install): check logs before confirming agent standalone installation

henrybarreto:


install.sh
```diff
@@ -136,6 +136,27 @@ standalone_install() {
...
+    trap 'echo "❗ Interrupted. Disabling shellhub-agent..."; $SUDO systemctl disable --now shellhub-agent; exit 1' INT
+
+    echo "🔍 Checking service status..."
+    echo "Please wait for the service to start. This may take a few seconds."
+    echo "Press Ctrl+C to cancel the installation."
+    echo ""
+    echo "If it takes too long, please check the logs with the command:"
+    echo "journalctl -f -u shellhub-agent"
+
+    $SUDO journalctl -f -u shellhub-agent --since "$(systemctl show -p ActiveEnterTimestamp shellhub-agent | cut -d= -f2)" | while read -r line; do
```
otavio, line 148: We must have some kind of timeout, so if it takes more than 15 seconds or so, we may fail directly.

PR #4788, opened 2025-05-10
feat: podman support

No9:
This PR adds podman support to install.sh
It looks for podman before docker as aliasing docker with podman with `alias docker=podman` is a common scenario and attempting to use the docker params would cause issues.

It also installs the container as `root` but further PRs are being investigated to look at rootless options

**Test Environment**
```
$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="40 (Xfce)"

$ podman --version 
podman version 5.4.2
``` 
**Tested as root and user with the following output**
```
TENANT_ID=<TENENT-ID> SERVER_ADDRESS=https://cloud.shellhub.io sh ./install.sh 
🛠️ Welcome to the ShellHub Agent Installer Script

📝 Summary of chosen options:
- Server address: https://cloud.shellhub.io
- Tenant ID: <TENENT-ID>
- Install method: podman
- Agent version: v0.18.2

🐳 Installing ShellHub using podman method...
📥 Downloading ShellHub container image...
70cb3f21e378e59bd9650fdac1de5c856b063dceda300c79e69826d2128b0598
🚀 Starting ShellHub container in Agent mode...
5fcd62c0dc48f847421f86b33c272cef81f462240e7f907aa6064ca1917ddf24
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,70 @@
 #!/bin/sh
 
 # Overridden variables from Go template: {{.Overrides}}
+podman_install() {
+    [ -n "${KEEPALIVE_INTERVAL}" ] && ARGS="$ARGS -e SHELLHUB_KEEPALIVE_INTERVAL=$KEEPALIVE_INTERVAL"
+    [ -n "${PREFERRED_HOSTNAME}" ] && ARGS="$ARGS -e SHELLHUB_PREFERRED_HOSTNAME=$PREFERRED_HOSTNAME"
+    [ -n "${PREFERRED_IDENTITY}" ] && ARGS="$ARGS -e SHELLHUB_PREFERRED_IDENTITY=$PREFERRED_IDENTITY"
+
+    echo "📥 Downloading ShellHub container image..."
+
+    {
+        $SUDO podman pull -q shellhubio/agent:$AGENT_VERSION
+    } || { echo "❌ Failed to download shellhub container image."; exit 1; }
+
+    MODE=""
+    DEFAULT_CONTAINER_NAME="shellhub"
+
+    case "$1" in
+        "")
+            ;;
+        "agent")
+            shift 1
+            ;;
+        "connector")
+            MODE="connector"
+            DEFAULT_CONTAINER_NAME="shellhub-connector"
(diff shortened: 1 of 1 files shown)


gustavosbarreto (approved):

install.sh
```diff
@@ -1,6 +1,70 @@
...
 # Overridden variables from Go template: {{.Overrides}}
+podman_install() {
+    [ -n "${KEEPALIVE_INTERVAL}" ] && ARGS="$ARGS -e SHELLHUB_KEEPALIVE_INTERVAL=$KEEPALIVE_INTERVAL"
+    [ -n "${PREFERRED_HOSTNAME}" ] && ARGS="$ARGS -e SHELLHUB_PREFERRED_HOSTNAME=$PREFERRED_HOSTNAME"
+    [ -n "${PREFERRED_IDENTITY}" ] && ARGS="$ARGS -e SHELLHUB_PREFERRED_IDENTITY=$PREFERRED_IDENTITY"
+
+    echo "📥 Downloading ShellHub container image..."
+
+    {
+        $SUDO podman pull -q shellhubio/agent:$AGENT_VERSION
```
henrybarreto, line 12: Hi, @No9 

That is amazing, but I found a small problem in this line.

If the user doesn't have the `docker.io` defined in the registries, this command will fail like this:

```
Error: short-name "shellhubio/agent:v0.18.3" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"
```

Could you, please, modify it to something like this?

```suggestion
        $SUDO podman pull -q docker.io/shellhubio/agent:$AGENT_VERSION
```

henrybarreto (changes requested):

henrybarreto:
@gustavosbarreto, should we check if the service is running to avoid an error like this?

```
Error: statfs /var/run/podman/podman.sock: no such file or directory
```

I thought in this command to check it. Does it make sense?

```
sudo systemctl is-active podman.socket
```

install.sh
No9: Updated in follow up commit

henrybarreto (approved):

PR #4808, opened 2025-05-14
fix(gateway): unsure certification renew when startup the gateway

henrybarreto:


gateway/certbot.go
```diff
@@ -84,14 +87,29 @@ func (cb *CertBot) generateCertificate() {
 	cb.stopACMEServer(acmeServer)
 }
 
-func (cb *CertBot) generateCertificateFromDNS(provider DNSProvider) {
+func (cb *CertBot) generateIniFile(provider DNSProvider) (*os.File, error) {
 	fmt.Println("Generating SSL certificate with DNS")
```
gustavosbarreto, line 91: Duplicated message

gateway/certbot.go
```diff
@@ -84,14 +87,29 @@ func (cb *CertBot) generateCertificate() {
 	cb.stopACMEServer(acmeServer)
 }
 
-func (cb *CertBot) generateCertificateFromDNS(provider DNSProvider) {
+func (cb *CertBot) generateIniFile(provider DNSProvider) (*os.File, error) {
```
gustavosbarreto, line 90: I think `generateProviderCredentialsFile` would be a more descriptive name for this function. It better reflects the purpose of creating a credentials file for the DNS provider, rather than just indicating that it generates a generic .ini file.

gateway/certbot.go
henrybarreto: Message removed

PR #4811, opened 2025-05-16
refactor(ui): refactor Notifications code

luizhf42:
This PR refactors the notification-related code, including its components, store, types, and tests. 

- The component was separated into two, preventing a single component from having multiple responsibilities and improving maintainability. 
- Unnecessary code was removed and the logic and conditionals were improved.
- The store was reorganized, making it less coupled to the pending devices/containers implementation, and facilitating a future extension and addition of more notification types. It's not a big change, but it already prepares the path for a future, more complex rework.
 
**Work in progress**


ui/src/components/Devices/DeviceActionButton.vue
```diff
@@ -112,16 +111,13 @@ const close = () => {
...
   try {
     emit("update");
-    if (
-      window.location.pathname === "/devices/pending"
-          || window.location.pathname === "/devices"
-    ) {
-      store.dispatch("devices/refresh");
-      store.dispatch("notifications/fetch");
-    }
+    if (window.location.pathname.includes("devices")) await store.dispatch("devices/refresh");
```
otavio, line 117: I would match for the prefix, so there is no risk of having any device' substrings matched by mistake.

ui/src/interfaces/INotification.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export enum NotificationType {
+  DEVICE = "device",
+  CONTAINER = "container"
+}
+
+type NotificationData = IDevice | IContainer;
+
+export interface INotification {
+  id: string | number;
```
otavio, line 12: Why don't we restrict it to a string only? Is there any valid use case for a number here?

ui/src/components/Devices/DeviceActionButton.vue
luizhf42: Makes sense. I replaced the `includes` with `startsWith("/devices")`, making it less error-prone. Additionally, now there is also a verification for the `/containers` path, instead of refreshing the containers list every time as a fallback.

otavio (approved):

PR #4821, opened 2025-05-19
refactor(ui): add new snackbar code to admin

luizhf42:
This PR refactors the snackbar in `/admin` to the new plugin-based approach, directly importing it from the `src` folder, updating the components' code, and removing the old Snackbar components, store, and interfaces.


ui/admin/src/components/Snackbar/SnackbarNew.vue
```diff
@@ -0,0 +1,22 @@
+<template>
```
otavio, line 1: Why are you using the new suffix in the component name? I think that we should not use that because it is kind of confusing.

ui/admin/src/plugins/snackbar.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+interface ISnackbarPlugin {
+  showSuccess: (msg: string) => void;
+  showError: (msg: string) => void;
+  showInfo: (msg: string) => void;
+  showWarning: (msg: string) => void;
+  getMessage: () => string;
+  getType: () => SnackbarType;
+  getShow: () => boolean;
```
otavio, line 20: This seems to have a lot of duplication from the regular user interface. Why can we not share the components between the administration and the regular user interface?

ui/admin/src/components/Snackbar/SnackbarNew.vue
luizhf42: I'm just using it for now to keep the old Snackbar component until the refactoring is done, preventing the snackbar feature from breaking between the commits. After that, the old components will be removed, and the new one will be renamed to `Snackbar.vue`.

PR #4842, opened 2025-05-22
fix(api): avoid active field update when updating session fields

henrybarreto:
The store function that retrieves a session sets the `Active` field based on the existence of a document in another collection. When that document exists, `Active` is set to `true`. However since we reuse the same [models.Session] struct and modify it in the update method, it is persisted in the database wrongly. Something similar happens with the field `Device`, but it is ignored by the BSON, it isn't persistented on the database.


henrybarreto:
Closed in favor of https://github.com/shellhub-io/shellhub/pull/4843

PR #4858, opened 2025-05-28
feat(api): add flexible device resolution with query parameter

### LINKED ISSUE #4857: feat: Device Resolution by Attribute
### Discussed in https://github.com/shellhub-io/shellhub/discussions/4848

<div type='discussions-op-text'>

<sup>Originally posted by **No9** May 25, 2025</sup>
Currently the device details page is made available through the internal guid in the navigation path.
e.g. `https://cloud.shellhub.io/devices/d1ae8d18090f7cb30e1bb827ec6775bd586c0394faacc98a7bc4c2990addce4d`
This is a bit problematic for 3rd party integrations where an external systems would like to provide a link to the device remote connection capability in shellhub. 
The problem is mainly due to some form of sync integration needing to be developed.
These can be brittle and lead to complexity due to features like compensating transactions being required during the device registration. This problem compounds as more backend management systems are added to the device.

However shellhub has a great feature in the registration process that allows for over riding of the hostname using `SHELLHUB_PREFERRED_HOSTNAME`. 
This allows for a shared id to be established across different management systems greatly simplifying this problem. 

To bring this full circle it would be great for the UI to support `https://cloud.shellhub.io/devices?hostname=SHELLHUB_PREFERRED_HOSTNAME` as a navigation feature directly to the device details page. 

Let me know any thoughts on this feature. 
</div>

# Proposal: Device Resolution by Attribute

## Problem

Today, the device details page in ShellHub is only accessible using the device’s internal UID:

```
https://cloud.shellhub.io/devices/d1ae8d18090f7cb30e1bb827ec6775bd586c0394faacc98a7bc4c2990addce4d
```

This creates problems for 3rd party systems that need to link to ShellHub. These systems usually know the device by its hostname, not UID. This forces developers to write extra logic to keep data in sync, which makes integration harder.

## Goal

Allow users and systems to access the device details page in ShellHub using the `hostname` instead of the `UID`, with no extra sync logic.

## Proposal Overview

This proposal introduces a generic approach to resolving devices by any supported attribute, not just the hostname. The first version will support hostname, but the structure will allow easy addition of other attributes in the future

This change will need updates to both the **API** (@shellhub-io/backend-owners) and the **UI** (@shellhub-io/frontend-owners).

### API Endpoint: `/api/devices/resolve`

**Usage:**

```
GET /api/devices/resolve?hostname=my-device-name
```

**Parameters:**

| name     | description            |
| -------- | ---------------------- |
| hostname | Hostname of the device |

**Response:**

```json
{
  "uid": "d1ae8d18090f7cb30e1bb827ec6775bd...",
  "hostname": "my-device-name",
  ....
}
```

> NOTE: In the future, other identifiers like `mac` or `serial_number` can be supported with similar query parameters.

If the device is not found, the API returns a 404 error.

## UI Route: `/devices/:identifier?resolver=hostname`

**Examples:**

* **By UID (default):** `/devices/d1ae8d18090f7cb30e1bb827ec6775bd...`
* **By Hostname:** `/devices/my-device-name?resolver=hostname`

**Behavior:**

If `resolver=hostname` is in the URL:

* The UI calls `/api/devices/resolve?hostname=:identifier`
* Then it redirects or loads `/devices/:uid` internally to keep the UI behavior the same

If there is no resolver, the UI assumes `:identifier` is a UID and loads the page directly according to the current behavior.

## Benefits

* Makes the UI route simple and consistent (`/devices/:identifier`), regardless of the identifier type
* The `resolver` parameter keeps it clear and explicit what kind of identifier is being used
* External systems can generate device links using only the hostname, without needing to fetch the UID first
* Reduces integration complexity and removes the need for custom sync logic
* The API becomes more flexible by allowing queries using different types of identifiers (hostname, mac, etc.)
* Easy to extend the solution in the future, with minimal changes to the API and UI

heiytor:
closes #4857 

Implement device resolver functionality that allows fetching devices using attributes beyond the default UID. The new `resolver` query parameter accepts different attribute types while automatically scoping results to the current tenant context for security.

Currently supported resolvers:
- uid (default behavior)
- hostname


api/services/device.go
```diff
@@ -73,10 +75,22 @@ func (s *service) ListDevices(ctx context.Context, req *requests.DeviceList) ([]
...
-	device, err := s.store.DeviceGet(ctx, uid)
+func (s *service) GetDevice(ctx context.Context, req *requests.DeviceGet) (*models.Device, error) {
+	ns, err := s.store.NamespaceGet(ctx, req.TenantID)
+	if err != nil {
+		return nil, NewErrNamespaceNotFound(req.TenantID, err)
+	}
+
+	var device *models.Device
+	switch req.Resolver {
+	case "hostname":
```
otavio, line 86: Please use a const as an enumeration so we avoid using strings here.

PR #4867, opened 2025-05-30
fix(ui): change namespace add panel using env flag

luannmoreira:
Use computed `openVersion` to reflect `isCommunity` from env vars.
This enables conditional UI behavior based on environment config.

### DIFF
--- a/ui/src/components/Namespace/NamespaceAdd.vue
+++ b/ui/src/components/Namespace/NamespaceAdd.vue
@@ -68,18 +68,19 @@
 </template>
 
 <script setup lang="ts">
-import { computed, ref } from "vue";
+import { computed } from "vue";
 import { useField } from "vee-validate";
 import * as yup from "yup";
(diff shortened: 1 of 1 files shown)


ui/src/components/Namespace/NamespaceAdd.vue
```diff
@@ -68,18 +68,19 @@
...
 import { useStore } from "@/store";
 import handleError from "@/utils/handleError";
 import useSnackbar from "@/helpers/snackbar";
+import { envVariables } from "@/envVariables";
 
 const store = useStore();
 const snackbar = useSnackbar();
 const model = defineModel({ default: false });
-const openVersion = ref(false);
+const openVersion = computed(() => envVariables.isCommunity);
```
otavio, line 83: I suggest changing the name of the variable in another commit to better identify whether we are concerned about the community version or not.

PR #4885, opened 2025-06-03
fix(ui): fix player playback time for sessions longer than 24h

luizhf42:
This PR fixes the player's playback time formatting, making it behave properly in sessions longer than 1 day. The formatting logic was rewritten and moved to a separate util file. The tests were also updated to test the new util behavior.

### DIFF
--- a/ui/src/utils/playerPlayback.ts
+++ b/ui/src/utils/playerPlayback.ts
@@ -1,0 +1,15 @@
+const formatPlaybackTime = (timeInSeconds: number): string => {
+  if (!Number.isFinite(timeInSeconds) || timeInSeconds < 0) return "00:00";
+
+  const hours = Math.floor(timeInSeconds / 3600);
+  const minutes = Math.floor((timeInSeconds % 3600) / 60);
+  const seconds = Math.floor(timeInSeconds % 60);
+
+  const pad = (n: number) => n.toString().padStart(2, "0");
+
+  return hours > 0
+    ? `${hours}:${pad(minutes)}:${pad(seconds)}`
+    : `${pad(minutes)}:${pad(seconds)}`;
+};
+
+export default formatPlaybackTime;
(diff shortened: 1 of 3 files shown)


ui/tests/components/Sessions/Player.spec.ts
```diff
@@ -124,10 +125,13 @@ describe("Asciinema Player", () => {
   });
 
   it("Formats time correctly", () => {
-    expect(wrapper.vm.formatTime(3661)).toBe("01:01:01"); // hh:mm:ss if session is longer than 1 hour
-    expect(wrapper.vm.formatTime(61)).toBe("01:01"); // mm:ss otherwise
-    expect(wrapper.vm.formatTime(59)).toBe("00:59"); // Less than 1 minute
-    expect(wrapper.vm.formatTime(0)).toBe("00:00"); // Zero time
+    expect(formatPlaybackTime(3661)).toBe("1:01:01"); // hh:mm:ss if session is longer than 1 hour
+    expect(formatPlaybackTime(61)).toBe("1:01"); // mm:ss otherwise
+    expect(formatPlaybackTime(59)).toBe("0:59"); // Less than 1 minute
```
gustavosbarreto: could we ensure that minutes are also zero-padded when hours are not shown?

ui/src/utils/playerPlayback.ts
```diff
@@ -0,0 +1,15 @@
+const formatPlaybackTime = (timeInSeconds: number): string => {
+  if (!Number.isFinite(timeInSeconds) || timeInSeconds < 0) return "00:00";
+
+  const hours = Math.floor(timeInSeconds / 3600);
```
gustavosbarreto, line 4: consider using `Math.floor` instead of `Math.round` to make sure we're only showing time that's actually elapsed. This avoids cases where something like 01:59.9 shows up as 02:00, which could feel misleading in a playback context.

ui/tests/components/Sessions/Player.spec.ts
luizhf42: Sure. I'll change this.

PR #4911, opened 2025-06-11
feat(api): add session's events listing

henrybarreto:


api/routes/routes.go
```diff
@@ -91,7 +91,7 @@ func NewRouter(service services.Service, opts ...Option) *echo.Echo {
 	internalAPI.GET(GetPublicKeyURL, gateway.Handler(handler.GetPublicKey))
 	internalAPI.POST(CreatePrivateKeyURL, gateway.Handler(handler.CreatePrivateKey))
 	internalAPI.POST(EvaluateKeyURL, gateway.Handler(handler.EvaluateKey))
-	internalAPI.GET(EventsSessionsURL, gateway.Handler(handler.EventSession))
+	internalAPI.GET(SaveEventsSessionsURL, gateway.Handler(handler.SaveEventsSession))
```
otavio: Are you sure this should be a GET method instead of a POST?

henrybarreto: Yeah. I'm sure.

This endpoint starts a WebSocket connection to send events from an SSH session, and the protocol requires GET, if I'm not wrong.

henrybarreto:
The change this PR propose must be adapted to the PG implementation when merged.

PR #4921, opened 2025-06-17
feat(api): introduce device cache counter

heiytor:


api/store/mongo/device.go
```diff
@@ -282,7 +282,7 @@ func (s *Store) DeviceDelete(ctx context.Context, uid models.UID) error {
 	return err
 }
 
-func (s *Store) DeviceCreate(ctx context.Context, d models.Device, hostname string) error {
+func (s *Store) DeviceCreate(ctx context.Context, d models.Device, hostname string) (bool, error) {
```
otavio, line 285: I'd like to know if this API signature is the best one. I understand that you are trying to figure out whether it has been inserted or not. Boolean is not a straightforward API because it does not communicate other possible values for the feature. What about using constants that are clear when comparing whether it was or was not inserted?

otavio (changes requested):

api/store/mongo/device.go
gustavosbarreto: I get your point about the boolean not being super descriptive, but this pattern is pretty common in Go. Also, for consistency, the rest of this API already follows this same pattern. Changing just this one function to use a different style would make the API inconsistent.

otavio (approved):

PR #4933, opened 2025-06-20
fix(api): remove namespace from cache when device count is increased

henrybarreto:


api/store/mongo/namespace.go
```diff
@@ -414,5 +414,9 @@ func (s *Store) NamespaceIncrementDeviceCount(ctx context.Context, tenantID stri
 		return store.ErrNoDocuments
 	}
 
+	if err := s.cache.Delete(ctx, strings.Join([]string{"namespace", tenantID}, "/")); err != nil {
+		log.Error(err)
```
otavio, line 418: It would be important to have a test case to cover this error, so that this does not happen in the future.

PR #4957, opened 2025-06-30
fix(api): count the right collection for sessions events

henrybarreto:


api/store/mongo/session.go
```diff
@@ -368,7 +368,7 @@ func (s *Store) SessionListEvents(ctx context.Context, uid models.UID, seat int,
 
 	queryCount := query
 	queryCount = append(queryCount, bson.M{"$count": "count"})
-	count, err := AggregateCount(ctx, s.db.Collection("firewall_rules"), queryCount)
+	count, err := AggregateCount(ctx, s.db.Collection("sessions_events"), queryCount)
```
otavio, line 371: We don't have a test opening this feature, so that is likely the reason why this hasn't been caught before.

henrybarreto: We're refactoring this feature, and the tests for it has been added there.

https://github.com/shellhub-io/shellhub/pull/4911/files#diff-a4a75e243be9c8a860693981e03e565330ec0057edd6f2dbba17fda9cfd00f50R474

henrybarreto: This pull request is just to avoid a wrong output until #4911 be merged.

PR #4978, opened 2025-07-09
feat(ssh): improve ssh error messages

henrybarreto:


ssh/server/server.go
```diff
@@ -31,6 +31,46 @@ type Server struct {
...
 
+const (
+	InvalidSSHIDMessage = `SSHID Format Error
+==================
+
+The SSHID format is incorrect. You need to include your username.
+
+Correct format: username@namespace.device@host
+
+Examples:
```
otavio: My suggestion is to use the server address when providing examples, so the user can have a real example of how it would look like.

henrybarreto: I opted not to use the real server to avoid encouraging users to rely on it, which could increase the load on the example. Additionally, there's a risk that someone could create a namespace and device similar to the example to mislead users, unless we create that first.

PR #4991, opened 2025-07-14
fix(ui): validation for X.509 certificate in SAML config

luannmoreira:
# Description

This PR fixes and enhances the SAML SSO configuration flow by introducing validation for the X.509 certificate field. It ensures that users provide a valid certificate in PEM format before submitting the configuration form.

## Changes

- Added validateX509Certificate utility to verify X.509 certificates using sshpk.
- Display inline error messages for invalid or empty certificate fields.
- Improved the hasError computed logic to account for certificate validity.
- Refactored how SAML mapping data is constructed and validated.
- Ensured only non-empty mappings are submitted.

## Benefits

- Prevents invalid X.509 certificates from being submitted.
- Improves user feedback with clear validation errors.
- Reduces chances of SAML misconfiguration due to bad input.

## Testing

- Attempted to submit form with empty and malformed certificates.
- Verified correct error messages display.
- Confirmed form submission only succeeds with a valid certificate.

### DIFF
--- a/ui/admin/src/components/Instance/SSO/ConfigureSSO.vue
+++ b/ui/admin/src/components/Instance/SSO/ConfigureSSO.vue
@@ -12,6 +12,7 @@
           <div v-if="checkbox" data-test="idp-metadata-section">
             <v-text-field
               v-model="IdPMetadataURL"
+              :error-messages="IdPMetadataURLError"
               label="IDP Metadata URL"
               variant="underlined"
               required
@@ -21,6 +22,7 @@
           <div v-else data-test="idp-manual-section">
             <v-text-field
               v-model="acsUrl"
+              :error-messages="acsUrlError"
               label="IdP SignOn URL"
               variant="underlined"
               required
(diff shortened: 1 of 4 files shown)


ui/admin/src/components/Instance/SSO/ConfigureSSO.vue
```diff
@@ -201,59 +205,86 @@ const resetFields = () => {
...
+  } else {
+    x509CertificateErrorMessage.value = ""; // clear error
   }
-  return (
-    acsUrl.value.trim() === ""
-    || entityID.value.trim() === ""
-    || x509Certificate.value.trim() === ""
-    || mappings.value.some((mapping) => !mapping.key || mapping.value.trim() === "")
-  );
 });
```
luizhf42: Maybe we could extract this arrow function to a method and call it with the [update model event](https://vuetifyjs.com/en/api/v-textarea/#events-update:modelValue) to listen to changes in `x509Certificate` instead of setting up a watcher? Just a personal opinion.

luannmoreira: Great observation @luizhf42, in fact, using the v-on directive `@update:model-value` is a better practice then using a watcher to observe the validation, I've sent a new push to this PR changing this feature.

otavio (approved):

luizhf42 (approved):

otavio (approved):

PR #5201, opened 2025-09-01
 feat(agent): use multistream and yamux muxer instead of HTTP server

henrybarreto:


tests/environment/utils.go
```diff
@@ -4,13 +4,13 @@ import (
 	"bytes"
 	"io"
 	"net"
+	"slices"
```
otavio: Please verify the commit message.

PR #5207, opened 2025-09-04
refactor(ui): refactor utils

luizhf42:
This pull request refactors the application's utils. The refactoring involves simplifying logic, removing unused utils, and reusing the main UI's utils in the admin, preventing code duplication.


ui/admin/src/components/Instance/SSO/ConfigureSSO.vue
```diff
@@ -248,7 +248,9 @@ const close = () => {
 
 const isCertificateValid = computed(() => {
   if (!x509Certificate.value.trim()) return false;
-  return validateX509Certificate(x509Certificate.value);
+  try {
+    return isX509CertificateValid(x509Certificate.value);
```
otavio: Why do you need to have this in a try-catch block?

ui/src/components/WebEndpoints/WebEndpointDelete.vue
```diff
@@ -61,10 +61,7 @@ const update = () => {
   showDialog.value = false;
 };
 
-const hasAuthorizationDeleteWebEndpoint = () => {
-  const { role } = authStore;
-  return !!role && hasPermission(authorizer.role[role], actions.webendpoint.delete);
-};
+const hasAuthorizationDeleteWebEndpoint = () => hasPermission(authorizer.role[authStore.role], actions.webendpoint.delete);
```
otavio: Is it better for you to centralize all the checking permissions functions in a single module, where you could avoid duplication and also centralize the code so all the components use the same functions?

ui/admin/src/components/Instance/SSO/ConfigureSSO.vue
luizhf42: Because sshpk's `parseCertificate` throws a `CertificateParseError` when the certificate is invalid and cannot be parsed. However, I'll change this to move that try-catch block to the util file instead of having it inside of the component's logic.

ui/src/components/WebEndpoints/WebEndpointDelete.vue
luizhf42: Yes, simplifying that permission/authorizer util would help and reduce a lot of code repetition. I will work on this now.

otavio (approved):

PR #5222, opened 2025-09-09
fix(agent,pkg): enhance exec ssh command for proper environment variables

henrybarreto:
We are addressing an important issue related to the execution of SSH commands within the agent package. The primary goal of this modification was to ensure that environment variables are loaded correctly when executing commands.

To achieve this, we replaced the previous implementation that utilized `setpriv` with a more reliable approach using `su`. This change allows us to create internal commands more effectively while maintaining the necessary user context.

### DIFF
--- a/pkg/agent/server/modes/host/command/command_native.go
+++ b/pkg/agent/server/modes/host/command/command_native.go
@@ -28,6 +28,28 @@ func NewCmd(u *osauth.User, shell, term, host string, envs []string, command ...
 		groups = append(groups, u.GID)
 	}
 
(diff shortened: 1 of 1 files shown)


znaniye:
I'm running the native command, but the environment variables don't look correct. For example, running `ssh user-example@cloud.shellhub.io /run/current-system/sw/bin/env` returns:
```
SHELL=/bin/sh
PWD=/home/user-example
HOME=/home/user-example
TERM=
SHLVL=0
SHELLHUB_HOST=example
_=/run/current-system/sw/bin/env
```
which is quite different from what I expected to have. If you need more logs or anything else I can send them, thanks!

PR #5223, opened 2025-09-09
fix(ui): fix change password behavior when confirming new password

luizhf42:
This pull request fixes the `ChangePassword` component behavior when validating the new password. Previously, the "confirm password" field would only be validated when its value changed — if the new password changed after, the validation wouldn't happen. Now it correctly validates when both values change. Additionally, the component now gives proper visual feedback when the request is sent with the current password with an incorrect value.

### DIFF
--- a/ui/src/components/User/ChangePassword.vue
+++ b/ui/src/components/User/ChangePassword.vue
@@ -22,6 +22,7 @@
 
           <v-text-field
             v-model="newPassword"
+            @update:model-value="handleNewPasswordChange"
             label="New password"
             :append-icon="showNewPassword ? 'mdi-eye' : 'mdi-eye-off'"
             :type="showNewPassword ? 'text' : 'password'"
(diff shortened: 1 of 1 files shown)


ui/src/components/User/ChangePassword.vue
```diff
@@ -135,6 +138,17 @@ const showNewPassword = ref(false);
 const showConfirmPassword = ref(false);
 const { name, email, username, recoveryEmail } = authStore;
 
+const handleNewPasswordChange = () => {
+  if (!newPasswordConfirm.value || !newPassword.value) return;
```
otavio, line 142: We should only consider it empty if both are empty or if both are equal to each other. In the way that it is now, if any of them is empty and I try to change the password, it is not going to identify that they are not equal.

otavio (changes requested):

ui/src/components/User/ChangePassword.vue
luizhf42: Actually, if any of the fields are empty, the Save Password  button becomes disabled, so the user can't try to change the password without populating both fields _and_ ensuring they're equal. I've done this to prevent the fields from showing errors even before the user has started to fill in the inputs. Do you think it's better to change that logic anyway?

PR #5229, opened 2025-09-10
feat(ssh): add prometheus metrics endpoints to return ssh data

henrybarreto:


docker-compose.dev.yml
```diff
@@ -73,3 +73,61 @@ services:
       - ./pkg:/go/src/github.com/shellhub-io/shellhub/pkg
       - ./api:/go/src/github.com/shellhub-io/shellhub/api
       - ./.golangci.yaml:/.golangci.yaml
+  prometheus:
```
henrybarreto: I don't know if we should have `promtheus` and `grafana` on the dev's docker compose.

What do you think?

otavio: We need a way to make sure that the metrics are there and working as expected; however, I am unsure if we should add a huge and heavy stack to the development environment. Wouldn't simpler alternatives be for this testing environment?

PR #5235, opened 2025-09-12
refactor(api): update store interface to use entity objects to save and delete documents

heiytor:


otavio (dismissed):

api/store/mongo/namespace.go
```diff
@@ -364,123 +318,65 @@ func (s *Store) NamespaceUpdate(ctx context.Context, tenant string, changes *mod
...
+			return 0, FromMongoError(err)
 		}
 
-		switch {
-		case res.MatchedCount < 1: // tenant not found
-			return nil, store.ErrNoDocuments
-		case res.ModifiedCount < 1: // member not found
-			return nil, ErrUserNotFound
+		for _, tenantID := range tenantIDs {
+			if err := s.cache.Delete(ctx, strings.Join([]string{"namespace", tenantID}, "/")); err != nil {
```
henrybarreto: Could we use the transaction's context into the cache like this?

```suggestion
			if err := s.cache.Delete(sCtx, strings.Join([]string{"namespace", tenantID}, "/")); err != nil {
```

api/store/mongo/namespace.go
```diff
@@ -364,123 +318,65 @@ func (s *Store) NamespaceUpdate(ctx context.Context, tenant string, changes *mod
...
+				log.Error(err)
+			}
 		}
 
-		objID, err := primitive.ObjectIDFromHex(memberID)
-		if err != nil {
-			return nil, err
+		collections := []string{"devices", "sessions", "firewall_rules", "public_keys", "recorded_sessions", "api_keys"}
```
henrybarreto: Should we delete the tunnels/web endpoints too?

PR #5289, opened 2025-09-29
fix(ssh): use CRLF on each line of banner message

henrybarreto:
The RFC 4252 defines that each line on banner message should be separated by a CRLF, `\r\n`, what wasn't happening because we're reading the message from a file.


ssh/server/server.go
```diff
@@ -66,7 +67,7 @@ func NewServer(opts *Options, tunnel *httptunnel.Tunnel, cache cache.Cache) *Ser
 			logger.Info("new connection established")
 
 			message := func(msg string) string {
-				return fmt.Sprintf("%s\r\n", msg)
+				return fmt.Sprintf("%s\r\n", strings.ReplaceAll(msg, "\n", "\r\n"))
```
otavio: Add a comment explaining that this is necessary due to this specific RFC.

Since you are using new lines, I don't believe you need to include a carriage return on the first line.

otavio (changes requested):

ssh/server/server.go
henrybarreto: Done. Do you think it is OK?

otavio:
Please fix the CI  errors before we merge.

PR #5290, opened 2025-09-30
fix(api): correct pagination index in device cleanup task

heiytor:
The Paginator expects 1-based page numbers, but the loop was passing 0-based indices. This caused MongoDB to receive negative values for $skip when page was 0, resulting in query failures.

Changed pagination to use page+1 to align with the 1-indexed expectation of the Paginator struct.


otavio:
My main question about this change is whether it is consistent with the rest of the code. If so, couldn't we fix that throughout the entire source code and avoid needing to do it manually?

PR #5302, opened 2025-10-01
feat(agent): add env variable to permit logins with empty passwords

henrybarreto:


otavio (dismissed):

pkg/agent/pkg/osauth/utils.go
```diff
@@ -9,3 +14,24 @@ type User struct {
...
+		// NOTE: By default, we dont allow login with empty password.
+		return false
+	}
+
+	if strings.ToLower(v) == "true" {
+		return true
+	}
+
+	return false
+}
```
otavio, line 26:
```
func PermitEmptyPasswords() bool {
    return strings.EqualFold(os.Getenv("SHELLHUB_PERMIT_EMPTY_PASSWORDS"), "true")
}

```

henrybarreto: Done

PR #5311, opened 2025-10-03
feat(ui): add a terminal theme picker

luizhf42:
This pull request adds a terminal theme picker to ShellHub's web terminal.

**Work in progress**

### DIFF
--- a/ui/src/components/Terminal/TerminalThemePicker.vue
+++ b/ui/src/components/Terminal/TerminalThemePicker.vue
@@ -1,0 +1,129 @@
+<template>
+  <v-navigation-drawer
+    v-model="showDrawer"
+    location="right"
+    disable-resize-watcher
+    width="300"
+    class="theme-drawer bg-v-theme-surface"
+    data-test="theme-drawer"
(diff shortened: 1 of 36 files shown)


ui/src/components/Terminal/TerminalThemePicker.vue
```diff
@@ -0,0 +1,108 @@
+<template>
+  <v-list>
+    <v-list-item v-if="terminalThemes.length === 0" class="text-center py-4" title="No themes available" />
+    <v-list-item
+      v-else
+      v-for="theme in terminalThemes"
+      :key="theme.name"
+      @click="selectTheme(theme)"
+      :class="{ 'v-list-item--active': selectedTheme === theme.name }"
```
gustavosbarreto: Wouldn’t it be better to use the `active` prop here instead of manually applying the 'v-list-item--active' class?

luizhf42: Yes, the prop is simpler and cleaner. I've changed this in the last push.

PR #5326, opened 2025-10-08
feat(ui): replace v-textarea with FileTextComponent for key inputs

luannmoreira:
# Description

This PR refactors the SSH key input fields across the Private and Public Key add/edit components by replacing the default `v-textarea` with a new reusable `FileTextComponent`. This improves input validation, paste/file handling, and provides a consistent UX for handling key data in various formats.

- The `FileTextComponent` abstracts shared behaviors such as:
- Validation using `isKeyValid`
- Paste and drag-drop support
- Dynamic error messaging
- Description/help text display

## Changes

- Replaced `<v-textarea>` with `<FileTextComponent>` in:
  -  `PrivateKeyAdd.vue`
  - `PrivateKeyEdit.vue`
  -  `PublicKeyAdd.vue`
  -  `PublicKeyEdit.vue`
- Passed appropriate props to FileTextComponent:
  -  `validator`, `invalid-message`, `description-text`, etc.
- Added error handling via `@error` events.
- Updated input field behavior for better user experience (e.g., auto-naming from file input).
- Registered Vuetify lab components `VFileUpload` and `VFileUploadItem`.
- Updated `isKeyValid` usage for both private and public keys.
- Minor cleanups and indentation consistency.

## Tests

This test suite verifies the behavior of the reusable `FileTextComponent.vue`, which supports both file-based and text-based SSH key input. It ensures the component provides consistent user experience, validation, and accessibility when handling SSH key data.

### Rendering Modes:

- Renders in file mode by default.
- Supports `textOnly` and `startInText` props to control initial/input mode.

### File Upload Handling:

- Accepts files with supported extensions and MIME types.
- Rejects unsupported types and oversized files.
- Emits appropriate events: `update:modelValue`, `file-name`, `file-processed`, and `error`.

### Validation:

- Applies custom validator function to both text and file input.
- Displays `invalidMessage` on failed validation.

### Paste and Drag-Drop Support:

- Accepts pasted files or text content.
- Automatically switches to text mode on pasted text.
- Processes pasted files using the same validation logic.

### Textarea Behavior:

- Binds model correctly.
- Displays errors and description text.
- Allows switching back to file mode 

### Event Handling:

- Tests emitted events for all key actions (input, error, clear, switch mode).
- Confirms error propagation on invalid inputs or `FileReader` failures.

### Edge Cases:

- Handles empty files, empty pasted content, and unexpected input gracefully.
- Validates behavior when clearing input or switching modes.

## How to Test

1. Open the UI and navigate to the Private Keys and Public Keys sections.

- Test all four dialogs:
  - Add Private Key
  - Edit Private Key
  - Add Public Key
  - Edit Public Key

- For each dialog:
  1. Paste a valid and invalid SSH key.
  2. Upload key files (e.g., .pem, .pub, .key, .txt).

2. Confirm that:

  - Valid keys enable the submit button.
  - Invalid keys display appropriate error messages.
  - Suggested names populate from filenames (on public key upload).
  - Confirm nothing breaks when switching between file/text modes (if applicable).

3. Validate that key formats listed in the description are still supported:

  - RSA, DSA, ECDSA (NIST P-*), ED25519
  - PEM (PKCS#1, PKCS#8), OpenSSH

### DIFF
--- a/ui/src/components/Fields/FileTextComponent.vue
+++ b/ui/src/components/Fields/FileTextComponent.vue
@@ -1,0 +1,537 @@
+<template>
+  <div
+    class="file-text-capture"
+    :class="{ 'is-disabled': disabled }"
+    ref="rootEl"
+    @paste.capture="onPaste"
+  >
+    <v-file-upload
+      v-if="mode === 'file' && !textOnly"
+      ref="uploadEl"
+      v-model="files"
+      :accept="effectiveAccept"
+      :max-size
+      :multiple="false"
+      :disabled
+      :show-size
+      density="comfortable"
+      :border="localError && 'opacity-100 error'"
+      :class="localError ? 'text-error bg-v-theme-surface' : 'py-3 bg-v-theme-surface'"
+      @update:model-value="onFiles"
+      data-test="file-text-capture"
+    >
+      <template #icon>
+        <slot name="icon">
+          <v-col cols="12" sm="12" class="py-0">
+            <v-icon :size="smAndDown ? 40 : 50">mdi-file-upload-outline</v-icon>
+          </v-col>
+        </slot>
+      </template>
+
+      <template #title>
+        <slot name="title">
+          <v-row no-gutters class="d-flex justify-start">
+            <v-col cols="12">
+              <div class="ftc-title text-subtitle-1 font-weight-medium">
+                Paste, drop a file or <span class="text-primary">click to browse</span>
+              </div>
+              <div class="ftc-sub text-body-2 mt-1">
+                Accepts .pub, .pem, .key, .txt (max {{ Math.round(props.maxSize / 1024) }} KB)
+              </div>
+              <div class="ftc-desc text-caption mt-1">
+                {{ descriptionText }}
+              </div>
+            </v-col>
+          </v-row>
+
+        </slot>
+      </template>
+
+      <template #item="{ file, props: itemProps }">
+        <v-file-upload-item
+          v-bind="itemProps"
+          :key="file?.name || file?.lastModified"
+          lines="one"
+          nav
+        >
+          <template #prepend>
+            <v-avatar size="32" rounded>
+              <v-icon>mdi-file-document-outline</v-icon>
+            </v-avatar>
+          </template>
+
+          <template #clear="{ props: clearProps }">
+            <v-btn
+              color="error"
+              v-bind="clearProps"
+              icon="mdi-close"
+              size="small"
+              variant="text"
+              @click.stop="onClearClick(clearProps)"
+              title="Remove file"
+            />
+          </template>
+        </v-file-upload-item>
+      </template>
+    </v-file-upload>
+
+    <div v-else class="mt-2">
+      <v-textarea
+        ref="textareaRef"
+        v-model="textModel"
+        :label="textareaLabel"
+        :hint="textareaHint"
+        :messages="descriptionText"
+        :disabled
+        auto-grow
+        rows="3"
+        :error-messages="localError"
+        data-test="ftc-textarea"
+        @update:model-value="onTextInput"
+      >
+        <template #append-inner>
+          <v-btn
+            icon="mdi-upload"
+            variant="text"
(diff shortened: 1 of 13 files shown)


ui/src/components/FileTextComponent.vue
```diff
@@ -0,0 +1,470 @@
...
+  files.value = val;
+  await nextTick();
+  internalUpdate.value = false;
+};
+
+// v-model binding
+const textModel = computed({
+  get: () => props.modelValue ?? "",
+  set: (v: string) => emit("update:modelValue", v ?? ""),
+});
```
luizhf42: Wouldn't it be better to use `defineModel`?

ui/src/components/FileTextComponent.vue
```diff
@@ -0,0 +1,470 @@
...
+  /**
+   * Density (spacing) of the internal Vuetify components.
+   * Useful for compact dialogs or large modals.
+   */
+  density?: Density;
+
+  /**
+   * Visual variant of the Vuetify components (plain, outlined, etc.)
+   */
+  variant?: Variant;
```
luizhf42: Since ShellHub is starting to standardize the inputs with the outlined variant, don't you think it's better to keep that component variant always as outlined and remove that prop?

ui/src/components/FileTextComponent.vue
luannmoreira: Yes! It is a better way of handling the v-model binding by Vue 3.3+ standards, I've made this change in my last push.

ui/src/components/FileTextComponent.vue
luannmoreira: At first I wouldn't because I wanted to treat this component as an internal fully customizable one, but by default, `v-fileupload` doesn't have a `variant` prop, so at the end it was a useless piece of code, we could've made `v-textarea` more customizable, but as you said, `v-textarea` is set by default as outlined too and we don't have current plans of changing that.

 I've removed this piece of code in my last push.

ui/src/components/PublicKeys/PublicKeyAdd.vue
```diff
@@ -107,17 +112,21 @@
...
-          data-test="data-field"
-          rows="3"
-          auto-grow
+          class="mb-2"
+          :accept="accept"
+          :textarea-label="'Public key data'"
+          :max-size="maxSize"
+          :validator="(t) => isKeyValid('public', t)"
+          invalid-message="This is not a valid public key."
+          :enable-paste="true"
```
luizhf42: I don't think there's a need to pass those `accept`, `max-size`, and `enable-paste` props since they equal the default values from `FileTextComponent`. Also, `textarea-label` can be passed as a regular string instead of using a v-bind.

Applies to the other Key Add/Edit components too.

luannmoreira: I've seen to rolled back these patterns on a late push, thanks for this review, I've updated it on my last push.

ui/src/components/PublicKeys/PublicKeyAdd.vue
```diff
@@ -207,6 +223,16 @@ const {
   resetField: resetPublicKeyData,
 } = useField<string>("publicKeyData", yup.string().required(), { initialValue: "" });
 
+const accept = ".pub,.pem,.key,.txt,text/plain,application/x-pem-file,application/octet-stream";
```
luizhf42: I think `accept` could also be removed here since it's a duplicate of the `FileTextComponent` default value.

ui/src/components/Fields/FileTextComponent.vue
```diff
@@ -0,0 +1,544 @@
...
+  /**
+   * Whether to show the file size in the upload list (Vuetify prop).
+   */
+  showSize?: boolean;
+
+  /**
+   * Density (spacing) of the internal Vuetify components.
+   * Useful for compact dialogs or large modals.
+   */
+  density?: Density;
```
luizhf42: This prop is unused.

PR #5382, opened 2025-10-17
feat(gateway): enable or disable access logs from nginx

taka41s:
## Summary

This pull request introduces a configuration option to enable or disable Nginx access logs.

## Context

Access logs were causing disk space issues, so this change allows us to control their usage depending on the environment or operational needs.

## Details

- Added a configurable flag to toggle Nginx access logs.


gateway/nginx/nginx.conf
```diff
@@ -78,8 +78,12 @@ http {
...
+  
+    {{ if $cfg.EnableAccessLogs }}
     access_log /dev/stdout nginxlog_json;
+    {{ else }}
+    access_log off;
+    {{ end }}
 
     include /etc/nginx/conf.d/*.conf;
-}
+}
```
otavio:
```suggestion
}

```

otavio (changes requested):

otavio:
Could you please squash the changes into a single commit? I'm unsure whether we should do that. I'm questioning if this is the best approach or if you should instead modify your server configuration to rotate logs and clean up as needed. I'm considering this because if any issues arise, it could be difficult to identify the root cause without gateway information. What's your opinion on this?

taka41s:
> Could you please squash the changes into a single commit? I'm unsure whether we should do that. I'm questioning if this is the best approach or if you should instead modify your server configuration to rotate logs and clean up as needed. I'm considering this because if any issues arise, it could be difficult to identify the root cause without gateway information. What's your opinion on this?

### 1. **Cost Savings**
- Access logs = 80-90% of total log volume in high-traffic environments

### 2. **Performance Troubleshooting**
- Useful for isolating log I/O bottlenecks during performance issues
- Load testing scenarios where logs can skew results

### 3. **Safe Default (Opt-out)**
- Logs **enabled by default** - zero impact on existing deployments
- Fully backward compatible

### 4. **Industry Standard**
- NGINX has native `access_log off;`

### 5. **Minimal Implementation**
- 1 config field + 4 lines in template
- Easy to test, no side effects

### 6. **Real Use Cases**
- Health checks polluting logs (millions of redundant entries)
- Temporary disable during incidents (disk full)
- GDPR/privacy compliance scenarios



> "This adds optional access log control via env var while keeping logs **enabled by default** (current behavior). Useful for cost reduction in cloud environments, performance troubleshooting, and load testing. Implementation is minimal (4 lines), backward compatible, and follows industry standards. Zero risk for existing deployments."

otavio:
Could you please squash the changes into a single commit?

.env
```diff
@@ -184,3 +184,6 @@ SHELLHUB_INTERNAL_HTTP_CLIENT_API_BASE_URL=http://api:8080
 
 # The base URL for the Enterprise service.
 SHELLHUB_INTERNAL_HTTP_CLIENT_ENTERPRISE_BASE_URL=http://cloud:8080
+
+# Set false to disable access logs for gateway nginx
+ENABLE_ACCESS_LOGS=true
```
otavio: I was reviewing the change one last time, and I think we should specify the prefix for consistency and clarify that it is related to the gateway. Using the term 'access' is misleading here because it could refer to anything, so having the prefix and specific details for the gateway will reduce this confusion. Could you do this one last time, and I can then approve the change?

taka41s: You mean something like `GATEWAY_ACCESS_LOGS` or `NGINX_ACCESS_LOGS`?

otavio: `SHELLHUB_GATEWAY_ACCESS_LOGS`

PR #5403, opened 2025-10-21
refactor(ui): move DataTable to Tables dir and update imports

luannmoreira:
# Description

This PR reorganizes the DataTable.vue component by moving it into a dedicated Tables/ directory. All import paths across both the admin and client UI components have been updated accordingly. The old DataTable.vue file has been removed from its original location.

## Changes
 
- Moved ui/src/components/DataTable.vue → ui/src/components/Tables/DataTable.vue
- Updated all component imports referencing DataTable.vue in:
- Admin UI (ui/admin/src/components/...)
- Client UI (ui/src/components/...)
- Deleted the old DataTable.vue file

## Why

This change improves directory structure and component organization by grouping all table-related components under a single Tables/ folder. It aligns with a cleaner, modular approach to UI development and makes table components easier to locate and maintain.

## Affected Areas

All components using DataTable.vue, including:
Announcements, Devices, Sessions, Users, API Keys, Tags, Namespaces, Connectors, Firewall Rules, Public Keys, Web Endpoints

## Testing
 
1. Verified UI components render correctly with the updated import path
2. No changes to DataTable.vue logic—just a file move and refactor
3. Confirmed old file was safely removed

### DIFF
--- a/ui/tests/components/Tables/DataTable.spec.ts
+++ b/ui/tests/components/Tables/DataTable.spec.ts
@@ -1,0 +1,207 @@
+import { createVuetify } from "vuetify";
+import { mount, VueWrapper, flushPromises } from "@vue/test-utils";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import DataTable from "@/components/Tables/DataTable.vue";
+
+vi.mock("vuetify", async () => {
+  const actual = await vi.importActual<typeof import("vuetify")>("vuetify");
+  return {
+    ...actual,
+    useDisplay: () => ({
+      smAndDown: { value: false },
+      thresholds: {
+        value: {
+          sm: 600,
+          md: 960,
+          lg: 1280,
+          xl: 1920,
+          xxl: 2560,
+        },
+      },
+    }),
+  };
+});
+
+type DataTableWrapper = VueWrapper<InstanceType<typeof DataTable>>;
+
+describe("DataTable", () => {
+  const vuetify = createVuetify();
+  let wrapper: DataTableWrapper;
+
+  const defaultHeaders = [
+    { text: "Name", value: "name", sortable: true },
+    { text: "Owner", value: "owner", sortable: false },
+    { text: "Created", value: "createdAt", sortable: true },
+  ] as const;
+
+  const mountWrapper = (
+    props: Partial<InstanceType<typeof DataTable>["$props"]> = {},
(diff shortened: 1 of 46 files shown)


ui/tests/components/Tables/DataTable.spec.ts
luizhf42, line 1: I think it's better to use the `wrapper.find` method with `data-test` or something more "stable" than Vuetify's styling classes.

ui/tests/components/Tables/DataTable.spec.ts
```diff
@@ -0,0 +1,217 @@
+import { createVuetify } from "vuetify";
+import { DOMWrapper, mount, VueWrapper, flushPromises } from "@vue/test-utils";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+// 🔁 ADJUST THIS PATH to your component file
```
luizhf42:
```suggestion

```

ui/tests/components/Tables/DataTable.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  afterEach(() => {
+    vi.clearAllMocks();
+    wrapper.unmount();
+    document.body.innerHTML = "";
+  });
+
+  it("is a Vue instance", () => {
+    expect(wrapper.exists()).toBe(true);
+  });
```
luizhf42, line 84: Is this assertion necessary? If the wrapper fails to mount, all the other tests will fail, so this test doesn't provide much help.
Obviously, it doesn't cause any harm, but for me, it's almost dead code. What do you think?

luannmoreira: As it is a standard test we've been using for a while now, I've preferred to keep it, even if it's redundant at first, it keeps the eyes open for a possible "testing the own test" situation, if this test returns an error, and the other ones pass, it will show that the test file is testing itself and not the real component, showing us a bad pattern on testing, however, I've changed the `wrapper.exists()` to `wrapper.vm.exists()` as it's the binding between the component and the test.

ui/tests/components/Tables/DataTable.spec.ts
luannmoreira: I've made changes in the `DataTable.vue` and the `DataTable.spec.ts` to respect more data-table attributes than finding the Vuetify styling classes, can you check?

luizhf42 (approved):

PR #5408, opened 2025-10-22
refactor(ui): centralize tag-based filtering logic in TagSelector

luannmoreira:
# Description

This PR improves the clarity and maintainability of tag-based filtering in the `TagSelector` component.

## Changes

Replaced `setVariantListFilter` with a cleaner `setFilter()` helper that sets the appropriate store filter based on variant.

Updated `getItems()` to build the tag filter inline with consistent formatting.

Improved formatting in `tagIsSelected()` for readability.

Minor cleanup of unnecessary ternaries and improved whitespace consistency.

## Why

- Centralizes and simplifies how encoded filters are built and applied.
- Prevent a double request incident in the request.

### DIFF
--- a/ui/src/components/Tags/TagSelector.vue
+++ b/ui/src/components/Tags/TagSelector.vue
@@ -62,7 +62,6 @@
 
 <script setup lang="ts">
 import { computed, onMounted, onUnmounted, ref, watch, nextTick } from "vue";
-import axios, { AxiosError } from "axios";
 import handleError from "@/utils/handleError";
 import useSnackbar from "@/helpers/snackbar";
 import useContainersStore from "@/store/modules/containers";
@@ -94,9 +93,13 @@ let observer: IntersectionObserver | null = null;
 
 const hasMore = computed(() => tagsStore.numberTags > fetchedTags.value.length);
 
-const getTagName = (tag: ITag): string => (typeof tag === "string" ? tag : tag.name);
+const getTagName = (tag: ITag): string => typeof tag === "string" ? tag : tag.name;
+
 const getSelectedTagNames = (): string[] => selectedTags.value.map((t) => getTagName(t));
-const tagIsSelected = (tag: ITag): boolean => selectedTags.value.some((sel) => getTagName(sel) === getTagName(tag));
+
+const tagIsSelected = (tag: ITag): boolean => selectedTags.value.some(
+  (sel) => getTagName(sel) === getTagName(tag),
+);
 
 const resetPagination = (): void => {
   currentPage.value = 1;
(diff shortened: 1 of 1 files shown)


ui/src/components/Tags/TagSelector.vue
```diff
@@ -132,22 +132,31 @@ const loadInitialTags = async (): Promise<void> => {
...
-  }];
-  const encodedFilter = btoa(JSON.stringify(filter));
+const selectTag = async (item: ITag): Promise<void> => {
   try {
-    await fetchDevices(encodedFilter);
+    tagsStore.setSelected({ variant: props.variant, tag: item });
+    const names = getSelectedTagNames();
+    const encoded = buildEncodedTagsFilter(names);
+    setVariantListFilter(encoded);
   } catch (error: unknown) {
```
luizhf42: In my opinion, the try/catch block isn't necessary since nothing called here returns a promise or throws errors. What do you think?

luannmoreira: I've changed all the PR code, since it could be solved with less changes, I'm resolving the conversation because this piece of code doesn't exist anymore.

ui/src/components/Tags/TagSelector.vue
```diff
@@ -132,19 +136,22 @@ const loadInitialTags = async (): Promise<void> => {
...
-    container: () => containersStore.fetchContainerList({ filter }),
-  }[props.variant];
-  await fetch();
+const fetchDevices = async (filter?: string): Promise<void> => {
+  const encoded = filter && filter.length ? filter : undefined;
+  if (props.variant === "device") {
+    devicesStore.deviceListFilter = encoded;
+  } else {
+    containersStore.containerListFilter = encoded;
+  }
```
luizhf42: That method doesn't need to be async nor return a Promise, since it doesn't perform any async operation. Also, `fetchDevices` doesn't seem like a good name.

ui/src/components/Tags/TagSelector.vue
```diff
@@ -132,19 +136,22 @@ const loadInitialTags = async (): Promise<void> => {
...
-  }];
+  const filter = [
+    {
+      type: "property",
+      params: { name: "tags.name", operator: "contains", value: tagNames },
+    },
+  ];
   const encodedFilter = btoa(JSON.stringify(filter));
   try {
     await fetchDevices(encodedFilter);
```
luizhf42: Thus, the try/catch block isn't necessary here, nor the `async` keyword or the `Promise<void>` return.

luannmoreira: Removed useless async and await code in last push

ui/src/components/Tags/TagSelector.vue
luannmoreira: Removed useless async and await code in last push and changed the name to setFilter as it's more appropriate to the actions made in the function.

PR #5414, opened 2025-10-23
fix(ui): improve device search UX in WebEndpointCreate form

luannmoreira:
# Description

This PR resolves an issue where the `v-autocomplete` in the `WebEndpointCreate.vue` component was not searching device options correctly.

## Problem

- The search input wasn't properly bound or controlled, causing:
- Search queries to not trigger expected results
- Filters to persist incorrectly across dialog sessions

## Fix

- Bound search input via `v-model:search="deviceSearch"`
- Replaced `@update:search="fetchDevices"` with a proper handler: `onSearchUpdate`
- Cleared search filters and refetched device list on empty input and dialog close
- Added cleanup in `resetFields()` and a `clearFilterAndRefetch() `utility
- Adjusted base64 encoding logic for consistency

## Testing

- [x] Typing in autocomplete filters device list in real time
- [x] Clearing search input resets to full device list
- [x] Dialog reset works as expected on close
- [x] No regression in form validation or submission logic

### DIFF
--- a/ui/src/components/WebEndpoints/WebEndpointCreate.vue
+++ b/ui/src/components/WebEndpoints/WebEndpointCreate.vue
@@ -27,6 +27,8 @@
               class="mt-1"
               label="Host"
               :error-messages="hostError"
+              hint="IPv4 or IPv6 only"
+              persistent-hint
               data-test="host-text"
             />
           </v-col>
@@ -39,6 +41,7 @@
               label="Port"
               :error-messages="portError"
               variant="outlined"
+              type="number"
               data-test="port-text"
             />
           </v-col>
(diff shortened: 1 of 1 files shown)


ui/src/components/WebEndpoints/WebEndpointCreate.vue
```diff
@@ -152,33 +166,47 @@ const predefinedTimeouts = ref([
...
+    }),
   { initialValue: "127.0.0.1" },
 );
 
 const { value: port, errorMessage: portError, resetField: resetPort } = useField<number>(
   "port",
-  yup.number().integer().max(65535).required(),
+  yup.number().typeError("Port is required").integer().min(1)
+    .max(65535)
+    .required(),
```
luizhf42: I think the "Port is required" message should be the `required()` parameter, and the `typeError()` should contain something like "Port is a number between 1 and 65535"

PR #5500, opened 2025-11-10
refactor(ui): redesign layout and improve user metadata display

luannmoreira:
# Description

This PR redesigns the `UserDetails` view for improved usability, clarity, and extensibility. It introduces a new layout for displaying user metadata and modernizes the action UI. It also includes minor supporting refactors to related components.

### Redesigned the layout of `UserDetails.vue`:

- Uses a two-column structure to better organize user information

### Adds labels and chips for fields like:
 
- UID
- Name, Email, Username
- Status (with status chip)
- MFA status
- Email marketing opt-in
- Namespace limits
- Auth methods
- Timestamps (`created_at, last_login`)
- Replaces the old icon-based actions (login and delete) with a dropdown menu
- Moves control of the delete dialog outside of `UserDelete.vue` via `v-model`
- Updates `IAdminUser` interface to support optional and _nullable_ fields
- Updates unit and snapshot tests to reflect new structure and behavior

## Motivation

- The previous design was minimal and lacked structured metadata presentation. This redesign provides:
- A more scalable and readable layout
- Better support for future user attributes
- Cleaner separation of concerns between components
- Improved accessibility and UI consistency

## Testing

- All existing unit tests updated and passing
- New layout and conditional rendering verified via snapshots and manual UI checks


ui/admin/src/components/User/UserDelete.vue
```diff
@@ -44,7 +27,7 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits(["update"]);
-const showDialog = ref(false);
+const showDialog = defineModel<boolean>({ default: false });
```
luizhf42: I prefer using the `required: true` option here, since the `v-model` is, in fact, required for the dialogs, and if we forget to add it, there will be errors in the build process.

ui/admin/tests/unit/views/UserDetails/index.spec.ts
```diff
@@ -42,39 +50,86 @@ describe("User Details", async () => {
...
   });
 
-  it("Renders the component", () => {
+  it("is a Vue instance", () => {
+    expect(wrapper.exists()).toBe(true);
+  });
+
+  it("renders the component", () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
```
luizhf42: Since the test suite is very complete and tests all the important elements, there is no need for a snapshot test, since it serves the same purpose, with the downside of needing to be updated almost every time there's a Vuetify bump. What do you think?

ui/admin/src/interfaces/IUser.ts
```diff
@@ -4,21 +4,21 @@ export type UserAuthMethods = Array<"saml" | "local">;
...
-  }
-  preferences: {
+  recovery_email?: string | null;
+  mfa?: { enabled?: boolean } | null;
+  namespacesOwned: number;
+  preferences?: {
     auth_methods: UserAuthMethods;
-  }
+  };
+  password?: string;
```
luizhf42: There's no `password` field returned in the user details API response.

ui/admin/tests/unit/views/UserDetails/index.spec.ts
luannmoreira: I agree, it is redundant as there's tests meant for the instance and each element on the component, proving it's rendering.

PR #5532, opened 2025-11-17
refactor(api): migrate membership invitations to a dedicated collection

heiytor:


openapi/spec/cloud-openapi.yaml
```diff
@@ -112,14 +112,14 @@ paths:
     $ref: paths/api@connector@{uid}.yaml
   /api/connector/{uid}/info:
     $ref: paths/api@connector@{uid}@info.yaml
-  /api/namespaces/{tenant}/members/accept-invite:
-    $ref: paths/api@namespaces@{tenant}@members@accept-invite.yaml
-  /api/namespaces/{tenant}/members/decline-invite:
-    $ref: paths/api@namespaces@{tenant}@members@decline-invite.yaml
-  /api/namespaces/{tenant}/members/invites:
-    $ref: paths/api@namespaces@{tenant}@members@invites.yaml
+  /api/namespaces/{tenant}/invitations/links:
```
otavio, line 115: As those empty points are being added here, I think this comment should be included in the previous comment, so there is no need for us to rename any point that we are just creating.

PR #5549, opened 2025-11-21
feat(ui): enhance NamespaceDetails view with full namespace info

luannmoreira:
# Description:

This PR enhances the `NamespaceDetails` component to present a richer and more structured view of namespace data. It also introduces namespace deletion support, both in the UI and the underlying store/API layer.

## What's New

- `NamespaceDetails` View Enhancements
- Redesigned layout using `v-card` and responsive `v-col/v-row` grid.

###  Added detailed namespace metadata:

- Name, Tenant ID, Owner (email and ID)
- Device breakdown: total, accepted, pending, rejected
- Settings: session recording status, connection announcements
- Created at timestamp
- Namespace type chip (e.g., team)
- Improved member list:
- Email, ID, role (with icon), status, and added/expiry timestamps
- Proper formatting with `v-chip` and structured layout

#### Added edit/delete controls as dropdown items in both:

- `NamespaceList.vue` (inline per row)
- `NamespaceDetails.vue` (actions menu)
- `Converted NamespaceEdit.vue` to use `v-model` and emit `@update` on save

#### Store/API Updates

- `namespaces.ts `(store): added `deleteNamespace` action
- `api/namespaces.ts`: added `deleteNamespace` method

### Updated unit snapshots and test coverage for:

- `NamespaceDetails.vue`
- `NamespaceEdit.vue`
- `NamespaceList.vue`

Added assertions for richer rendering and deletion workflows

## Checklist

- [x] UI changes visually verified
- [x] Unit tests updated or added
- [x] Pinia Store and API methods extended
- [x] No regressions in namespace CRUD flows

### DIFF
--- a/ui/admin/src/components/Namespace/NamespaceDelete.vue
+++ b/ui/admin/src/components/Namespace/NamespaceDelete.vue
@@ -1,0 +1,66 @@
+<template>
+  <MessageDialog
+    v-model="showDialog"
+    title="Namespace Deletion"
+    icon="mdi-delete-alert"
+    icon-color="error"
+    confirm-color="error"
+    confirm-text="Remove"
+    :confirm-loading="isLoading"
+    cancel-text="Close"
+    confirm-data-test="remove-btn"
+    cancel-data-test="close-btn"
+    @close="showDialog = false"
+    @confirm="remove"
+    @cancel="showDialog = false"
+  >
+    <p data-test="content-text">
+      This action cannot be undone. This will permanently delete
+      <strong>{{ displayOnlyTenCharacters(name) }}</strong> and its related
+      data.
+    </p>
+  </MessageDialog>
+</template>
+
+<script setup lang="ts">
+import { ref } from "vue";
+import { useRouter, useRoute } from "vue-router";
+import { displayOnlyTenCharacters } from "@/utils/string";
+import handleError from "@/utils/handleError";
+import useSnackbar from "@/helpers/snackbar";
+import MessageDialog from "@/components/Dialogs/MessageDialog.vue";
+import useNamespacesStore from "@admin/store/modules/namespaces";
+
+const props = defineProps<{ tenant: string; name: string }>();
+const emit = defineEmits(["update"]);
+
+const namespacesStore = useNamespacesStore();
+const snackbar = useSnackbar();
+const router = useRouter();
+const route = useRoute();
+
+const showDialog = defineModel<boolean>({ required: true });
+const isLoading = ref(false);
+
+const remove = async () => {
+  isLoading.value = true;
+  try {
+    await namespacesStore.deleteNamespace(props.tenant);
+    snackbar.showSuccess("Namespace deleted successfully.");
+
+    if (route.name === "namespaceDetails") {
+      await router.push({ name: "namespaces" });
+    } else {
+      emit("update");
+    }
+    showDialog.value = false;
+  } catch (error: unknown) {
+    snackbar.showError("An error occurred while deleting the namespace.");
+    handleError(error);
+  } finally {
+    isLoading.value = false;
+  }
+};
+
+defineExpose({ isLoading });
+</script>
--- a/ui/admin/src/components/Namespace/NamespaceEdit.vue
+++ b/ui/admin/src/components/Namespace/NamespaceEdit.vue
@@ -1,23 +1,4 @@
 <template>
-  <v-tooltip
-    bottom
-    anchor="bottom"
-  >
-    <template #activator="{ props }">
-      <v-icon
-        tag="button"
-        dark
-        v-bind="props"
-        tabindex="0"
-        aria-label="Edit Namespace"
-        data-test="dialog-btn"
-        icon="mdi-pencil"
-        @click="showDialog = true"
-      />
-    </template>
-    <span>Edit</span>
-  </v-tooltip>
-
   <FormDialog
     v-model="showDialog"
     title="Edit Namespace"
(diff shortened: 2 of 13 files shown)


ui/admin/src/components/Namespace/NamespaceDelete.vue
```diff
@@ -0,0 +1,69 @@
...
+const namespacesStore = useNamespacesStore();
+const snackbar = useSnackbar();
+const router = useRouter();
+const route = useRoute();
+
+const showDialog = defineModel<boolean>({ required: true });
+const isLoading = ref(false);
+
+const name = computed(() => props.name);
+const tenant = computed(() => props.tenant);
```
luizhf42: I think there's no need for those computed properties. Using the props directly already works.

ui/admin/src/components/Namespace/NamespaceEdit.vue
```diff
@@ -118,16 +102,21 @@ const closeDialog = () => {
...
-      settings: { session_record: sessionRecord.value },
+      settings: {
+        ...props.namespace.settings,
+        session_record: sessionRecord.value,
+      },
     });
     await namespacesStore.fetchNamespaceList();
     snackbar.showSuccess("Namespace updated successfully.");
     showDialog.value = false;
+    emit("update");
```
luizhf42: There are two calls to the namespace list route: from the direct store call and from the `update` event emitted, which triggers the method in the parent. You could remove the store call of that component, leaving only the emit.

ui/admin/src/components/Namespace/NamespaceList.vue
```diff
@@ -25,12 +25,18 @@
...
+              tabindex="0"
+              class="text-decoration-underline cursor-pointer"
+              @click="goToUser(namespace.owner)"
+              @keyup.enter="goToUser(namespace.owner)"
+            >
+              {{ getOwnerEmail(namespace) }}
+            </span>
+            <span v-else>
+              {{ namespace.owner }}
+            </span>
```
luizhf42: Don't you think we could show the link in any case (email or ID), instead of showing only for the email?

ui/admin/src/components/Namespace/NamespaceList.vue
```diff
@@ -136,3 +222,9 @@ onMounted(async () => {
   await fetchNamespaces();
 });
 </script>
+
+<style scoped>
+.cursor-pointer {
+  cursor: pointer;
+}
+</style>
```
luizhf42, line 224: Vuetify already has a `.cursor-pointer` class: https://vuetifyjs.com/en/styles/cursor/#cursor

ui/admin/src/views/NamespaceDetails.vue
```diff
@@ -132,5 +327,42 @@ const sumDevicesCount = (namespace: IAdminNamespace) => {
...
+  font-size: 0.75rem;
+  font-weight: 500;
+  text-transform: uppercase;
+  letter-spacing: 0.1666666667em;
+  line-height: 2.667;
+}
+
+.cursor-pointer {
+  cursor: pointer;
+}
```
luizhf42:
```suggestion

```

ui/admin/src/views/NamespaceDetails.vue
```diff
@@ -2,125 +2,320 @@
...
-                @keyup="goToUser(namespace.owner)"
-              >{{ value }}</span>
-            </div>
-            <div v-else>
+                data-test="namespace-member-email"
+                @click="goToUser(member.id)"
+                @keyup.enter="goToUser(member.id)"
+              >
+                {{ member.email }}
+              </span>
```
luizhf42: Wouldn't it be better to use a `<router-link>` here? No need for methods and event listeners. Also, the default tag is an `<a>`, which is more semantic.

ui/admin/src/components/Namespace/NamespaceList.vue
luizhf42: Also, the `<router-link>` suggestion applies here too.

ui/admin/src/views/NamespaceDetails.vue
luannmoreira: Agreed, It's better to use router-link in this case, I've made this change in my last push [041890c](https://github.com/shellhub-io/shellhub/commit/041890cf092db27ce2d6e809632938ef7b522ea0)

ui/admin/src/views/NamespaceDetails.vue
luannmoreira: I've made this change in my last push since vuetify has a built-in class for cursor-pointer [041890c](https://github.com/shellhub-io/shellhub/commit/041890cf092db27ce2d6e809632938ef7b522ea0)

ui/admin/src/components/Namespace/NamespaceList.vue
luannmoreira: I've made this change in my last push since vuetify has a built-in class for cursor-pointer [041890c](https://github.com/shellhub-io/shellhub/commit/041890cf092db27ce2d6e809632938ef7b522ea0)

ui/admin/src/components/Namespace/NamespaceList.vue
luannmoreira: I've changed this component from span to router-link and created a method to get the email or ID of the owner in my last push [041890c](https://github.com/shellhub-io/shellhub/commit/041890cf092db27ce2d6e809632938ef7b522ea0)

ui/admin/src/components/Namespace/NamespaceEdit.vue
luannmoreira: I've removed the fetch call from this component as the emit should be the one to return a fetch request on my last push [041890c](https://github.com/shellhub-io/shellhub/commit/041890cf092db27ce2d6e809632938ef7b522ea0)

ui/admin/src/components/Namespace/NamespaceDelete.vue
luannmoreira: I've removed this computed properties and kept only the props in the NamespaceDelete in my last push  [041890c](https://github.com/shellhub-io/shellhub/commit/041890cf092db27ce2d6e809632938ef7b522ea0)

ui/admin/src/views/NamespaceDetails.vue
```diff
@@ -2,135 +2,367 @@
...
+          <div data-test="namespace-owner-field">
+            <div class="item-title">Owner:</div>
+            <router-link
+              :to="{ name: 'userDetails', params: { id: namespace.owner } }"
+              class="unstyled-link cursor-pointer"
+            >
+              <a class="text-decoration-underline cursor-pointer">
+                {{ getOwnerLabel(namespace) }}
+              </a>
+            </router-link>
```
luizhf42, line 100: There's no need to put an `<a>` here since the `<router-link>` automatically becomes an anchor tag on render, so you can just put the text directly in the link's slot. The cursor pointer and underline styles are also already present in its implementation.

luannmoreira: Oh, that was a mistake left on the code, thank you for the review, fixed it in [d08dcd6](https://github.com/shellhub-io/shellhub/commit/d08dcd643031d22e4f69f6222c75cda6b2754dd6)

luizhf42 (approved):

PR #5567, opened 2025-11-27
refactor(ui): enhance items-per-page input handling in DataTable

luannmoreira:
# Summary

This PR refactors the pagination controls in the `DataTable` component to improve both layout and input handling for the "Items per page" control.

## What’s Changed

- Replaced div-based layout with Vuetify's `v-row` and `v-col` for improved responsiveness and alignment.
- Introduced `internalItemsPerPage` as a controlled ref to decouple UI input from the external model.
- Added input sanitization logic to enforce values between 1 and 100.
- Implemented numeric-only input restrictions via keydown and paste events.
- Added validation rules with error messages for invalid inputs.
- Ensured model updates trigger pagination reset to page 1.

## Motivation

Previously, the input allowed arbitrary values and immediate binding, which could cause invalid or disruptive pagination behavior. This update adds guardrails, improving UX and data integrity.

How to Test

1. Use the "Items per page" combobox in the DataTable
- Try entering:
- Valid numbers (e.g., 10, 25, 50)
- Invalid numbers (e.g., 0, 200)
- Non-numeric input (e.g., letters, symbols)
2. Observe:
- Validation messages
- Auto-correction on blur or enter
- Pagination resets to page 1 after change

Notes

- This does not introduce new features but hardens the component.
- Should be safe to merge as-is, but QA feedback on edge cases is welcome.

### DIFF
--- a/ui/src/components/Tables/DataTable.vue
+++ b/ui/src/components/Tables/DataTable.vue
@@ -65,58 +65,88 @@
       data-test="loading"
     />
 
-    <div
+    <v-row
       v-if="itemsPerPageOptions?.length"
-      class="d-flex w-100 justify-end align-center"
+      class="w-100 pt-3"
+      align="center"
+      justify="end"
       data-test="pager"
     >
-      <span
-        class="text-subtitle-2 mr-4"
-        data-test="ipp-label"
-      >Items per page:</span>
+      <v-col
+        cols="auto"
+        class="pa-0"
+      >
+        <span
+          class="text-subtitle-2 mr-4"
+          data-test="ipp-label"
+        >
+          Items per page:
+        </span>
+      </v-col>
 
-      <div>
+      <v-col
+        cols="auto"
+        class="pa-0"
+      >
         <v-combobox
-          v-model="itemsPerPage"
+          v-model="internalItemsPerPage"
           :items="itemsPerPageOptions"
           outlined
           variant="underlined"
-          hide-details
-          class="mb-4"
+          center-affix
+          type="number"
+          hide-spin-buttons
+          hide-details="auto"
+          hide-no-data
+          class="mb-4 mr-1 w-100"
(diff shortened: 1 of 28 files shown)


ui/src/components/Tables/DataTable.vue
```diff
@@ -134,8 +165,65 @@ const props = defineProps<{
...
+
+  if (num < 1) return "Minimum is 1";
+  if (num > 100) return "Maximum is 100";
+
+  return true;
+};
+
+const onItemsPerPageInput = (value: number) => {
+  internalItemsPerPage.value = value;
+};
```
luizhf42: Why do you need that handler? I think a pure `v-model` in the combobox would be sufficient here.

luannmoreira: The internal input should accept temporary invalid values to show the validation (e.g., empty string or numbers above or below 100 and 1) but parent must always hold a valid number.

luizhf42: Since the handler only modifies the `internalItemsPerPage`, the behavior is the same as using a `v-model`, which is just a [syntax sugar](https://vuejs.org/guide/components/v-model#under-the-hood) for that implementation. Therefore, it should work normally, since all the other validations and sanitizing would run normally in the internal ref, without passing invalid values to the parent component.

luannmoreira: Oh, you're absolutely right, I've misunderstood your first comment, I see that the `@update:model-value` was a redundant and not needed v-on since v-model would've had the same behavior. 

This was solved in my [last push](https://github.com/shellhub-io/shellhub/pull/5567/commits/0b903daf6551b6314b45bb81e7ec75feca86b8a3), thank you for the clarification.

luizhf42 (approved):

PR #5588, opened 2025-12-03
fix: ensure nsenter and setpriv match the target architecture

CloudCray:
`setpriv` is causing an error on some `aarch64` systems - this PR ensures that `setpriv` is compiled for the target architecture, rather than the build architecture.


gustavosbarreto:
@CloudCray  Thanks for the fix! Could you update the commit message to follow conventional commits format? Something like:
`fix: ensure nsenter and setpriv match the target architecture`

PR #5847, opened 2026-02-19
chore(ui,openapi): update OpenAPI Spec

luizhf42:
This pull request generates the new OpenAPI specification for the UI, applying the latest updates, refactors, and fixes in the API documentation.


luizhf42:
> I think it would be better if we don't commit those files, as they will be constantly changing.

Makes sense. Since the intention is to improve the API documentation for the new UI as much as possible, that specification will likely change a lot until then. I'll close this PR for now.

PR #5943, opened 2026-03-04
Fix(agent): Improve tunnel preservation during agent upgrade when using tunnel. Housekeeping of shellhub-agent files in own directory

ltan10:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] New Feature
- [x] Feature Improvment
- [ ] Refactoring
- [ ] Documentation
- [x] Other, please describe:

### Description:

* Organized go based agent config and key files into own dedicated directory in `/etc/shellhub-agent/`
  * Does not remove existing `/etc/shellhub.key` and `/etc/shellhub-agent.env` or `/opt/shellhub/`
  * Best migration procedure would be to copy `shellhub.key` to `/etc/shellhub-agent/shellhub.key` prior to performing installation/upgrade
* Modified default PRIVATE_KEY location value for podman, snap and docker installation in shell install script to `/etc/shellhub-agent`
* ***Restarts shellhub-agent service at end of installation step instead of ending shellhub-agent at start of installation procedure.***
  * ***Reduces risk of removing device from shellhub tunnel network if agent installation/re-installation/upgrade process was performed over shellhub tunnel***

### Migration Guide \**Optional\**

The following guide also applies when migrating from `runc-shellhub-agent` to the native `go-shellhub-agent`.

The following steps are only needed if you want to retain the same device unique ID in ShellHub and avoid creating a new pending request.

If the migration steps are not performed, the device will be registered as a new device, and the pending request will need to be accepted.

1. Manually create `/etc/shellhub-agent/` directory
2. Copy `shellhub.key` from `/opt/shellhub-agent/shelhlub.key` or `/etc/shellhub.key` to `/etc/shellhub-agent/shellhub.key`
3. Perform agent installation/upgrade steps as normal
4. Ensure installation/upgrade was successful and tunnel is active 
5. Able to remove the following `/opt/shellhub-agent/` and `/etc/shellhub.key` and `/etc/shellhub-agent.env`

### DIFF
--- a/agent/installer.go
+++ b/agent/installer.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	agentEnvFile     = "/etc/shellhub-agent.env"
+	agentEnvFile     = "/etc/shellhub-agent/shellhub-agent.env"
 	agentServiceFile = "/etc/systemd/system/shellhub-agent.service"
 	agentServiceName = "shellhub-agent"
 )
@@ -26,7 +26,7 @@ Wants=network-online.target
 Requires=local-fs.target
 
 [Service]
-EnvironmentFile=/etc/shellhub-agent.env
+EnvironmentFile=/etc/shellhub-agent/shellhub-agent.env
 ExecStart={{.BinaryPath}}
 Restart=on-failure
 RestartSec=5
@@ -83,12 +83,12 @@ func registerInstallerCommands(rootCmd *cobra.Command) {
 
 	installCmd.Flags().String("server-address", "", "ShellHub server address")
 	installCmd.Flags().String("tenant-id", "", "Namespace tenant ID")
-	installCmd.Flags().String("private-key", "/etc/shellhub.key", "Path to the agent private key file")
+	installCmd.Flags().String("private-key", "/etc/shellhub-agent/shellhub.key", "Path to the agent private key file")
 	installCmd.Flags().String("preferred-hostname", "", "Preferred device hostname")
 	installCmd.Flags().String("preferred-identity", "", "Preferred device identity")
 	installCmd.Flags().Uint("keepalive-interval", 30, "Keepalive interval in seconds")
 	installCmd.MarkFlagRequired("server-address") //nolint:errcheck
-	installCmd.MarkFlagRequired("tenant-id")       //nolint:errcheck
+	installCmd.MarkFlagRequired("tenant-id")      //nolint:errcheck
 
 	rootCmd.AddCommand(installCmd)
 
@@ -129,6 +129,10 @@ func agentInstall(cfg installerConfig) error {
 	binaryPath, err := filepath.EvalSymlinks(exe)
 	if err != nil {
 		return fmt.Errorf("failed to resolve symlinks: %w", err)
+	}
+
+	if err := os.MkdirAll(filepath.Dir(agentEnvFile), 0755); err != nil {
+		return fmt.Errorf("failed to create directory %s: %w", filepath.Dir(agentEnvFile), err)
 	}
 
 	if err := writeAgentEnvFile(cfg); err != nil {
--- a/agent/installer.go
+++ b/agent/installer.go
@@ -117,9 +117,9 @@ func agentInstall(cfg installerConfig) error {
 		return fmt.Errorf("systemd is not available on this system")
 	}
 
-	// Stop existing service before overwriting files (re-install / upgrade).
-	// Ignore error — service may not exist yet.
-	exec.Command("systemctl", "disable", "--now", agentServiceName).Run() //nolint:errcheck
+	// Best practice would be to disable service here before install/upgrade
+	// If upgrade performed over tunnel, ssh session disconnect, binary gets sighup.
+	// Service will restart at end of install/upgrade
 
 	exe, err := os.Executable()
 	if err != nil {
@@ -147,8 +147,14 @@ func agentInstall(cfg installerConfig) error {
 		return fmt.Errorf("failed to reload systemd daemon: %w", err)
 	}
 
-	if err := exec.Command("systemctl", "enable", "--now", agentServiceName).Run(); err != nil {
+	// For upgrade over tunnel support, just enable service, service reboot later
+	if err := exec.Command("systemctl", "enable", agentServiceName).Run(); err != nil {
 		return fmt.Errorf("failed to enable service: %w", err)
+	}
+
+	// Restarts service to upgraded binary, session dies but tunnel remains active.
+	if err := exec.Command("systemctl", "restart", agentServiceName).Run(); err != nil {
+		return fmt.Errorf("failed to restart service: %w", err)
 	}
 
 	return nil
--- a/agent/installer.go
+++ b/agent/installer.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	agentEnvFile     = "/etc/shellhub-agent.env"
+	agentEnvFile     = "/etc/shellhub-agent/shellhub-agent.env"
 	agentServiceFile = "/etc/systemd/system/shellhub-agent.service"
 	agentServiceName = "shellhub-agent"
 )
@@ -27,7 +27,7 @@ Wants=network-online.target
 Requires=local-fs.target
 
 [Service]
-EnvironmentFile=/etc/shellhub-agent.env
+EnvironmentFile=/etc/shellhub-agent/shellhub-agent.env
 ExecStart={{.BinaryPath}}
 Restart=on-failure
 RestartSec=5
@@ -94,7 +94,11 @@ func registerInstallerCommands(rootCmd *cobra.Command) {
 
 	installCmd.Flags().String("server-address", "", "ShellHub server address")
 	installCmd.Flags().String("tenant-id", "", "Namespace tenant ID")
-	installCmd.Flags().String("private-key", "/etc/shellhub.key", "Path to the agent private key file")
+	installCmd.Flags().String(
+		"private-key",
+		"/etc/shellhub-agent/shellhub.key",
+		"Path to the agent private key file",
+	)
 	installCmd.Flags().String("install-key", "", "Install key used to enroll the device")
 	installCmd.Flags().String("preferred-hostname", "", "Preferred device hostname")
 	installCmd.Flags().String("preferred-identity", "", "Preferred device identity")
@@ -139,6 +143,10 @@ func agentInstall(cfg installerConfig) error {
 	binaryPath, err := filepath.EvalSymlinks(exe)
 	if err != nil {
 		return fmt.Errorf("failed to resolve symlinks: %w", err)
+	}
+
+	if err := os.MkdirAll(filepath.Dir(agentEnvFile), 0755); err != nil {
+		return fmt.Errorf("failed to create directory %s: %w", filepath.Dir(agentEnvFile), err)
 	}
 
 	if err := writeAgentEnvFile(cfg); err != nil {
--- a/agent/installer.go
+++ b/agent/installer.go
@@ -133,7 +133,10 @@ func agentInstall(cfg installerConfig) error {
 		return fmt.Errorf("systemd is not available on this system")
 	}
 
-	exec.Command("systemctl", "disable", "--now", agentServiceName).Run() //nolint:errcheck
+	// Do not disable/stop the service here. During an upgrade over the SSH tunnel,
+	// stopping the service disconnects the session and causes the agent to receive
+	// SIGHUP, which can cause the installation to fail. The service is restarted
+	// at the end of the install/upgrade procedure.
 
 	exe, err := os.Executable()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


otavio (commented):
The biggest problem I foresee here is migrating the field-deployed agents. 

So even though I agree with the idea of this change, we need to consider how we will migrate the existing agents to work with this new approach or provide a backward-compatible way for them to keep working. 

How do you foresee solving this issue?

ltan10:
@shellhub-io/admin

ltan10:
Rebased branch. Proposed change completed.

@otavio any chance PR to be merged for next release?

otavio:
@ltan10 There is still a conflict that needs to be resolved. Can you take a look?

Besides this, I am wondering what tests of the upgrade you have done?

ltan10:
@otavio rebased to the master branch to resolve the conflicts. I've also updated the comments to make more sense.

As for the testing, I have only performed the `STANDALONE` installation method, in both an upgrade and fresh installation capacity. There is no issues with a fresh agent install.

For an upgrade context, I have tested the upgrade over an active remote ShellHub SSH tunnel multiple times, and was successful each of those times. However like with every remote configuration there is a chance that it may fail.

Every upgrade for runc-agent and go-agent always rewrites it's shellhub-agent.env as is normal behavior.

Observations for the tests are mentioned in previous comments. I'll try to summarize them here.

When upgrading from the runc-agent to "ungrouped" go-agent or "ungrouped" go-agent to this "grouped" version, the existing private key is not migrated. As a result, the new agent will generate a new key, and the device will be treated as a new device registration and await acceptance on the dashboard. Subsequently any existing `shellhub.key` and `shellhub-agent.env` from the existing install needs to be manually cleaned.

A migration guide is included in the PR description with steps to preserve the existing key and avoid re-registration.

I have NOT tested the `containerized` install/upgrade of this new agent. It would be great if someone on the team is able to verify their behavior.

agent/installer.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	agentEnvFile     = "/etc/shellhub-agent.env"
+	agentEnvFile     = "/etc/shellhub-agent/shellhub-agent.env"
```
otavio, line 18: `agentEnvFile` now points at `/etc/shellhub-agent/shellhub-agent.env`, but `agent/pkg/agentd/envfile.go:11` still has:

```go
const defaultEnvFilePath = "/etc/shellhub-agent.env"
```

That is the only production caller (`agent/pkg/agentd/agent.go:171` via `LoadConfigFromEnv`), so the agent runtime never reads the file this installer writes — only systemd does, through `EnvironmentFile=`. Result: `shellhub-agent login` and any manual invocation break on a fresh install, and on an upgrade the stale `/etc/shellhub-agent.env` keeps pointing at the old key path.

More generally, the layout is now described by two hardcoded strings in two packages. A single seam in `agent/pkg/agentd` (`ConfigDir()` / `EnvFilePath()` / `PrivateKeyPath()`) consumed by both the installer and the runtime would have made this class of drift impossible — and there are several more instances of it in `install.sh` below.

agent/installer.go
```diff
@@ -94,7 +94,11 @@ func registerInstallerCommands(rootCmd *cobra.Command) {
 
 	installCmd.Flags().String("server-address", "", "ShellHub server address")
 	installCmd.Flags().String("tenant-id", "", "Namespace tenant ID")
-	installCmd.Flags().String("private-key", "/etc/shellhub.key", "Path to the agent private key file")
+	installCmd.Flags().String(
+		"private-key",
+		"/etc/shellhub-agent/shellhub.key",
+		"Path to the agent private key file",
+	)
```
otavio, line 101: Changing the default is the easy half. There is no code that looks at the legacy `/etc/shellhub.key` (or `/opt/shellhub/shellhub.key`), so an existing device upgraded without the manual steps from the PR description generates a fresh key here and re-enrols as a new pending device.

Please add a fallback: if the new path does not exist and a legacy key does, use (or migrate) the legacy one. The prose guide should be a convenience, not a prerequisite for keeping a device's identity.

agent/installer.go
```diff
@@ -129,7 +133,10 @@ func agentInstall(cfg installerConfig) error {
 		return fmt.Errorf("systemd is not available on this system")
 	}
 
-	exec.Command("systemctl", "disable", "--now", agentServiceName).Run() //nolint:errcheck
+	// Do not disable/stop the service here. During an upgrade over the SSH tunnel,
+	// stopping the service disconnects the session and causes the agent to receive
+	// SIGHUP, which can cause the installation to fail. The service is restarted
+	// at the end of the install/upgrade procedure.
```
otavio, line 139: Our code style allows exactly two kinds of comment: a doc comment on an exported declaration, and a suppression's reason. This is neither — and it's the hardest case of the rule, because it documents code that is no longer here, so nothing in the file is named by it.

The content is genuinely valuable, though: "stopping the service kills the SSH session, the agent takes SIGHUP and the install dies" is exactly the kind of reason that has nowhere in the code to live. Please move it verbatim into the commit body of d5427ac, where `git blame` reaches it from the line that raised the question. Right now the commit says only "Remove pre-install service disable step", so deleting this comment loses the reason entirely.

agent/installer.go
```diff
@@ -141,6 +148,10 @@ func agentInstall(cfg installerConfig) error {
 		return fmt.Errorf("failed to resolve symlinks: %w", err)
 	}
 
+	if err := os.MkdirAll(filepath.Dir(agentEnvFile), 0755); err != nil {
```
otavio, line 151: Two things on this line.

**Mode.** This directory holds the agent private key and a `0600` env file carrying `SHELLHUB_INSTALL_KEY` and the tenant ID. `0755` leaves it world-traversable and makes the per-file bits the only thing protecting those secrets. `0700` is the right mode for a root-owned secret directory.

**Style.** The rest of this file uses Go octal literals (`0o600` at line 207, `0o644` at line 217); please match.

```suggestion
	if err := os.MkdirAll(filepath.Dir(agentEnvFile), 0o700); err != nil {
```

agent/installer.go
```diff
@@ -153,10 +164,17 @@ func agentInstall(cfg installerConfig) error {
 		return fmt.Errorf("failed to reload systemd daemon: %w", err)
 	}
 
-	if err := exec.Command("systemctl", "enable", "--now", agentServiceName).Run(); err != nil {
+	// NOTE: Only enable the service here to maintain the SSH tunnel if one is active.
+	// The service will be restarted after the upgrade.
```
otavio, line 168: Same as above — not a doc comment on an exported symbol and not a suppression reason, so it doesn't survive our comment rule. The `enable`-without-`--now` decision and its pairing with the `restart` below belongs in the commit message.

agent/installer.go
```diff
@@ -153,10 +164,17 @@ func agentInstall(cfg installerConfig) error {
...
+	if err := exec.Command("systemctl", "enable", agentServiceName).Run(); err != nil {
 		return fmt.Errorf("failed to enable service: %w", err)
 	}
 
+	// Finally, restart the service with the newly installed version.
+	if err := exec.Command("systemctl", "restart", agentServiceName).Run(); err != nil {
+		return fmt.Errorf("failed to restart service: %w", err)
+	}
```
otavio, line 176: The comment goes the way of the other two.

On the code: this restart is also a failure seam worth handling. `standalone_install()` in `install.sh` reacts to a non-zero exit from `install` by running `$SUDO rm -f "$INSTALL_BIN"` (line 428) — but by this point the unit has already been `enable`d and its service file written. A restart failure therefore leaves an enabled unit whose `ExecStart` points at a binary that was just deleted, and the next boot fails silently.

Either disable the unit on the error path here, or don't enable until the restart has succeeded.

Also worth capturing stderr from these `exec.Command` calls (`CombinedOutput`) — as written, a failure reaches the user as a bare "failed to restart service" with nothing to act on.

install.sh
```diff
@@ -206,7 +206,7 @@ podman_install() {
   esac
 
   if [ -z "$MODE" ]; then
-    ARGS="$ARGS -e SHELLHUB_PRIVATE_KEY=${PRIVATE_KEY:-/host/etc/shellhub.key}"
+    ARGS="$ARGS -e SHELLHUB_PRIVATE_KEY=${PRIVATE_KEY:-/host/etc/shellhub-agent/shellhub.key}"
```
otavio, line 209: This default moved, but the matching one 35 lines down in the same function did not — line 244 still reads:

```sh
_CKEY="${PRIVATE_KEY:-/host/etc/shellhub.key}"
enroll_agent_interactively "$WRAPPER_PATH" "${_CKEY#/host}"
```

That is the path `enroll_agent_interactively` **waits on**. With no explicit `PRIVATE_KEY`, the container now writes the key to the new location while the wait loop at line 123 polls the old one, times out after 30s, and reintroduces the login-vs-daemon key race that loop exists to prevent.

(GitHub won't let me comment on line 244 — it isn't in the diff.)

install.sh
```diff
@@ -294,7 +294,7 @@ docker_install() {
...
```
otavio, line 297: Same problem as the podman path: line 333 still has `_CKEY="${PRIVATE_KEY:-/host/etc/shellhub.key}"`, so the enrollment wait polls the old location while the container writes the new one.

(Line 333 isn't in the diff, so I can't comment on it directly.)

install.sh
```diff
@@ -378,7 +378,7 @@ snap_install() {
 
     sudo snap set shellhub server-address="$SERVER_ADDRESS"
     sudo snap set shellhub tenant-id="$TENANT_ID"
-    sudo snap set shellhub private-key="${PRIVATE_KEY:-/etc/shellhub.key}"
+    sudo snap set shellhub private-key="${PRIVATE_KEY:-/etc/shellhub-agent/shellhub.key}"
```
otavio, line 381: This one is correct, but its sibling in `standalone_install()` was missed — line 437 still reads:

```sh
enroll_agent_interactively "$SUDO $INSTALL_BIN" "${PRIVATE_KEY:-/etc/shellhub.key}"
```

So the native installer writes the key to `/etc/shellhub-agent/shellhub.key` (per the new flag default) while enrollment waits on `/etc/shellhub.key`.

That is 3 of 6 defaults left stale across this file. Rather than chasing each literal, consider a single variable near the top — `DEFAULT_PRIVATE_KEY=/etc/shellhub-agent/shellhub.key` — that every method derives from.

(Line 437 isn't in the diff, hence the comment here.)

otavio (changes requested):
Thanks for digging into this, and especially for the `systemctl restart`-at-the-end finding — the SIGHUP-over-the-tunnel diagnosis in the thread is correct and that part of the change is the right shape.

I've left inline comments on the specific lines. Below are the items that have no line in the diff to hang off.

### 1. No code-level migration path (blocking)

This is the point I raised earlier in the thread, and it is still open. The PR gives a migration path in prose only: the user must hand-create `/etc/shellhub-agent/` and copy the key before upgrading. Every field-deployed agent whose operator doesn't read the release notes re-enrols as a new pending device, leaving the old accepted device behind with a clashing hostname to be deleted by hand.

We can't ship a silent identity change for existing installs. The agent should fall back to the legacy location when the new one is absent — read `/etc/shellhub.key` (and `/opt/shellhub/shellhub.key`) if `/etc/shellhub-agent/shellhub.key` does not exist, and ideally migrate it on first run. Same for the env file. That turns the migration guide into a safety net rather than a prerequisite.

### 2. `agent/pkg/agentd/envfile.go:11` still points at the old path

```go
const defaultEnvFilePath = "/etc/shellhub-agent.env"
```

This is the only production caller (`agent/pkg/agentd/agent.go:171`, via `LoadConfigFromEnv`), so the runtime never reads the new env file except through systemd's `EnvironmentFile=`. Two consequences:

- **Regression on a fresh install:** `shellhub-agent login` and any manual non-systemd invocation now fail — `LoadConfigFromEnv` requires `SHELLHUB_PRIVATE_KEY`/`SHELLHUB_SERVER_ADDRESS` and the fallback file it was given in 6f57bd00b no longer exists at that path.
- **Split brain on upgrade:** systemd reads the new file, while the stale `/etc/shellhub-agent.env` survives and is what hand-run invocations read — pointing at the *old* key path.

### 3. Docs still document the old paths

`ui/apps/docs/src/pages/get-started/install/{standalone,docker,podman,raspberry-pi}.mdx` still reference `/etc/shellhub.key`.

### 4. The new defaults have no test coverage

`install.bats` passes (82/82), but only because it always exports `PRIVATE_KEY` (line 20) — it never exercises the defaults, in either direction. That is why the three stale defaults flagged inline slipped through. Worth a case that asserts the default path per install method.

### 5. Commit messages

Per our code style, the reason behind a workaround lives in the commit message, not in a comment — see the inline notes. Concretely:

- `0aa4b7b` ("Create directory if doesn't exist.") gives no rationale at all for the path move: no migration story, no reason for the directory mode. Its `chore(agent)` scope also mislabels changes to `install.sh` and `agent/packaging/`.
- `d5427ac` says "Remove pre-install service disable step" but omits *why* — the SIGHUP mechanism currently survives only in the banned comment. That explanation is the valuable part; please move it into the commit body in full.

Happy to re-review once the migration fallback and the stale paths are sorted.

PR #5947, opened 2026-03-05
fix(ssh): add WebSocket ping/pong to per-session revdial connections

### LINKED ISSUE #5946: "Idle" SSH connections are forcefully disconnected
### Discussed in https://github.com/shellhub-io/shellhub/discussions/5945

<div type='discussions-op-text'>

<sup>Originally posted by **zoopp** March  5, 2026</sup>
I'm running into an interesting issue with my setup. I host ShellHub CE on GCP behind an [External Proxy Network Load Balancer](https://docs.cloud.google.com/load-balancing/docs/tcp), essentially with an architecture like this:
image

The problem I keep encountering is that idle SSH connections get forcibly disconnected after a couple of minutes of inactivity:
image

This happens both when connecting through ShellHub and when using native ssh directly. Interestingly, the disconnect does not occur if there is periodic activity on the connection. For example, running something like `watch <cmd>` (which produces output every couple of seconds) keeps the connection alive indefinitely.

However, I expected that SSH keep alive settings might mitigate this, at least for native SSH connections. I've experimented with options like:
- TCPKeepAlive
- ServerAliveInterval
- ServerAliveCountMax

... but the connection still gets dropped after a few minutes of inactivity.

Does anyone have suggestions on what could be done to prevent this, at least for native SSH connections, without changing the overall GCP setup?</div>

otavio:
## Summary

Fix idle SSH session disconnections caused by intermediaries (load balancers, NAT, reverse proxies) closing silent WebSocket connections. This PR addresses #5946 with per-session WebSocket keepalive and race condition fixes in the WebSocket adapter.

### Per-session WebSocket ping/pong (`pkg/revdial`)

The V1 (revdial) per-session WebSocket connections had no server→agent keepalive. During idle SSH sessions, the only traffic was the agent's SSH keepalive (one-way, every `KEEPALIVE_INTERVAL` seconds). The server→agent direction was silent, causing intermediaries to detect a half-idle connection and close it.

Now `Ping()` is called on the agent-side adapter in `grabConn()`, sending WebSocket ping frames every 30s. The server automatically responds with pongs (gorilla/websocket handles this), creating bidirectional traffic that keeps the connection alive through all intermediaries.

### WebSocket adapter `Close()` race fix (`pkg/wsconnadapter/wsconnadapter.go`)

The old `Close()` had a race condition: concurrent callers (e.g., pong timeout `AfterFunc` firing while normal teardown runs) could panic by sending to an already-closed channel, or call `conn.Close()` multiple times. Now `sync.Once` ensures the entire cleanup — stopping the ping goroutine and closing the WebSocket connection — happens exactly once.

Subsequent `Close()` calls return the stored error from the first call rather than nil, preserving error semantics for callers that check the return value.

### WebSocket adapter `Ping()` init race fix (`pkg/wsconnadapter/wsconnadapter.go`)

The old `Ping()` guarded against re-initialization with a nil check (`if a.pongCh != nil`), which is racy under concurrent calls — both callers see nil, both create channels and goroutines, leaking the first set. Now `sync.Once` guarantees initialization happens exactly once.

## Community testing

| Tester | Setup | Result |
|--------|-------|--------|
| @zoopp | GCP Load Balancer, agent v0.21.4 | 4 sessions idle 30+ min — no disconnections |
| @ltan10 | External nginx proxy, self-compiled agent | 600 min session — no disconnections |

Note: @ltan10 also reported a **separate issue** with the main control connection dying at 20-27 min intervals. This is unrelated to per-session keepalive — the main connection already has four independent keepalive sources at 30s. Their issue points to external proxy behavior (max connection duration, connection pool limits).

## Test plan

- [x] `go build` — all services compile
- [x] Integration tests pass in CI (both postgres and mongo)
- [x] @zoopp: tested, sessions stay alive 30+ minutes
- [x] @ltan10: tested, session stayed alive 600 minutes
- [x] Final re-test with updated branch (nginx change removed)

Fixes #5946

### DIFF
--- a/api/services/session.go
+++ b/api/services/session.go
@@ -82,7 +82,27 @@ func (s *service) KeepAliveSession(ctx context.Context, uid models.UID) error {
 
 	session.LastSeen = clock.Now()
 
-	return s.store.SessionUpdate(ctx, session)
+	if err := s.store.SessionUpdate(ctx, session); err != nil {
+		return err
+	}
+
+	activeSession, err := s.store.ActiveSessionResolve(ctx, store.SessionUIDResolver, string(uid))
+	if err != nil {
+		// Active session was TTL-deleted; recreate it.
+		if err := s.store.ActiveSessionCreate(ctx, session); err != nil {
+			log.WithError(err).WithField("session_uid", uid).Warn("failed to recreate active session")
+		}
+
+		return nil
+	}
+
+	activeSession.LastSeen = session.LastSeen
+
+	if err := s.store.ActiveSessionUpdate(ctx, activeSession); err != nil {
+		log.WithError(err).WithField("session_uid", uid).Warn("failed to update active session's last seen")
+	}
+
+	return nil
 }
 
 func (s *service) UpdateSession(ctx context.Context, uid models.UID, model models.SessionUpdate) error {
--- a/api/services/session_test.go
+++ b/api/services/session_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"testing"
+	"time"
 
 	goerrors "errors"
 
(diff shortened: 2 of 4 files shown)


otavio:
@gustavosbarreto Good call — I've moved `Ping()` to `grabConn()` on the agent side as you suggested. This distributes the goroutine cost across agents instead of concentrating it on the server.

This fix works for the immediate issue, but we'll need to think more deeply about how we want to implement this long-term. Given that we've never had agent integration in other languages, we need to find a way to enable this back-and-forth communication that accounts for all clients. Some of our customers rely on a large number of devices, and in those cases this extra WebSocket ping/pong transmission on every active session could impact performance.

PR #6043, opened 2026-03-17
feat(cli): add user list command

geovannewashington:
## What changed

Add `cli user list` command to list users. The output is displayed
as a table with the columns USERNAME, EMAIL, and ROLE.

Also redirect the `.env.override` loading message in `bin/utils`
to stderr so it does not interfere with piped output.

## Why

Allow users to be listed directly from the CLI and ensure the
command output behaves correctly when used with pipes.

## How to test
1. Start the environment:
   `make start`

2. Run:
   `./bin/cli user list`

3. Verify users are displayed in a table.

4. Test piping:
   `./bin/cli user list | head -n 2`

The output should contain the table header and the first user.

Closes shellhub-io/team#92

### DIFF
--- a/cli/cmd/helpers.go
+++ b/cli/cmd/helpers.go
@@ -1,0 +1,23 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"text/tabwriter"
+
+	"github.com/shellhub-io/shellhub/pkg/models"
+)
+
+// helper to print the listing table
+func printUsersTable(out io.Writer, users []models.User) {
+	w := tabwriter.NewWriter(out, 0, 0, 2, ' ', 0)
+	fmt.Fprintln(w, "USERNAME\tEMAIL\tROLE")
+	for _, u := range users {
+		role := "user"
+		if u.Admin {
+			role = "admin"
+		}
+		fmt.Fprintf(w, "%s\t%s\t%s\n", u.Username, u.Email, role)
+	}
+	w.Flush()
+}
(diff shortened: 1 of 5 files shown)


ltan10:
@geovannewashington don't know what to say, i've tried all the steps from 1. - 7.
My terminal was using xterm, did try changing to xterm-256color. But alas still seeing leading spaces.

1. No hidden characters visible, only `$` line termination observed `USERNAME  EMAIL                     ROLE$`
2. Yes your logic was correctly detecting pipe and triggering `-T` flag for raw output.
3. No change
4. Unfortunately still observing same behavior
5. No carriage return observed, weirdly enough we do not see the hex representation of the leading spaces, but the hex dump themselves have the leading spaces.
```
0000000 55 53 45 52 4e 41 4d 45 20 20 45 4d 41 49 4c 20
                                                       0000020 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20
                                                                                                              0000040 20 20 20 20 52 4f 4c 45 0a 61 64 6d 69 6e 20 20
                                                                                                                                                                     0000060 20 20 20 6c
```
6. No change
7. Together with 6. piping to `cat -A` all it did was add extra spaces before line termination. But leading spaces still present.

Further test tried:

8. Tried flushing after every single print. no change.
9. Using unedited `user.go`. When piping to file or when piping to `tail` first, leading spaces is not observed.
10. I tried bypassing tabwriter and using fixed delimiters, `fmt.Printf("%-15s %-25s %-10s\n", u.Username, u.Email, role)`,  still observed same leading spaces.
11. Tried `./bin/cli` with and without the `TTY_FLAG` `-T` no differences observed.

PR #6248, opened 2026-04-28
feat: add device custom fields

danielgatis:
## Motivation

As device fleets grow, operators often need to attach context-specific metadata to individual devices — things like physical location, environment tags, asset identifiers, or team ownership. Without a dedicated mechanism, this information ends up scattered across external spreadsheets or encoded into hostnames, making it hard to query and maintain.

This feature introduces **custom fields**: a simple key-value map attached to each device. Operators can freely define any metadata they need, and that metadata becomes a first-class citizen in the dashboard — visible in the device list, searchable alongside the hostname, and fully manageable from the device detail page without touching the API directly.

https://github.com/user-attachments/assets/32f357be-424d-4068-ab8a-62263d059343

---

## Summary

- Add `custom_fields map[string]string` to the Device model, stored as JSONB in PostgreSQL and an embedded document in MongoDB
- Expose via the existing `PUT /devices/{uid}` endpoint — nil means no change, empty map clears all fields
- Search in the device list also queries custom field values (OR with hostname search)
- Display and manage custom fields in the device detail page (add, delete with confirmation)
- Add custom fields column in the devices list table

## Changes

| Scope | Description |
|---|---|
| `pkg` | Add `CustomFields` to `Device` model and `DeviceUpdate` request DTO |
| `api` | Service logic, MongoDB `$objectToArray` filter, PostgreSQL JSONB filter, migration 002 |
| `openapi` | Add `custom_fields` to device schema and PUT request body |
| `docker` | Add `binutils-gold` to Alpine images for ARM64 linker support (see note below) |
| `ui` (Vue) | Display custom fields on device detail page |
| `ui-react` | Custom fields column in list, management UI in detail page, search integration |
| `test(api)` | Unit tests for PG/Mongo filters and service UpdateDevice cases |
| `test(ui-react)` | Tests for devices list and device details custom fields behavior |

### Note: binutils-gold in Alpine Dockerfiles

During local development on an ARM64 machine (Apple Silicon), the Docker builds were failing with:

```
/usr/lib/gcc/x86_64-alpine-linux-musl/14.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find 'ld'
```

The root cause is that `golangci-lint` is compiled with `-fuse-ld=gold`, which requires the `gold` linker (`binutils-gold` package) to be present on the system. Alpine's default `binutils` does not include it. Adding `binutils-gold` to the Alpine images that install `golangci-lint` resolves the build failure on ARM64 hosts.

## Test plan

- [x] `go test ./api/services/... ./api/store/pg/internal/... ./api/store/mongo/internal/...`
- [x] `cd ui-react/apps/console && npx vitest run src/pages/devices/__tests__`
- [x] `PUT /devices/{uid}` with `{"name": "x", "custom_fields": {"env": "prod"}}` → 200
- [x] `GET /devices/{uid}` → response includes `custom_fields`
- [x] Search on `/devices` by custom field value returns matching devices
- [x] Device detail page shows custom fields with add/delete UI

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -58,7 +58,7 @@ FROM base AS development
 ARG GOPROXY
 ENV GOPROXY ${GOPROXY}
 
-RUN apk add --update openssl openssh-client util-linux setpriv
+RUN apk add --update openssl build-base binutils-gold openssh-client util-linux setpriv
 RUN go install github.com/air-verse/air@v1.62 && \
     go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
 
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -75,7 +75,7 @@ ARG GOPROXY
 ARG MJML_VERSION
 ENV GOPROXY=${GOPROXY}
 
-RUN apk add --update openssl build-base docker-cli npm
+RUN apk add --update openssl build-base binutils-gold docker-cli npm
 RUN npm install -g mjml@${MJML_VERSION}
 RUN go install github.com/air-verse/air@v1.62 && \
     go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3 && \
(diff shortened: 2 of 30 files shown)


danielgatis:
@gustavosbarreto 

When you have some free time, please consider doing another review.

danielgatis:
@gustavosbarreto 

🎉 looks like the PR is good now

Any chance to get it released in earlier future?

gustavosbarreto:
Thanks! It'll go out in the next release.

PR #6250, opened 2026-04-28
fix(ui-react): allow sidebar to collapse and expand on desktop

luannmoreira:
## What

Restored the ability to collapse and expand the sidebar on desktop/tablet screen sizes. The toggle button in the sidebar footer now works on all viewport sizes, not just mobile.

## Why

A previous refactor (`b726d1ec`) simplified `useSidebarLayout` by hardcoding `isOpen = isDesktop`, removing the desktop expand/collapse state and never passing `onToggle` to the desktop sidebar instances. This left the chevron button wired up in `SidebarShell` but unreachable.

## Changes

- **`useSidebarLayout`**: Added `desktopExpanded` state (defaults `true`) and a `toggleDesktop` handler. `isOpen` on desktop now reflects user preference instead of being hardcoded to `true`.
- **`AppLayout` / `AdminLayout`**: Pass `onToggle={handlers.toggleDesktop}` to the desktop sidebar, activating the footer button.
- **`SidebarShell` footer**: Fixed a layout bug where the invisible `<p>` label (opacity-0, whitespace-nowrap) still occupied flex space, pushing the chevron button outside the 60px collapsed sidebar and clipping it under `overflow-hidden`. Replaced the opacity trick with conditional rendering and switched the footer to `justify-center` when collapsed so the chevron is properly centered.

### DIFF
--- a/ui-react/apps/console/src/components/layout/AdminLayout.tsx
+++ b/ui-react/apps/console/src/components/layout/AdminLayout.tsx
@@ -12,7 +12,7 @@ export default function AdminLayout() {
     <div className="flex flex-col h-screen bg-background">
       <div className="flex flex-1 min-h-0 overflow-hidden">
         {isDesktop ? (
-          <AdminSidebar expanded={isOpen} />
+          <AdminSidebar expanded={isOpen} onToggle={handlers.toggleDesktop} />
         ) : (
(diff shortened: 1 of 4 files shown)


luannmoreira:
Reverting — will revisit the approach.

PR #6283, opened 2026-05-07
fix(api): propagate web endpoints envs to api container

luizhf42:
## What
Pass `SHELLHUB_WEB_ENDPOINTS` and `SHELLHUB_WEB_ENDPOINTS_DOMAIN` to the `api` service in `docker-compose.yml`.

## Why
The cloud init reads `SHELLHUB_WEB_ENDPOINTS` (see `cloud/pkg/envs/envs.go` and `cloud/internal/cloud/init.go`) to decide whether to call `handler.WithWebEndpoints(true)`, which is what registers the web-endpoints routes (`cloud/internal/cloud/routes/routes.go`). The variable was already wired into `ssh`, `ui`, `ui-react`, and `gateway`, but never reached `api` — so the flag stayed false in the API container, the routes were never registered, and every web-endpoints request returned 404.

## Changes
- **docker-compose.yml**: add `SHELLHUB_WEB_ENDPOINTS` and `SHELLHUB_WEB_ENDPOINTS_DOMAIN` to the `api` service environment, matching how the other services already consume them.

## Testing
- With `SHELLHUB_WEB_ENDPOINTS=true` in `.env.override`, restart the `api` container and confirm the web-endpoints routes (e.g. `GET /api/namespaces/:tenant/devices/:uid/webendpoints`) respond instead of 404.
- With the var unset/false, behavior is unchanged from before.


gustavosbarreto:
LGTM. Technically `SHELLHUB_WEB_ENDPOINTS*` is a cloud/enterprise feature and the right place for these envs would be the `cloud/docker-compose.enterprise.dev.yml` overlay, but since the CE `docker-compose.yml` already passes them through to `ssh`, `ui`, `ui-react` and `gateway`, it makes sense to keep the existing pattern for now. I'll open a follow-up issue on the cloud repo so we can migrate everything at once later.

PR #6292, opened 2026-05-08
fix(dev): replace binutils-gold with CGO_ENABLED=0 in dev images

### LINKED ISSUE #6285: dev: replace deprecated binutils-gold with CGO_ENABLED=0 in dev images
## Context

PR #6248 added `binutils-gold` to 6 development Dockerfiles to fix
ARM64 (Apple Silicon) builds failing with `cannot find 'ld'` during
`go install` of dev tools (air, golangci-lint, mockery).

The fix works but has issues:

1. **Wrong root cause**: the PR attributed `-fuse-ld=gold` to
   golangci-lint, but it actually comes from Go's own toolchain
   (`/usr/local/go/pkg/tool/linux_arm64/link`) when CGO is enabled
   on linux/arm64. Verified by reproducing in
   `golang:1.25.8-alpine3.22` ARM64 container.

2. **gold is deprecated**: removed/marked deprecated in binutils 2.44
   (Jan 2025). Alpine will drop the package eventually.

3. **Bloat for x86_64 users**: the package is only needed on ARM64,
   but was added unconditionally to all dev images.

4. **CGO not actually needed**: the tools installed (air, golangci-lint,
   mockery) are pure Go and don't require cgo.

## Reproduction

```sh
# Fails on ARM64 without binutils-gold
docker run --rm --platform linux/arm64 golang:1.25.8-alpine3.22 sh -c '
  apk add --no-cache git build-base
  CGO_ENABLED=1 go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
'
# → /usr/local/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
# → collect2: fatal error: cannot find 'ld'

# Works on ARM64 without binutils-gold
docker run --rm --platform linux/arm64 golang:1.25.8-alpine3.22 sh -c '
  apk add --no-cache git
  CGO_ENABLED=0 go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
'
# → success, binary works
```

## Proposed fix

Replace `binutils-gold` install + default cgo with `CGO_ENABLED=0`
on all `go install` of dev tools, in 6 Dockerfiles:

```dockerfile
# Remove binutils-gold (and build-base where it was added only for this) from apk add
RUN apk add --update openssl ...

# Add CGO_ENABLED=0 to each go install
RUN CGO_ENABLED=0 go install github.com/air-verse/air@v1.62 && \
    CGO_ENABLED=0 go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3 && \
    CGO_ENABLED=0 go install github.com/vektra/mockery/v2/...@v2.53.2
```

Files affected:
- `agent/Dockerfile`
- `api/Dockerfile`
- `cli/Dockerfile`
- `gateway/Dockerfile`
- `openapi/Dockerfile`
- `ssh/Dockerfile`

## Future-proofing note

If we ever need cgo in dev tools or in the actual app build, the
modern replacement for gold is `ld.lld` (LLVM linker), available
via `apk add lld`. To force Go to use it on ARM64 alpine:
`ENV CGO_LDFLAGS="-fuse-ld=lld"`.

gustavosbarreto:
Closes #6285.

## Summary
- `binutils-gold` was added in #6248 to fix `cannot find 'ld'` on linux/arm64 alpine when installing dev tools (air, golangci-lint, mockery).
- The real trigger is Go's `net`/`os/user` compiling with cgo by default, not the tools themselves — they are pure Go.
- Drop `binutils-gold` (and `build-base` where it was added only to support the cgo link step) and use `CGO_ENABLED=0` inline on each `go install`.

`CGO_ENABLED=0` is set inline only on the dev tool installs. App builds (via `air`, or the production `builder` stage) are unaffected.

Verified on linux/arm64 (emulated):
- agent dev image builds without `binutils-gold` / `build-base`
- `air -v` and `golangci-lint --version` run inside the arm64 image
- `go build -tags docker` of the agent succeeds inside the dev image and the resulting binary runs (`agent --help`)

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -58,9 +58,9 @@ FROM base AS development
 ARG GOPROXY
 ENV GOPROXY ${GOPROXY}
 
-RUN apk add --update openssl build-base binutils-gold openssh-client util-linux setpriv
-RUN go install github.com/air-verse/air@v1.62 && \
-    go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
+RUN apk add --update openssl openssh-client util-linux setpriv
+RUN CGO_ENABLED=0 go install github.com/air-verse/air@v1.62 && \
+    CGO_ENABLED=0 go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
 
 WORKDIR $GOPATH/src/github.com/shellhub-io/shellhub
 
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -75,11 +75,11 @@ ARG GOPROXY
 ARG MJML_VERSION
 ENV GOPROXY=${GOPROXY}
 
-RUN apk add --update openssl build-base binutils-gold docker-cli npm
+RUN apk add --update openssl docker-cli npm
 RUN npm install -g mjml@${MJML_VERSION}
-RUN go install github.com/air-verse/air@v1.62 && \
-    go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3 && \
-    go install github.com/vektra/mockery/v2/...@v2.53.2
+RUN CGO_ENABLED=0 go install github.com/air-verse/air@v1.62 && \
+    CGO_ENABLED=0 go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3 && \
+    CGO_ENABLED=0 go install github.com/vektra/mockery/v2/...@v2.53.2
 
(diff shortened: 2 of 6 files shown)


gustavosbarreto:
@danielgatis could you test this in your environment? You introduced `binutils-gold` in #6248 to fix the linux/arm64 build, so it would be good to confirm this swap to `CGO_ENABLED=0` still works for you. Thanks!

PR #6314, opened 2026-05-12
fix(api): load cloud env files in docker-compose wrapper

geovannewashington:
## What changed?

The `utils` script now automatically sources `../cloud/.env` and
`../cloud/.env.override` when the cloud repo is present as a sibling
directory. `SHELLHUB_BILLING`, which was duplicated from the cloud repo,
has been removed from the shellhub `.env`.

## Why

Cloud-specific environment variables were being duplicated in the
shellhub `.env`, which is not ideal since they are only relevant when
the private cloud repo is present. By loading the cloud repo's env
files directly, each repo owns its own configuration.

## How to test

> **Note:** `SHELLHUB_CLOUD=true` must be set in the **shellhub**
> `.env.override`, not in the cloud repo's `.env.override`. The cloud
> files are only sourced after the shellhub override is loaded.

> **Note:** This change only affects users with access to the private
> cloud repository.

1. Clone the cloud repo as a sibling of the shellhub repo
2. Start the environment with `./bin/docker-compose up -d`
3. Verify that cloud-specific vars like `SHELLHUB_BILLING` are correctly
   loaded from `../cloud/.env`


.env
```diff
@@ -113,7 +113,6 @@ SHELLHUB_ENTERPRISE_ADMIN_PASSWORD=
 # SHELLHUB_NPM_REGISTRY=http://localhost:4873
 
 # Billing configuration
-SHELLHUB_BILLING=stripe
 STRIPE_PUBLISHABLE_KEY=
```
otavio: If you are removing the cloud-specific variables, I believe you should also remove the others.

otavio (changes requested):

.env
geovannewashington: I believe `STRIPE_PUBLISHABLE_KEY` is the only remaining variable that could be considered cloud-specific, but it's actually a UI concern; it's the public Stripe key used by the frontend to tokenize payment inputs. ShellHub repo owns the UI and the `docker-compose.yml`.

The shellhub `docker-compose.yml` would need to stop referencing it, which means removing Stripe payment UI from the open-source frontend entirely and/or making it conditionally loaded. Neither is a small change.

geovannewashington: @otavio Follow-up: turns out there's a cleaner and simpler solution now addressed that allowed the removal of  `STRIPE_PUBLISABLE_KEY` from CE.

PR #6401, opened 2026-06-02
chore(ui): remove deprecated Vue frontend

luizhf42:
## What
Removes the unmaintained Vue `ui` application and every reference to its Compose service across runtime routing, CI/CD, and developer tooling, leaving `ui-react` as the sole frontend.

## Why
`ui-react` has reached parity and is ready to ship as the only frontend, while the Vue `ui` app is fully stale and no longer developed. Retiring it completes the migration and drops a second frontend that no longer earns its place in the build, CI, and gateway.

## Changes
- **gateway**: `/v1/` now returns a `301` redirect to `/` instead of proxying `ui:8080`, so existing bookmarks of the old UI land on `ui-react`. This was the only place routing production traffic to the old app.
- **compose**: removed the `ui` service from `docker-compose.yml`, `docker-compose.dev.yml`, and `docker-compose.enterprise.yml`, along with the gateway `depends_on: ui` and the `ui_node_modules` volume.
- **ci**: removed `ui` from the `qa.yml` matrix, paths filter, and the six `[UI]` steps, and from the `docker-build.yml` / `docker-publish.yml` matrices. `shellhubio/ui` is no longer built or published.
- **devscripts**: dropped the `vue` mode from `lint-code` and the `ui` case from `test-unit` (plus its now-dead `npm_test` helper).
- **tests**: dropped the unused `ServiceUI` handle from the integration harness; the backend tests never consumed it and `ui-react` is still health-gated by the test stack.
- **config**: removed the `ui/` entries from `.dockerignore` and `CODEOWNERS`.
- **ui/**: deleted the directory (824 files).

## Dependencies
Pairs with shellhub-io/cloud#2379, which removes the matching orphaned `ui` overrides from the cloud Compose files. Merge the two together (or cloud first) — until then the enterprise/cloud dev stack fails Compose validation.

## Testing
- `./bin/docker-compose config -q` validates for the base, dev, enterprise, and full enterprise/cloud stacks (the last with cloud#2379 checked out as the sibling repo); the merged render lists only `ui-react`.
- After bringing the stack up, `curl -sI http://localhost/v1/` returns `301` with `Location: /`.
- Confirm no production or cloud deployment still pulls the `shellhubio/ui` image before tagging a release, since it is no longer published.


luizhf42:
> **`tests/environment/` still references the removed `ui` service (high)**

Fixed. Removed the reference to the old UI service.

PR #6463, opened 2026-06-09
fix(api): upsert active_sessions to prevent duplicate key error on session update

### LINKED ISSUE #6459: active_sessions duplicate-key: ActiveSessionCreate raw insert has no ON CONFLICT
## Summary

`ActiveSessionCreate` does a raw `INSERT` into `active_sessions` with no `ON CONFLICT` clause, but the table's primary key is `session_id`. A single session inserts twice during its normal lifetime, so the second insert fails with:

```
ERROR: duplicate key value violates unique constraint "active_sessions_pkey"
```

The error is currently swallowed as a `Warning`, so it does not break session establishment, but it spams logs and the underlying raw-insert-without-upsert is a real defect.

## Root cause

`api/store/pg/session.go` (`ActiveSessionCreate`) is a plain insert:

```go
e := entity.ActiveSessionFromModel(activeSession)
if _, err := db.NewInsert().Model(e).Exec(ctx); err != nil {
    return fromSQLError(err)
}
```

into a table whose PK is `session_id` (`migrations/001_initial_schema.tx.up.sql`, `PRIMARY KEY (session_id)`).

It is called from `UpdateSession` (`api/services/session.go`) on **every** update where the re-resolved session is `Authenticated`, not gated to the authenticated transition:

```go
if session.Authenticated {
    if err := s.store.ActiveSessionCreate(ctx, session); err != nil {
        log.WithError(err).WithField("session_id", session.UID).Warn("failed to activate the session")
    }
}
```

## Reproduction (no reconnect needed)

Within a single recorded (pty) shell session:

1. `Session.Auth` -> `UpdateSession{Authenticated: true}` -> first `ActiveSessionCreate` insert (succeeds).
2. During the data pipe, `Recorded(seat)` -> `UpdateSession{Recorded: true}`. The service re-resolves the session (still `Authenticated=true`), so it calls `ActiveSessionCreate` **again** -> duplicate key on `active_sessions_pkey`.

The row is only removed on session close (`DeactivateSession`), so nothing clears it between those two calls. Re-auth / reconnect of the same session UID is an additional trigger but is not required.

## Impact

- Continuous `failed to activate the session` warnings plus a Postgres error per affected session.
- Wasted insert round-trips.
- The swallowed error masks the fact that `active_sessions` state is not being refreshed (`seen_at` never updated on the second call).

## Suggested fix

Make the write idempotent, either:

- `ON CONFLICT (session_id) DO UPDATE SET seen_at = EXCLUDED.seen_at` (turns it into a proper upsert that also refreshes the timestamp), or
- gate `ActiveSessionCreate` to the authenticated transition only (insert once), or
- `ON CONFLICT (session_id) DO NOTHING` if refreshing `seen_at` here is not needed.

The upsert is preferred since it both removes the error and keeps `seen_at` current.

geovannewashington:
## What

`ActiveSessionCreate` now uses `ON CONFLICT (session_id) DO UPDATE SET seen_at = NOW()` instead of a plain `INSERT`.

## Why

The service calls `ActiveSessionCreate` on every `UpdateSession` where `Authenticated=true`, not only on the initial authenticated transition. A single pty session triggers it at least twice - once on auth and again on the recorded update - causing a duplicate key violation on `active_sessions_pkey`. The error was swallowed as a warning so sessions were not broken, but it spammed logs and left `seen_at` stale on subsequent calls.

## Changes

- One-line change in `api/store/pg/session.go`: plain insert replaced with an upsert.
- No migration needed - no schema change.
- No service-layer changes - the fix is at the store boundary where the constraint is enforced.

## Testing

Open a recorded shell session through ShellHub and confirm no `failed to activate the session` warnings appear in the api service logs.

Closes: #6459

### DIFF
--- a/api/store/pg/session.go
+++ b/api/store/pg/session.go
@@ -154,7 +154,10 @@ func (pg *Pg) ActiveSessionCreate(ctx context.Context, session *models.Session) 
 
 	activeSession := &models.ActiveSession{UID: models.UID(session.UID), LastSeen: session.StartedAt, TenantID: session.TenantID}
 	e := entity.ActiveSessionFromModel(activeSession)
-	if _, err := db.NewInsert().Model(e).Exec(ctx); err != nil {
+	if _, err := db.NewInsert().Model(e).
+		On("CONFLICT (session_id) DO UPDATE").
+		Set("seen_at = NOW()").
+		Exec(ctx); err != nil {
 		return fromSQLError(err)
 	}
 


otavio:
There is any way for us to have a test to ensure that we don't have this error back in the future?

PR #6516, opened 2026-06-19
refactor(tests): migrate from mockery v2 to v3

### LINKED ISSUE #6509: Dev image ships only mockery v2; cloud needs v3 to regenerate mocks
## Problem

The dev container (`api/Dockerfile`) installs mockery **v2** only:

```
# api/Dockerfile:82
CGO_ENABLED=0 go install github.com/vektra/mockery/v2/...@v2.53.2
```

(ssh/gateway Dockerfiles pin v2.20.0; openapi pins v2.53.2.)

CE mocks are generated with v2 (`//go:generate mockery --name X --filename Y`), so that works for the `shellhub` repo.

The `cloud` repo, which is mounted into and built from the same `api` container, has moved to mockery **v3** (`cloud/.mockery.yaml`, `packages:` config, in-package `mock_<name>.go` with `//go:build mocks`). The v2 binary on `PATH` cannot read the v3 config, so there is no way to regenerate cloud mocks in-container with the provided tooling.

## Impact

When a CE store/service interface changes (e.g. the embedded `shellhub.Store` gains methods), the cloud `MockStore`/`MockService` must be regenerated. Today the only way to do that inside the dev container is:

```
go run github.com/vektra/mockery/v3@<version> 
```

which downloads v3 ad hoc on every run and is not discoverable.

## Request

Make mockery v3 available in the dev image so cloud mocks can be regenerated with the project's standard tooling. Options, in rough order of preference:

1. Install both v2 (for CE) and v3 (for cloud) in the image, exposed under distinct names (e.g. `mockery` = v2, `mockery-v3` = v3), and pin the v3 version.
2. Migrate CE codegen to mockery v3 as well and ship a single v3 binary (larger change; unifies tooling).

Either way, pin the v3 version to match `cloud/.mockery.yaml`'s expectations so generated output is stable.

geovannewashington:
Fixes: shellhub-io/shellhub#6509
Companion PR (must merge together): shellhub-io/cloud#2406

## What

Migrates ShellHub CE from mockery v2 to a single mockery v3 binary
(v3.7.1). Mocks keep their existing `mocks/` subpackage layout but are
regenerated with v3's `Mock<Interface>` struct names. 
No `//go:build mocks` tag is introduced - the mocks live in sibling packages, so 
`go build`/`go test`/`golangci-lint`/`gopls` all resolve them without
`-tags mocks` or editor buildFlags.

## Why

The repo mixed multiple mockery v2 versions and per-interface
`//go:generate` directives. Moving to v3 gives one pinned toolchain and a
centralized per-module `.mockery.yaml`, while staying on the subpackage
layout that already works without build-tag plumbing.

## Changes (116 files)

- **5 per-module `.mockery.yaml`** (root, api, agent, gateway, ssh). CE is
  multi-module with no `go.work`, so each module gets its own config.
- **Mocks regenerated** under their existing `mocks/` dirs as
  `mock_<name>.go` with `Mock<Interface>` names; old v2 mock files removed
  (incl. a stale `api/store/mocks/cache.go` orphan and a `README`).
- **5 Dockerfiles** install `mockery/v3@v3.7.1` in their `development`
  stage only (api, ssh, gateway, openapi, agent).
- **`devscripts/gen-mock`** rewritten to one mockery pass per module, plus
  a new `tests/devscripts/gen-mock_test.sh` guarding the layout (no build
  tag, expected files present).
- **14 `//go:generate mockery` directives stripped**; one stray directive
  with no mock/consumer dropped; the regenerated uuid mock loses an unused
  `Parse`.
- **`.golangci.yaml`** excludes the generated `mocks/` directories.
- Test files updated to the `Mock<Interface>` symbols (note
  `uuid.MockUUID`).

## Testing

- Each module builds and tests without any `mocks` tag.
- `golangci-lint` clean.
- Only pre-existing failure is `TestMain_smoke` (gateway, needs rootless
  Docker) - not a regression.


gustavosbarreto:
Can we keep the mocks as a separate package here (the `mocks/` subpackage layout CE already had), and have cloud move to match CE on this, rather than CE adopting cloud's in-package + build-tag layout?

The v2 → v3 upgrade itself is great and worth doing. My only push is on placement: drop `//go:build mocks` (`mock-build-tags`) and the in-package output, and generate into a sibling package instead, roughly:

```yaml
# .mockery.yaml (v3)
template: testify
dir: "{{.InterfaceDir}}/mocks"
pkgname: "mocks"
```

The main reason is that it needs no extra build tag, so it integrates out-of-the-box with the toolchain. With a separate package, `go test`, `go build ./...`, `golangci-lint`, and gopls all work with zero flags. The in-package + `//go:build mocks` approach forces `-tags mocks` on every test/build/lint invocation, plus a `build.buildFlags: ["-tags=mocks"]` in everyone's editor config, otherwise the mock symbols come up as "undefined" / "file excluded from build" on a fresh clone. The separate package is excluded from production builds naturally (prod never imports it), so we get the same "no testify in the prod binary" guarantee without carrying the tag everywhere.

On import cycles (the usual reason to go in-package): CE compiled and tested fine on the subpackage layout before this PR, so we're not actually hitting them. The narrow case where a package mocks its own interface in its own internal test is solved by moving that test to the external `pkg_test` package, which we generally want anyway, and it needs no build tag.

For the "these generated files are noisy in the diff and in blame" concern, we can mark them as generated instead of pulling them out of the repo, via `.gitattributes`:

```
**/mocks/*.go linguist-generated=true
```

That collapses them in the PR diff (shows as "Load diff"), keeps them out of `git blame`, and drops them from the repo language stats, while `go test` still works on a clean clone with no extra tooling. Docs: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

PR #6530, opened 2026-06-22
docs: update screenshots and fix misleading information

Ossanes:
- Replace some screenshots to improve user understanding.
- Fix inaccurate text descriptions that could mislead users.


otavio:
Some images appear to no longer be referenced. Additionally, the file naming style seems inconsistent; please verify if you can maintain a consistent pattern to prevent mixing different styles within the same repository.

PR #6558, opened 2026-06-25
feat(ui): add column sorting to backend-ready data-tables

luannmoreira:
## What

Adds user-controllable column sorting to four console data-tables whose API endpoints already accept `sort_by`/`order_by`: **Devices**, **Containers**, **API Keys**, and the billing **Device Chooser** ("All Devices" tab). Frontend-only — no backend change.

## Why

These lists rendered a `DataTable` with no sorting, even though the component and the backing endpoints already supported it. Only the admin Devices list was wired up. This brings the user-facing tables in line.

Fixes: shellhub-io/team#165

## Changes

Each table mirrors the existing `admin/devices` pattern — local `sortBy`/`orderBy` state, a `handleSort` toggle (same field flips asc⇄desc; switching field resets direction; every sort resets to page 1), `sortField`/`sortOrder`/`onSort` passed to `DataTable`, and the data hook forwarding `sort_by`/`order_by` to the SDK query.

- **Devices** (`pages/devices`, `hooks/useDevices`): sortable **Hostname** (`name`), **Last Seen** (`last_seen`); default `last_seen` desc.
- **Containers** (`pages/containers`, `hooks/useContainers`): sortable **Hostname** (`name`), **Last Seen** (`last_seen`); default `last_seen` desc.
- **API Keys** (`pages/team/ApiKeysTab`, `hooks/useApiKeys`): sortable **Name** (`name`), **Created** (`created_at`), **Expires** (`expires_in`); default `created_at` desc. Field names match the backend sorter (`api/routes/api-key.go`, `api/services/api-key.go`).
- **Device Chooser** (`components/billing/DeviceChooserDialog`, "All" tab): sortable **Hostname**, **Last Seen**.

Column `key`s were aligned to the backend sort fields (`hostname`→`name`, `created`→`created_at`, `expires`→`expires_in`) because `DataTable` passes the column key as the sort field.

## Testing

TDD (red→green) per table. New/updated test files for all four assert that sortable headers render as controls and that clicking them drives the hook with the expected `sortBy`/`orderBy` and toggles direction.

`cd ui/apps/console && npm run test -- <the four table test files>` → 72 pass; `tsc --noEmit` and `eslint` clean.

Non-sortable derived columns (tags, online indicator, role) are intentionally left non-sortable. Tables whose endpoints lack sort params (Sessions, Firewall Rules, Public Keys, several admin lists) are out of scope and tracked separately in shellhub-io/team#165.

### DIFF
--- a/ui/apps/console/src/components/billing/DeviceChooserDialog.tsx
+++ b/ui/apps/console/src/components/billing/DeviceChooserDialog.tsx
@@ -73,7 +73,20 @@ export default function DeviceChooserDialog({
   const [page, setPage] = useState(1);
   const [selected, setSelected] = useState<NormalizedDevice[]>([]);
   const [error, setError] = useState<string | null>(null);
+  const [sortBy, setSortBy] = useState<"name" | "last_seen">("last_seen");
+  const [orderBy, setOrderBy] = useState<"asc" | "desc">("desc");
   const inFlightRef = useRef(false);
+
+  const handleSort = (field: string) => {
+    const f = field as "name" | "last_seen";
+    if (sortBy === f) {
+      setOrderBy((prev) => (prev === "asc" ? "desc" : "asc"));
+    } else {
+      setSortBy(f);
+      setOrderBy(f === "name" ? "asc" : "desc");
+    }
+    setPage(1);
+  };
 
   // Force the All tab whenever Suggested is empty so a refetch that returns
   // [] doesn't strand the user on a tab that would submit zero choices.
@@ -91,6 +104,8 @@ export default function DeviceChooserDialog({
     status: "accepted",
     search: debouncedSearch,
     enabled: tab === "all",
+    sortBy,
+    orderBy,
   });
 
   const totalPages = Math.max(1, Math.ceil(totalCount / PER_PAGE));
@@ -239,6 +254,9 @@ export default function DeviceChooserDialog({
               totalPages={totalPages}
               totalCount={totalCount}
               onPageChange={setPage}
+              sortField={sortBy}
+              sortOrder={orderBy}
+              onSort={handleSort}
             />
             <SelectedChips
               selected={selected}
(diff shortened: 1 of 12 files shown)


luizhf42:
Solid change overall. I was just thinking about extracting that sorting logic into a shared hook, because most of it is repeated in the pages: the field and order `useState`s, the order toggle, and the default order being `asc` on `name` fields.

This would reduce the code repetition and also make it easier to integrate into other tables in the future, if any other gets the sorting logic in the back-end. Do you think it's valid?

otavio:
> Solid change overall. I was just thinking about extracting that sorting logic into a shared hook, because most of it is repeated in the pages: the field and order `useState`s, the order toggle, and the default order being `asc` on `name` fields.
> 
> This would reduce the code repetition and also make it easier to integrate into other tables in the future, if any other gets the sorting logic in the back-end. Do you think it's valid?

@luannmoreira Please verify if we can implement it this way, so we avoid duplicating all that.

luannmoreira:
Good call @luizhf42 — addressed in 64c3e4ef2.

Extracted the repeated state into a generic `useTableSort<TField>` hook (`src/hooks/useTableSort.ts`) that owns the `sortBy`/`orderBy` state, the asc/desc toggle, and the default-ascending-on-`name` rule. It takes `defaultField`, an optional `defaultOrder` (defaults to `desc`), and an `onSortChange` callback for the `setPage(1)` reset:

```ts
const { sortBy, orderBy, handleSort } = useTableSort<SortField>({
  defaultField: "last_seen",
  onSortChange: () => setPage(1),
});
```

Migrated all 5 call sites — Devices, Containers, API Keys, Device Chooser, and the pre-existing admin Devices table (the pattern this PR mirrored), so there's no stray copy left. Behavior is unchanged; the existing sort tests (72 across the 4 suites) still pass. New sortable tables can now just reuse the hook.

luizhf42 (approved):

PR #6573, opened 2026-06-29
refactor(ui): centralize device & container action logic

luannmoreira:
## What
Accept/reject/remove dialog wiring and the billing-warning gate were duplicated across `DeviceDetails`, `ContainerDetails`, and both list pages. This extracts a single shared implementation that all four pages consume.

## Why
Closes shellhub-io/team#153. Any bug in the action flow, loading state, or error handling previously had to be fixed in each page independently. (The audit named three files; the real duplication spans four — `admin/devices` has no action logic, while `ContainerDetails` does.)

## Changes
- **`hooks/useActionDialogState`**: generic state core — `operation` state, `requestAction`, `close`, the cloud-gated `onBillingWarning`, and `runSuccess(action)`. All returned callbacks are `useCallback`-stable so list-page column `useMemo`s don't churn.
- **`hooks/useDeviceActions` / `useContainerActions`**: thin typed wrappers (the issue-named public API); `enableBillingWarning` defaults to `!!getConfig().cloud`.
- **`DeviceActionsPortal` / `ContainerActionsPortal`**: render the existing `*ActionDialog` + `BillingWarning` from the controller, preserving the `key`-based reset and the entity→`device`/`container` prop mapping.
- **Pages**: `DeviceDetails`, `devices/index`, `ContainerDetails`, `containers/index` drop their inline state/render in favor of the hook + portal. `ContainerDetails` passes `enableBillingWarning: false` to keep its existing no-billing-warning flow.
- The `*ActionDialog` and `BillingWarning` components are unchanged — they already centralized the mutations and 402 handling.

## Testing
- New unit tests: `useActionDialogState` (state, billing gating, stable refs, action forwarding) and both portals (prop mapping; `ContainerActionsPortal` asserts `BillingWarning` is absent when `enableBillingWarning: false`).
- Existing page-test mocks were retargeted from the dialogs to the new portals.
- Full console suite: 2597 passing; `eslint` and `tsc -b` clean.
- Reviewer note: confirm `ContainerDetails` still shows the inline 402 error (not the billing dialog) on a cloud build — that asymmetry vs. the other pages is intentional and preserved.

### DIFF
--- a/ui/apps/console/src/hooks/__tests__/useActionDialogState.test.ts
+++ b/ui/apps/console/src/hooks/__tests__/useActionDialogState.test.ts
@@ -1,0 +1,176 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { renderHook, act } from "@testing-library/react";
+import {
+  useActionDialogState,
+  type UseActionDialogStateOptions,
+} from "../useActionDialogState";
+import { useDeviceActions } from "../useDeviceActions";
+import { useContainerActions } from "../useContainerActions";
+import { defaultConfig, type ClientConfig } from "@/env";
+
+// Only the wrapper hooks read getConfig().cloud; the core hook never touches env.
+const mockGetConfig = vi.fn<() => ClientConfig>();
+vi.mock("@/env", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@/env")>()),
+  getConfig: () => mockGetConfig(),
+}));
+
+const entity = { uid: "device-123", name: "Device 123" };
+
+const setup = (options: Partial<UseActionDialogStateOptions> = {}) =>
+  renderHook(() => useActionDialogState({ enableBillingWarning: false, ...options }));
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  mockGetConfig.mockReturnValue({ ...defaultConfig, cloud: false });
+});
+
+describe("useActionDialogState", () => {
+  it("requestAction opens the dialog for an entity/action pair", () => {
+    const { result } = setup();
+
+    act(() => result.current.requestAction(entity, "remove"));
+
+    expect(result.current.operation).toEqual({ entity, action: "remove" });
+  });
+
+  it("close clears the operation", () => {
+    const { result } = setup();
+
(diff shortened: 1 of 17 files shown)


ui/apps/console/src/hooks/__tests__/useActionDialogState.test.ts
```diff
@@ -0,0 +1,565 @@
...
+  version: "",
+  enterprise: false,
+  cloud: false,
+  announcements: false,
+  webEndpoints: false,
+  onboardingUrl: "",
+  stripePublishableKey: "",
+  chatwootWebsiteToken: "",
+  chatwootBaseUrl: "",
+};
```
luizhf42: You could use the `env.ts` `defaultConfig` in the mock, getting rid of the re-defined default values, and manipulating only the `cloud` boolean.

ui/apps/console/src/hooks/useActionDialogState.ts
```diff
@@ -0,0 +1,113 @@
+import { useState, useCallback, useRef, useEffect } from "react";
+
+export interface Operation<TEntity = string, TAction extends string = string> {
+  entity: TEntity;
+  action: TAction;
+}
```
luizhf42: Why do those generics exist? Entity is always an object with `uid` and `name`, and action is always `accept`, `reject`, or `remove`, for both devices and containers. 

Also, the action dialogs have the exact duplicate entity interface.

ui/apps/console/src/hooks/__tests__/useActionDialogState.test.ts
luannmoreira: Done in 3ba8d27f7 — the mock now spreads the real `defaultConfig` and only flips `cloud`, dropping the duplicated literal.

ui/apps/console/src/hooks/useActionDialogState.ts
luannmoreira: Good call — removed in 3ba8d27f7. `useActionDialogState` now uses the concrete `EntityBase`/`EntityAction` types instead of generics, and both action dialogs reuse the shared `EntityBase` rather than redefining their own interface.

PR #129, opened 2020-05-21
ui: Integrating a device delete component 

noreng-jg:
A device delete component was added to
the device folder which consists in a 
vuetify dialog. Remaining tooltips for other 
icons were also set.

PR #6587, opened 2026-06-30
refactor(ui): promote DataTable, Pagination, PageLoader to design-system

luannmoreira:
## What
Promotes three reusable console components — `DataTable`, `Pagination`, and `PageLoader` — out of `@shellhub/console` and into `@shellhub/design-system/components`, so other apps can consume them. Pure relocation; the only behavioral change is `DataTable` adopting the package's `cn`.

## Why
`@shellhub/design-system` is the shared UI foundation, but these widely-used components were siloed in the console app. A design-system component cannot import from the console, so each transitive console dependency had to move with it.

Closes shellhub-io/team#151

Scope note: the issue also listed `Alert` (does **not** exist in the codebase) and `ConfirmDialog` (deferred — it drags `BaseDialog` plus the `useFocusTrap`/`useBackdropClose`/`useResetOnOpen` a11y hooks; better as its own PR). Issue #129 (Pagination ARIA) is already closed, so `Pagination` moved as-is.

## Changes
- **design-system**: added `DataTable`, `Pagination`, `PageLoader` under `components/` as **named exports** and to the barrel; relocated their test suites into the package. `DataTable` pulled in `PageLoader` (its only console-internal component dep).
- **cn**: `DataTable` now uses the package's `cn` (`twMerge(clsx(...))`) instead of the console's naive `filter(Boolean).join(" ")`. Class props (`Column.headerClassName`, `DataTableProps.rowClassName`) are now merged via `tailwind-merge` — conflicting utilities are deduplicated rather than concatenated. Behavior-preserving for current call sites.
- **package manifest**: declared the `@heroicons/react` peer dependency that `DataTable` requires (previously satisfied only by workspace hoisting).
- **importers**: repointed every console importer to `@shellhub/design-system/components` and deleted the old `common/` files. Default imports became named (`import { DataTable, type Column }`).
- **tests**: rewrote the two firewall page tests' `DataTable` `vi.mock` for the named export (mock path, `importOriginal` type, and factory key `default` → `DataTable`).

## Testing
- `tsc --noEmit` (console): clean — confirms no importer was missed (a default-vs-named mismatch would error).
- design-system vitest: 84 pass (`DataTable` 57, `Pagination` 16, `PageLoader` 11).
- console vitest for the rewritten mocks: `AdminFirewallRules` + `FirewallRules`, 38 pass.
- `eslint` on all touched console files: clean.
- No remaining `common/{DataTable,Pagination,PageLoader}` references in the console.

### DIFF
--- a/ui/apps/console/src/components/ManageTagsDrawer.tsx
+++ b/ui/apps/console/src/components/ManageTagsDrawer.tsx
@@ -18,7 +18,7 @@ import {
   ExclamationCircleIcon,
 } from "@heroicons/react/24/outline";
 import { Button, IconButton } from "@shellhub/design-system/primitives";
-import PageLoader from "@/components/common/PageLoader";
+import { PageLoader } from "@shellhub/design-system/components";
 
 const TAG_PATTERN = /^[a-zA-Z0-9]+$/;
 
--- a/ui/apps/console/src/components/billing/BillingPayment.tsx
+++ b/ui/apps/console/src/components/billing/BillingPayment.tsx
@@ -33,7 +33,7 @@ import { stripeErrorMessage } from "@/utils/stripeErrors";
 import FieldLabel from "@/components/common/fields/FieldLabel";
 import InputField from "@/components/common/fields/InputField";
 import BillingIcon from "./BillingIcon";
-import PageLoader from "@/components/common/PageLoader";
+import { PageLoader } from "@shellhub/design-system/components";
 
 const ELEMENTS_OPTIONS: StripeElementsOptions = {
   appearance: { theme: "night" },
--- a/ui/apps/console/src/components/billing/DeviceChooserDialog.tsx
+++ b/ui/apps/console/src/components/billing/DeviceChooserDialog.tsx
@@ -9,7 +9,7 @@ import {
 import { useNavigate } from "react-router-dom";
 import { ExclamationCircleIcon, XMarkIcon } from "@heroicons/react/24/outline";
 import BaseDialog from "../common/BaseDialog";
-import DataTable, { type Column } from "../common/DataTable";
+import { DataTable, type Column } from "@shellhub/design-system/components";
 import DistroIcon from "../common/DistroIcon";
 import OnlineDot from "../common/OnlineDot";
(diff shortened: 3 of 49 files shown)


luizhf42:
Do you think that website or docs will use those components someday? 

I opened an issue in the team (https://github.com/shellhub-io/team/issues/114) documenting promotion candidates, considering that only components used in more than one app should be there to prevent cluttering the design system with components that could be local in the app's common components. 

By the way, the issue excluded those 3 from the promotion due to this: no use in docs or website, and I'm not sure if "future-proofing" is a good idea here. What do you think?

luannmoreira:
> Do you think that website or docs will use those components someday?
> 
> I opened an issue in the team ([shellhub-io/team#114](https://github.com/shellhub-io/team/issues/114)) documenting promotion candidates, considering that only components used in more than one app should be there to prevent cluttering the design system with components that could be local in the app's common components.
> 
> By the way, the issue excluded those 3 from the promotion due to this: no use in docs or website, and I'm not sure if "future-proofing" is a good idea here. What do you think?

You're right, and thanks for pointing me to #114. I'd missed it.

  I don't think the website or docs will realistically use these components. I checked the usage, and they're all console-only today. `DataTable` and `PageLoader` are each used in about 23 files, Pagination is only used in a handful, and the docs match is just a ## Pagination heading, not the component itself. The website is marketing React and the docs are Astro with no React, so there's no real need for a sortable data grid, pagination, or a page loader. This falls into the kind of speculative future-proofing you called out, and I agree the rule of three should apply here, just like with the Section primitives in step 8.

  The confusion came from following team#151, which lists these components for promotion. Looking back, #114 had already excluded them, and that's the issue that actually defines the scope of this work. I focused on the console and package without asking whether there was a real second consumer.

  I think the best approach is to close this PR and reconcile [#151](https://github.com/shellhub-io/team/issues/151)  with   [#114](https://github.com/shellhub-io/team/issues/114). My guess is that the promotion list in #151 is simply outdated. If anything is worth revisiting later, it'd probably be PageLoader, but even that mostly wraps the shared Spinner, so there's not much value until another consumer actually exists.

PR #6597, opened 2026-07-01
refactor(ui): adopt react-hook-form for the SignUp form

luannmoreira:
## What

Introduce [react-hook-form](https://react-hook-form.com/) as the shared
form-management solution for `@shellhub/console`, and migrate the SignUp form to
it as the pilot. Establishes a reusable pattern (RHF-bound field wrappers + a
resolver adapter) that subsequent forms reuse.

## Why

Every form in the console hand-threads `useState` + `onChange` + `touched` +
memoized validation. As forms grow (firewall rules, SAML, multi-step MFA) this
repetition becomes a maintenance burden. This is the first step of the
incremental migration tracked in `shellhub-io/team#152`, which now carries a
checklist of the remaining forms — each to be migrated in its own PR.

The `Controller`/`useController` pattern is used because the design-system field
components (`InputField`, `PasswordField`, `CheckboxField`) are controlled and
do not forward refs, so RHF's uncontrolled `register` mode does not apply.
Validation reuses the existing `validate()` via a thin resolver adapter rather
than adopting zod, keeping a single source of validation truth and avoiding a
new dependency for the pilot.

## Changes

- **fields/rhf**: new `FormInputField`, `FormPasswordField`, `FormCheckboxField`
  wrappers (generic over the form values type) that bind the existing controlled
  field components to RHF via `useController`. Each accepts an `error` override
  and an `onValueChange` hook so store-driven server errors stay orthogonal to
  client validation.
- **signUpResolver**: resolver adapter that calls the existing `validate()`,
  maps `FormErrors` to RHF's error shape, and adds the privacy-policy-required
  rule. Exports the `SignUpFormValues` type used as the form's generic.
- **SignUp**: replaced the manual field `useState` + `touched` + validation
  memos with `useForm`. Behavior is preserved: errors surface after blur, the
  invite flow keeps the email prefilled/disabled and present in the payload, and
  the submit button stays gated on server errors (`signUpServerFields`) in
  addition to `formState.isValid`. `acceptMarketing` stays local state since it
  is unvalidated.

## Testing

`vitest` (run inside the `ui` container). 45 tests added across the wrappers,
the resolver, and SignUp. Worth probing: the server-error path — the submit
button must stay disabled while a server-side field error is shown and
re-enable once that field is edited.

### DIFF
--- a/ui/apps/console/src/components/common/fields/rhf/FormInputField.tsx
+++ b/ui/apps/console/src/components/common/fields/rhf/FormInputField.tsx
@@ -1,0 +1,43 @@
+import { useController, type Control, type FieldValues, type Path } from "react-hook-form";
+import InputField from "@/components/common/fields/InputField";
+import type { ComponentProps } from "react";
+
+type InputFieldProps = Omit<
+  ComponentProps<typeof InputField>,
+  "value" | "onChange"
+>;
+
+type Props<T extends FieldValues> = InputFieldProps & {
+  name: Path<T>;
+  control: Control<T>;
+  /** Called on every value change, in addition to RHF's internal onChange. */
+  onValueChange?: (value: string) => void;
+};
+
+export default function FormInputField<T extends FieldValues>({
+  name,
+  control,
+  error: errorOverride,
+  onValueChange,
+  ...rest
+}: Props<T>) {
+  const {
+    field,
+    fieldState: { error: fieldError },
+  } = useController({ name, control });
+
+  const resolvedError = errorOverride ?? fieldError?.message;
+
+  return (
+    <InputField
+      {...rest}
+      value={field.value}
+      onChange={(v) => {
+        field.onChange(v);
+        onValueChange?.(v);
+      }}
+      onBlur={field.onBlur}
+      error={resolvedError}
+    />
+  );
+}
--- a/ui/apps/console/src/components/common/fields/rhf/index.ts
+++ b/ui/apps/console/src/components/common/fields/rhf/index.ts
@@ -1,0 +1,3 @@
+export { default as FormInputField } from "./FormInputField";
+export { default as FormPasswordField } from "./FormPasswordField";
+export { default as FormCheckboxField } from "./FormCheckboxField";
--- a/ui/apps/console/src/pages/__tests__/SignUp.test.tsx
+++ b/ui/apps/console/src/pages/__tests__/SignUp.test.tsx
@@ -1,0 +1,385 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { render, screen, cleanup, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { MemoryRouter } from "react-router-dom";
+import { useSignUpStore } from "@/stores/signUpStore";
+import SignUp from "../SignUp";
+
+/* ------------------------------------------------------------------ */
+/* Mocks                                                               */
+/* ------------------------------------------------------------------ */
+
+const mockNavigate = vi.hoisted(() => vi.fn());
+
+vi.mock("react-router-dom", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("react-router-dom")>();
+  return { ...actual, useNavigate: () => mockNavigate };
+});
+
+vi.mock("@/client", () => ({
+  registerUser: vi.fn(),
+  resendEmail: vi.fn(),
+  getValidateAccount: vi.fn(),
+}));
+
+import { registerUser as registerUserSdk } from "@/client";
+
+const mockedRegisterUser = vi.mocked(registerUserSdk);
+
+type SdkResponse<T = unknown> = { data: T; request: Request; response: Response };
+
+function mockSdkResponse<T>(data: T): SdkResponse<T> {
+  return {
+    data,
+    request: new Request("http://localhost"),
+    response: new Response(),
+  };
+}
+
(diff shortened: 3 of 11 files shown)


otavio:
@luannmoreira rebase plls

ui/apps/console/src/components/common/fields/rhf/FormInputField.tsx
```diff
@@ -0,0 +1,53 @@
...
+  hideLabel?: boolean;
+  error?: string;
+  errorRole?: "alert" | "status";
+  hint?: string;
+  appendIcon?: ReactNode;
+  prependIcon?: ReactNode;
+  variant?: "default" | "mono";
+  /** Called on every value change, in addition to RHF's internal onChange. */
+  onValueChange?: (value: string) => void;
+};
```
luizhf42: You could use the `InputField` props and only omit what isn't necessary (`onChange`, `value`, etc.) instead of manually repeating all the props here, just like you did in the other two form field components.

ui/apps/console/src/pages/setup/__tests__/signUpResolver.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  describe("validate() field error mapping", () => {
+    it("maps name error to RHF error shape", async () => {
+      const input: SignUpFormValues = { ...validInput, name: "" };
+      const result = await signUpResolver(input, undefined, emptyOptions);
+      expect(result.errors).toHaveProperty("name");
+      expect((result.errors as Record<string, { type: string; message: string }>).name.type).toBe("validate");
+      expect(typeof (result.errors as Record<string, { type: string; message: string }>).name.message).toBe("string");
+      expect(result.values).toEqual({});
+    });
```
luizhf42: Almost all tests in that block repeat the same formula: generate an invalid input based on the valid one → call the resolver → check if the result errors have the invalid property → check the `validate` type → check if the message is a `string` (also repeating the type casting) → expect values to be an empty object. This could be extracted and reused, I guess.

ui/apps/console/src/components/common/fields/rhf/__tests__/FormInputField.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+      username: "alice",
+      email: "alice@example.com",
+      password: "Secret123!",
+      confirmPassword: "Secret123!",
+      acceptPrivacyPolicy: true,
+    };
+
+    expect(_typeCheck.username).toBe("alice");
+  });
+});
```
luizhf42: I don't see much value in testing whether the components are re-exported. Do you think it's necessary?

ui/apps/console/src/components/common/fields/rhf/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as FormInputField } from "./FormInputField";
+export { default as FormPasswordField } from "./FormPasswordField";
+export { default as FormCheckboxField } from "./FormCheckboxField";
+export type { SignUpFormValues } from "@/pages/setup/signUpResolver";
```
luizhf42: I'm not sure if that export should be here. That type is only used by the local consumers in `@/pages`, so it doesn't seem like the right place for that export, and it isn't even imported from that file except for the "re-exports" tests.

ui/apps/console/src/components/common/fields/rhf/FormInputField.tsx
luannmoreira: Done in 620ee7108 — `FormInputField` now derives its props from `Omit<ComponentProps<typeof InputField>, "value" | "onChange">`, matching the other two wrappers.

ui/apps/console/src/pages/setup/__tests__/signUpResolver.test.ts
luannmoreira: Good call — extracted in 620ee7108 into an `expectFieldError` helper driven by an `it.each` table, so each case is now just the input that triggers the error.

ui/apps/console/src/components/common/fields/rhf/__tests__/FormInputField.test.tsx
luannmoreira: Agreed, removed in 1a89f4a61. (The wrapper suites were also consolidated into a single parametrized `formFields.test.tsx`.)

ui/apps/console/src/components/common/fields/rhf/index.ts
luannmoreira: You're right — removed the re-export in 1a89f4a61. The type stays in `@/pages/setup/signUpResolver`, where its real consumers import it from.

ui/apps/console/src/pages/__tests__/SignUp.test.tsx
```diff
@@ -0,0 +1,379 @@
...
+      // Edit the username field — server error should be cleared. The clear is
+      // synchronous with the keystroke, so waitForElementToBeRemoved can't be
+      // used (the element is already gone by call time); assert absence instead.
+      await user.type(screen.getByLabelText(/^username$/i), "a");
+
+      await waitFor(() =>
+        expect(
+          screen.queryByText(/this username already exists/i),
+        ).not.toBeInTheDocument(),
+      );
```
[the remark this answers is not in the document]
luannmoreira: Addressed in b0ec3f95d — the test now also asserts the submit button becomes enabled after the edit, binding it to the `clearSignUpServerField` gate rather than RHF's incidental revalidation.

PR #6639, opened 2026-07-13
fix(api): stop full-model updates from clobbering targeted-write columns

### LINKED ISSUE #6637: PostgreSQL store: full-model updates clobber columns maintained by separate targeted writes
## Summary

Several `*Update` methods in `api/store/pg/` issue a full-model bun update
(`db.NewUpdate().Model(x).WherePK()` with no `OmitZero()` and no `Column(...)`
allowlist). A full-model update writes every mapped column from the in-memory
snapshot. When the same table has a column that is maintained by a separate
targeted write (an atomic counter increment, a heartbeat, a NULL clear), a
full-model update that carries a stale value of that column silently overwrites
the concurrently-maintained value.

The failure is a lost update: it only shows up under concurrency (a counter
increment or heartbeat landing between the `resolve` and the `save` of a
full-model update), so it passes tests that use mocked stores and is easy to
miss in review.

## Root pattern

```go
setupKey, _ := store.SetupKeyResolve(...)  // snapshot, incl. used_times = 3
setupKey.Name = "renamed"                  // caller changes one field
store.SetupKeyUpdate(ctx, setupKey)         // full update rewrites used_times = 3
// meanwhile SetupKeyIncrementUsage bumped used_times to 4; now rolled back to 3
```

## Precedent already applied

`SetupKeyUpdate` (`api/store/pg/setup-key.go`) hit this: `OmitZero()` was masking
it (and separately breaking legitimate zero writes like clearing `expires_at` to
NULL, `disabled = false`, `usage_limit = 0`). The fix drops `OmitZero`, keeps the
full-model update, and marks the usage counters with bun's `,skipupdate` tag so
they stay out of every generated UPDATE while still allowing INSERT and explicit
`.Set("col = ...")` targeted writes:

```go
UsedTimes  int        `bun:"used_times,skipupdate"`
LastUsedAt *time.Time `bun:"last_used_at,nullzero,skipupdate"`
```

`,skipupdate` is the clean remedy only when the full-update method never
legitimately writes the racy column. Two of the cases below are mixed-writer:
the full update is itself a valid writer of the column, so a bare tag would break
a real path. Those need the writing path routed through a targeted update first.

### 1. HIGH, clean fix: `NamespaceUpdate` clobbers device counters

- Full update: `namespace.go:202` writes `devices_accepted/pending/rejected/removed_count` (mapped in `entity/namespace.go` `FromModel`).
- Maintainers: `NamespaceIncrementDeviceCount` (`namespace.go:218`), `NamespaceSyncDeviceCounts` (`namespace.go:242`, `264`).
- Increments fire on every device accept/reject/remove/delete, so the window is hot. No `NamespaceUpdate` caller (settings edit, rename) intends to write counts.
- Remedy: add `,skipupdate` to the four count fields in `entity/namespace.go`.

### 2. MED, clean fix: `DeviceUpdate` clobbers `custom_fields`

- Full update: `device.go:171` writes `custom_fields`.
- Maintainers: `DeviceSetCustomField` (`device.go:186`), `DeviceDeleteCustomField` (`device.go:205`), the only intended writers.
- No `DeviceUpdate` path (rename, status, merge, delete) means to write `custom_fields`, yet each carries a resolve-time snapshot.
- Remedy: add `,skipupdate` to `custom_fields` in `entity/device.go`.

### 3. HIGH impact, needs care: `DeviceUpdate` vs `DeviceHeartbeat` on `last_seen` / `disconnected_at`

- Full update: `device.go:171` writes `last_seen` and `disconnected_at`.
- Maintainer: `DeviceHeartbeat` (`device.go:225`), the high-frequency keepalive path.
- A rename/status/merge/delete `DeviceUpdate` that lands between a heartbeat and its next tick rolls `last_seen` back to the stale snapshot, so the device flickers offline.
- Not a bare `skipupdate`: the agent-auth online transition intentionally writes these columns through `DeviceUpdate` (`services/auth.go` sets `LastSeen`/`DisconnectedAt` and relies on `DeviceUpdate` to bring the device online). Route that transition through `DeviceHeartbeat` (or a dedicated targeted `.Set`) first, then add `,skipupdate`.

### 4. LOW-MED, mixed-writer: `UserUpdate` vs `preferred_namespace_id`

- Full update: `user.go:115` writes `preferred_namespace_id`.
- Maintainers: `NamespaceDeleteMembership` (`member.go:61`), `namespaceDeleteManyFn` (`namespace.go:353`) clear it.
- Not `skipupdate`: login legitimately sets it via `UserUpdate` (`services/auth.go`). A profile-edit `UserUpdate` carrying a stale populated value can restore a preferred namespace a concurrent deletion just cleared. Give preferred-namespace its own targeted write, or ensure clears win via re-fetch.

### 5. LOW: `UserInvitationUpdate` allowlist includes the incremented `invitations`

- `user_invitation.go:69` uses `Column("email","status","invitations","updated_at")`; `invitations` is separately incremented by the upsert at `user_invitation.go:33`.
- A status-change update carrying a stale count clobbers a concurrent re-invite increment.
- Remedy: drop `"invitations"` from the `Column(...)` allowlist.

### Secondary: `ActiveSessionUpdate` resets `created_at`

- `session.go:190` is a full-model update; `ActiveSessionFromModel` always stamps `CreatedAt: clock.Now()`, so every update overwrites the row creation timestamp with "now". Not a cross-writer clobber, but a derived-value bug. Fix with `,skipupdate` on `ActiveSession.CreatedAt` or switch the method to `OmitZero`.

## Summary table

| table | full-update method | column(s) at risk | maintainer | remedy |
|---|---|---|---|---|
| namespaces | `NamespaceUpdate` (namespace.go:202) | `devices_*_count` (x4) | `NamespaceIncrementDeviceCount`, `NamespaceSyncDeviceCounts` | `,skipupdate` x4 (clean) |
| devices | `DeviceUpdate` (device.go:171) | `custom_fields` | `DeviceSetCustomField` / `DeviceDeleteCustomField` | `,skipupdate` (clean) |
| devices | `DeviceUpdate` (device.go:171) | `last_seen`, `disconnected_at` | `DeviceHeartbeat` (device.go:225) | migrate auth online path to targeted write, then `,skipupdate` |
| users | `UserUpdate` (user.go:115) | `preferred_namespace_id` | `NamespaceDeleteMembership`, `namespaceDeleteManyFn` | targeted write for login, not `skipupdate` |
| user_invitations | `UserInvitationUpdate` (user_invitation.go:69) | `invitations` | `UserInvitationsUpsert` (user_invitation.go:33) | drop from `Column(...)` |
| active_sessions | `ActiveSessionUpdate` (session.go:190) | `created_at` (reset each update) | insert-only intent | `,skipupdate` on CreatedAt or `OmitZero` |

## Dismissed (verified safe)

`MembershipInvitationUpdate`, `PublicKeyUpdate`, `TagUpdate`, `SystemSet`,
`NamespaceUpdateMembership` are full-model updates whose tables have no column
touched by a separate targeted write. `APIKeyUpdate` (`OmitZero`) and
`SessionUpdate` (`OmitZero` + `ExcludeColumn` + explicit `Set`) are already
guarded. `SetupKeyUpdate` is the fix already applied.

## Suggested order

1. Apply the two clean `,skipupdate` fixes (namespaces counters, `custom_fields`).
2. Migrate the auth online transition to a targeted write, then tag `last_seen` / `disconnected_at`.
3. Handle `preferred_namespace_id` and the `invitations` allowlist entry.
4. Fix the `ActiveSessionUpdate` `created_at` reset.

A store-level (testcontainers) regression test that mutates a counter/heartbeat
column concurrently with a full-model update would lock each case, since the
mocked-store service tests cannot catch this class of bug.

otavio:
## What

Several `*Update` methods in `api/store/pg` issued full-model bun updates
(`db.NewUpdate().Model(x).WherePK()` with no allowlist) that rewrote every
mapped column from an in-memory snapshot. When a table has a column maintained
by a separate targeted write, a full-model update carrying a stale value
silently overwrote the concurrently-maintained value.

## Why

Closes #6637.

The failure is a lost update that only manifests under concurrency — a counter
increment or heartbeat landing between the `resolve` and the `save` of a
full-model update. It passes the mocked-store service tests and is easy to miss
in review.

## Changes

- **entity `,skipupdate` tags**: keep a column out of every generated UPDATE
  while still allowing INSERT and explicit `.Set(...)` writes. Applied to the
  four namespace device counters, device `custom_fields` / `last_seen` /
  `disconnected_at`, user `preferred_namespace_id`, `active_sessions.created_at`,
  and the `user_invitations.invitations` counter. Using the entity tag (rather
  than a per-method column allowlist) means every current and future update path
  is protected by one declaration.
- **device liveness → targeted writes**: `last_seen`/`disconnected_at` are now
  owned by `DeviceHeartbeat`. The agent-auth online transition routes through
  `DeviceHeartbeat`, and `OfflineDevice` uses a new targeted `DeviceOffline`
  store method instead of a full-model update (which also drops a redundant
  `DeviceResolve`).
- **preferred namespace → targeted write**: login persists it through a new
  `UserUpdatePreferredNamespace` store method; the membership-removal clear in
  `member.go` is now a single conditional `Set(... = NULL)` instead of a
  select-then-model-update.
- **invitations allowlist**: dropped from `UserInvitationUpdate`'s `Column(...)`
  list (now redundant with the entity tag).
- **regression tests**: `api/store/storetest/clobber_tests.go` reproduces each
  clobber by interleaving a targeted write with a stale full-model update and
  asserting the targeted column survives.

## Testing

`go test ./store/pg/... ./services/...` inside the api container. The new
store-level tests need testcontainers (a real PostgreSQL), which the
mocked-store service tests cannot substitute for this class of bug. Removing any
one `,skipupdate` tag makes the corresponding regression test fail.


otavio:
Addressed in ffad49ce7:

1. **`LeaveNamespace` missed clobber site** — routed the preferred-namespace clear through `UserUpdatePreferredNamespace(ctx, req.UserID, "")` so it actually clears now that the column is `skipupdate`.
2. **Dead `CreateUserToken` assignment** — removed the leftover `user.Preferences.PreferredNamespace = tenantID`, mirroring the `AuthLocalUser` cleanup.
3. **Cloud mock regeneration** — a paired `cloud` PR regenerating `pkg/store/mocks/mock_store.go` (adding `DeviceOffline` / `UserUpdatePreferredNamespace`) is being prepared.

PR #6687, opened 2026-07-20
fix(ui): use theme tokens in announcement prose styles

luizhf42:
## What

Announcement content (editor, admin preview, and user-facing modal) was unreadable in light mode because the ProseMirror prose styles used hardcoded dark-theme hex values.

## Why

Both `AnnouncementEditor.css` and `AnnouncementModal.css` hardcoded colors like `#e1e4ea` (text), `#1e2127` (code background), and `#2c2f36` (borders) — values that only work against the dark background. The design system already provides theme-aware CSS custom properties (`--c-text-primary`, `--c-border`, `--c-surface`, `--c-text-muted`) that switch automatically between light and dark mode.

## Changes

- **Prose styles**: replaced all hardcoded hex values with existing theme CSS variables (`--c-text-primary`, `--c-border`, `--c-surface`, `--c-text-muted`, `--color-primary-rgb`), making the rendered markdown adapt to both themes
- **Deduplication**: consolidated two near-identical CSS files (`AnnouncementEditor.css`, `AnnouncementModal.css`) into a single shared `announcement-prose.css` under `src/styles/`, used by all three consumers (`AnnouncementEditor`, `AnnouncementContent`, `AnnouncementModal`)
- **Editor-specific layout**: moved `min-height` and `padding` (only needed by the editable variant) to Tailwind arbitrary selectors on the wrapper div instead of baking them into the shared stylesheet


luizhf42:
> * **`ui/apps/console/src/components/announcements/__tests__/AnnouncementModal.test.tsx:9`** — `vi.mock("../AnnouncementModal.css", () => ({}))` still targets the file deleted in this PR. `AnnouncementModal.tsx` now imports `@/styles/announcement-prose.css`, so the intended "silence the CSS import" no longer points at the imported file. This does **not** break the suite (Vitest's default `css: false` already no-ops CSS imports — cf. `SessionPlayer.test.tsx`, which leaves its `.css` import unmocked — and this test fully mocks `@tiptap/react`, so `.announcement-prose` never renders). It's just a stale line worth updating or removing

Fixed. Removed the stale line from the test file.

PR #6691, opened 2026-07-21
feat: give code-paired devices their own install-key source

gustavosbarreto:
Devices enrolled through the tenant-less pairing-code flow (the agent
prints a code, the user accepts it in the console) were attributed to the
namespace's legacy install key. They landed in the legacy key's
registration history alongside genuinely keyless devices, so there was no
way to tell a code-paired device from one that enrolled with only a tenant
ID. This gives code pairing its own source.

**Backend**

Install keys are now discriminated by a `type` column (`user`, `legacy`,
`pairing`) instead of the `system` boolean plus name matching. Every
namespace has two auto-managed system keys: `legacy` for tenant-only
keyless enrollment, and `pairing` for code-paired devices. A unique partial
index keeps exactly one of each per namespace. Migration 015 backfills the
pairing key for existing namespaces; new namespaces get it at the same
creation chokepoint as the legacy key.

The pairing flow is its own acceptance, so a paired enrollment accepts
outright and ignores the pairing key's mode. It never fires the webhook,
the allowlist reject, or a manual hold, any of which would hang the agent
or misreport in the UI. The pairing key is not editable and has no
presentable secret.

Existing pairing devices already stamped with the legacy key are
byte-identical to keyless-legacy devices and cannot be re-attributed
reliably, so the migration leaves them as-is (fix-forward).

**UI**

The install keys list splits the two auto-managed keys into a "Built-in"
section, apart from the admin's "Custom keys". Pairing reads as "Code
pairing", not a mode.

The built-in keys always fill the table, so the old full-page onboarding
hero never showed. It is replaced by an in-section placeholder: with no
custom keys, the Custom keys section renders a "Create your first key"
block under the built-in rows. A key's activity page shows its source (the
install key for user and legacy keys, "Pairing code" for paired devices).

### DIFF
--- a/api/services/auth.go
+++ b/api/services/auth.go
@@ -148,19 +148,31 @@ func (s *service) appendInstallKeyEvent(ctx context.Context, key *models.Install
 }
 
 // enrollmentInstallKey resolves the install key for a fresh enrollment. With a presented key it
-// validates it, rejecting an invalid or system key. With no key it resolves the namespace's legacy
-// (system) key, so a keyless enrollment is governed by the legacy key's (manual) mode. It returns the
-// resolved key (nil only when no legacy key exists), the digest to store on the device, and an error
-// only when a presented key is invalid.
-func (s *service) enrollmentInstallKey(ctx context.Context, req requests.DeviceAuth) (*models.InstallKey, string, error) {
+// validates it, rejecting an invalid or system key. With no key it resolves a system key: the pairing
+// key when the enrollment comes from the pairing-code flow (paired), otherwise the namespace's legacy
+// key, so a tenant-only keyless enrollment is governed by the legacy key's (manual) mode. It returns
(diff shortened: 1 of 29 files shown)


gustavosbarreto:
Thanks. On the migration 015 note: you're right that the legacy-reservation argument doesn't extend to `pairing`. I've made it a deliberate, documented fix-forward rather than mutating user data (d86383ea0): a namespace that already holds a user key named `pairing` keeps it and skips its pairing system key, code pairing there still works (accepted explicitly by `acceptPairingDevice`), it only loses pairing-source attribution in that rare case, and new namespaces are unaffected. Renaming a user's key silently in a migration felt like a worse surprise than the documented gap.

The cloud `MockStore` regen and the `installKey.yaml` `system`→`type` change are both expected, as you noted.

PR #6715, opened 2026-07-22
refactor(ui): remove unnecessary DS re-exports from website

luizhf42:
## What

All website consumers now import `Reveal`, `ConnectionGrid`, `ShimmerCard`, `C`, `FONT_SANS`, and `FONT_MONO` directly from `@shellhub/design-system` instead of through two landing-page proxy files that were pure re-exports with zero added value. Both proxy files are deleted.

## Why

`pages/landing/constants.ts` and `pages/landing/components.tsx` re-exported design-system symbols verbatim — no aliases, no transformations, no local additions. This made it look like these were landing-page-specific symbols when they're shared DS primitives, and added a layer of indirection that obscured the true dependency origin.

Closes shellhub-io/team#184

## Changes

- **30 consumer files**: rewired imports from `./constants`, `./components`, `../landing/constants`, `../landing/components` to the canonical `@shellhub/design-system/{constants,components}` paths. Files that already had a partial DS import (e.g. `GlowOrbs`) got the new symbols merged into the existing import statement.
- **`pages/landing/constants.ts`**: deleted
- **`pages/landing/components.tsx`**: deleted

### DIFF
--- a/ui/apps/website/src/pages/enterprise/AdminPanel.tsx
+++ b/ui/apps/website/src/pages/enterprise/AdminPanel.tsx
@@ -1,7 +1,7 @@
 import { CheckIcon } from "@heroicons/react/24/outline";
 import { WindowChrome } from "@shellhub/design-system/primitives";
 import { Section, SectionHeader } from "@/components/marketing";
-import { Reveal, ShimmerCard } from "../landing/components";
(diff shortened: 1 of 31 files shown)


otavio:
rebase

PR #6760, opened 2026-07-29
fix(storetest): stop CreateDevice generating colliding MAC addresses

### LINKED ISSUE #6617: Flaky store tests: CreateDevice test helper generates colliding MACs under the accepted-mac unique index
## Summary

`TestPgStore/StatsStore` (and any store test that creates more than one accepted
device in the same namespace) is flaky. It fails intermittently with:

```
--- FAIL: TestPgStore/StatsStore/succeeds_with_specific_tenantID
    stats_tests.go:49:
        Error Trace: api/store/storetest/helpers.go:226
                     api/store/storetest/stats_tests.go:49
        Error: Received unexpected error:
               document duplicate
```

Seen on CI: https://github.com/shellhub-io/shellhub/actions/runs/28873799116/job/85643304102
A plain re-run of the same commit goes green, which is the tell-tale of a flake.

## Root cause

`CreateDevice` in `api/store/storetest/helpers.go` builds the default device MAC
from the wall clock:

```go
Identity: &models.DeviceIdentity{MAC: fmt.Sprintf(
    "%02x:%02x:%02x:%02x:%02x:%02x",
    time.Now().UnixNano()%256, time.Now().UnixNano()%256, ...)},
```

On a runner whose clock resolution is coarser than the six `time.Now()` calls,
all six octets collapse to the same value, so the whole MAC becomes
`VV:VV:VV:VV:VV:VV`. Two devices created back to back then collide on their MAC
with probability ~1/256 per pair.

That collision was harmless until migration `007_devices_accepted_mac_unique`
(#6600) added a unique index on `(namespace_id, mac) WHERE status = 'accepted'`.
Now a colliding MAC across two accepted devices in the same namespace turns into
a `document duplicate` on insert. `StatsStore/succeeds_with_specific_tenantID`
creates three accepted devices in `tenant1`, so it is the most frequent victim.

The default UID (`fmt.Sprintf("%064x", time.Now().UnixNano())`) has the same
weakness: two devices created within one clock tick collide on the primary key.

## Suggested fix

Give the test helper collision-free defaults instead of deriving them from the
clock. A package-level `atomic.Uint64` counter that feeds both the UID and the
MAC is enough:

```go
var deviceSeq atomic.Uint64

n := deviceSeq.Add(1)
uid := fmt.Sprintf("%064x", n)
mac := fmt.Sprintf("00:00:%02x:%02x:%02x:%02x", byte(n>>24), byte(n>>16), byte(n>>8), byte(n))
```

Callers that need a specific MAC already pass one through a `DeviceOption`, so
only the default changes. The other clock-derived defaults in this file
(namespace name, user email, session UID) are candidates for the same treatment
if they start colliding under their own unique constraints.

## Notes

This is a pre-existing issue in the test helper on `master`, exposed by the
#6600 unique index. It is not tied to any one feature branch; it can fail any PR
whose run happens to hit the collision.

geovannewashington:
## What

CreateDevice, the shared test helper used across the store test
suite, generated both the device UID and MAC address from
time.Now().UnixNano(). This change replaces that with a package-level
atomic counter so generated identifiers are guaranteed unique.

## Why

Devices are unique on (namespace_id, mac) while accepted. On machines
with coarse clock resolution, two calls to CreateDevice within the
same test run could land on the same nanosecond value and produce
identical MAC addresses, which then collided against that unique
constraint. This made the store suite intermittently fail with no
code change involved, which is exactly what made it hard to track
down.

Closes shellhub-io/shellhub#6617

## Changes

- Added a package-level atomic.Uint64 counter, deviceSeq, in
  storetest/helpers.go
- Added nextDeviceUID and nextDeviceMAC, both derived from the
  counter instead of the wall clock
- Wired CreateDevice to use these helpers instead of time.Now()
- Added helpers_test.go covering the new helpers directly

## Testing

Flakiness from clock collisions is hard to exercise on demand, since
it depended on timing that isn't reproducible on request. Instead,
this validates that the fix is correct by construction and does not
regress existing behavior:

- go test ./store/storetest/... passes, including the new
  helpers_test.go cases asserting nextDeviceUID and nextDeviceMAC
  never repeat across repeated calls
- go test ./store/pg/... passes, confirming the rest of the store
  suite still works against the new identifiers
- golangci-lint run ./store/... passes, including gosec, since the
  MAC byte extraction uses bit-masking rather than a truncating
  conversion

### DIFF
--- a/server/api/store/storetest/helpers.go
+++ b/server/api/store/storetest/helpers.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -194,6 +195,21 @@ func WithDeviceRemoteAddr(addr string) DeviceOption {
 	}
 }
 
+// deviceSeq backs the default device identifiers. A counter rather than the clock because
+// devices are unique on (namespace_id, mac) while accepted, and clock-derived values collide
+// often enough to fail the suite.
+var deviceSeq atomic.Uint64
+
+func nextDeviceUID() string {
+	return fmt.Sprintf("%064x", deviceSeq.Add(1))
+}
+
+func nextDeviceMAC() string {
+	n := deviceSeq.Add(1)
+
+	return fmt.Sprintf("00:00:%02x:%02x:%02x:%02x", (n>>24)&0xFF, (n>>16)&0xFF, (n>>8)&0xFF, n&0xFF)
+}
+
 // CreateDevice creates a device with default or customized values
 // Returns the generated device UID
 // If tenant is not provided via WithTenantID(), a default namespace will be created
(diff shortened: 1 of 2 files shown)


otavio:
@geovannewashington, please do a rebase of this pull request, as Gustavo has merged the servers consolidation pull request.

PR #6765, opened 2026-07-30
fix(server): declare a utf-8 character locale on web terminal sessions

### LINKED ISSUE #6763: When accessing the device terminal and running the ls command, Chinese directory names are displayed incorrectly due to character encoding issues.
### Describe the issue

When accessing the device terminal and running the ls command, Chinese directory names are displayed incorrectly due to character encoding issues.

otavio:
## What

The web terminal now declares `LC_CTYPE=C.UTF-8` to the device before starting the
shell, so tools that adapt their output to the locale (`ls`, `find`, `tree`, `busybox`)
print non-ASCII filenames as characters instead of octal escapes.

## Why

The web terminal was the only ShellHub client that sent no locale at session setup. The
device's shell therefore ran in the `C`/POSIX locale, where every byte >= 0x80 is
non-printable; GNU `ls` writing to a TTY escapes those bytes, producing output like
`''$'\344\270\255\346\226\207\347\233\256\345\275\225'` for `中文目录`. Connecting to
the same device with a native `ssh` client renders the names correctly, because native
clients already forward `LANG`/`LC_*`.

This only shows on devices whose own shell profile does not already set a UTF-8 locale —
the normal state for the embedded and IoT targets ShellHub is built for, which is why it
looked intermittent.

Closes #6763

## Changes

- **server/ssh/web**: send `LC_CTYPE=C.UTF-8` as an SSH `env` request before the shell
  request. `LC_CTYPE` governs character classification only and outranks `LANG`, so a
  device profile setting `LANG=C` cannot silently undo it; setting `LANG` instead would
  overreach and be overridable. `C.UTF-8` requires no generated locale data (present on
  glibc >= 2.35, always available on musl, satisfies BusyBox's UTF-8 check) — a
  language-specific locale such as `zh_CN.UTF-8` may be absent, and glibc then falls
  back to `C`, reintroducing the bug on exactly the devices this fixes. The value is
  hardcoded: the web terminal can only render UTF-8, so there is no coherent alternative
  to offer.
- **server/ssh/web**: session bring-up moved out of `newSession` into `prepareShell`
  over a narrow three-method interface (`Setenv`, `RequestPty`, `Shell`). `newSession`
  cannot be tested — it dials a fixed address and needs the cache and internal API
  client — and the property worth enforcing is an ordering one: the agent only records
  env received *before* the shell request, so "env then shell" is the contract, not
  merely "env was sent".
- **Failure handling**: a device or intermediary rejecting the `env` request logs at
  debug level and continues; the user simply gets today's behaviour rather than a broken
  session. A failing PTY or shell request still surfaces as `ErrPty`/`ErrShell`.

No agent or gateway change was needed. The gateway already proxies client channel
requests through generically, and the agent already accepts `LANG` and any `LC_*`
(`agent/server/modes/host/env.go`), mirroring OpenSSH's default `AcceptEnv`. Native
clients are unaffected.

## Testing

`TestPrepareShell` drives a fake that records bring-up calls in order, covering the
ordering contract and the non-fatal env failure. Verify by hand on a device with GNU
coreutils and no locale of its own — on a dev agent whose profile sets `LANG`, prefix
with `env -u LANG -u LC_ALL -u LC_CTYPE` to put the shell in the state a bare embedded
device is already in, then run `ls` over a directory holding non-ASCII names.

A device that emits a non-UTF-8 encoding (GBK/GB18030) still cannot render, because the
output relay replaces invalid UTF-8 with U+FFFD server-side. That is a second,
independent defect and ships separately as Part 2 of the spec on #6763.

### DIFF
--- a/server/ssh/web/session.go
+++ b/server/ssh/web/session.go
@@ -283,20 +283,8 @@ func newSession(ctx context.Context, service services.Service, handoff *webhando
 		return err
 	}
 
-	if err := agent.RequestPty("xterm", int(dim.Rows), int(dim.Cols), ssh.TerminalModes{
-		ssh.ECHO:          1,
-		ssh.TTY_OP_ISPEED: 14400,
-		ssh.TTY_OP_OSPEED: 14400,
-	}); err != nil {
-		logger.WithError(err).Debug("failed to request the pty on session")
-
-		return ErrPty
-	}
-
-	if err := agent.Shell(); err != nil {
-		logger.WithError(err).Debug("failed to request the shell on session")
-
-		return ErrShell
+	if err := prepareShell(logger, agent, dim); err != nil {
+		return err
 	}
 
 	go func() {
--- a/server/ssh/web/session.go
+++ b/server/ssh/web/session.go
@@ -315,7 +315,7 @@ func newSession(ctx context.Context, service services.Service, handoff *webhando
 			case messageKindResize:
 				dim := message.Data.(Dimensions)
 
-				if err := agent.WindowChange(int(dim.Rows), int(dim.Cols)); err != nil {
+				if err := agent.WindowChange(dim.rows(), dim.cols()); err != nil {
 					logger.WithError(err).Error("failed to change the size of window for terminal session")
 
 					return
--- a/server/ssh/web/shell.go
+++ b/server/ssh/web/shell.go
@@ -1,0 +1,55 @@
+package web
+
+import (
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ssh"
+)
+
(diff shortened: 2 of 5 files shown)


gustavosbarreto:
Diagnosis is right, the cause is locale.

#6772 does it in the agent: reads `LANG` and `LC_*` from the files a login reads, falls
back to `LANG=C.UTF-8` when the host configures none. Covers native clients, which the
gateway path cannot reach, with a before/after through the web terminal.

Note either way: the env request hits the channel proxy's generic `default` branch, so it
goes through `sess.Event` and adds one `sessions_events` row per web session.

PR #6766, opened 2026-07-30
fix(server,ui): send device output to the browser verbatim

otavio:
## What

The gateway stops rewriting the device's output. Whatever bytes the device wrote, the
browser receives; the terminal decodes them, with a user-selectable character encoding
(UTF-8 by default).

Part 2 of the spec on #6763. Part 1 (announcing `LC_CTYPE=C.UTF-8`, which closes that
issue) ships separately in #6765. The two are independent — this one is the reason a
GBK device can never work today, not what the original reporter hit.

## Why

`redirToWs` decoded the device's bytes to runes and re-encoded them, substituting U+FFFD
for anything that was not valid UTF-8. A device emitting GBK/GB18030 — common on Chinese
systems, and readable by a native client with a matching terminal encoding — could
therefore never be displayed correctly, because the information was destroyed
server-side. Running the old relay against controlled read boundaries:

| input | result |
|---|---|
| UTF-8 split mid-character across reads | preserved |
| GBK `中文目录` (`d6 d0 ce c4 …`) | corrupted → `ef bf bd ef bf bd …` |
| Latin-1 high bytes (`41 e9 e8 42`) | corrupted → `41 ef bf bd ef bf bd 42` |
| a chunk of only continuation bytes | dropped entirely |

Related: #6763

## Changes

- **server/ssh/web**: `redirToWs` deleted. stdout and stderr are copied verbatim into
  binary frames. The hand-rolled scan that kept a multi-byte sequence off a frame
  boundary is no longer needed, because the client's decoder is stateful across writes —
  and with it go two latent bugs it contained: a fallback branch that used the wrong
  length after a partial sequence had been carried over, and a path that silently dropped
  data when a buffer began with continuation bytes. Neither is expressible in a straight
  copy.
- **server/ssh/web**: stderr no longer goes out as a *text* frame. It was relayed with a
  plain copy onto the socket, concurrently with stdout's binary frames; RFC 6455 requires
  text frames to be valid UTF-8, so a sequence split across a stderr read was a protocol
  violation the browser is entitled to fail the connection over. Both streams now share
  one binary-frame writer.
- **server/ssh/web**: write serialization moved onto `Conn`, not onto that writer.
  Keep-alive pings and JSON control messages are produced by other goroutines and write
  to the same `x/net/websocket` frame writer, which is not goroutine-safe; guarding only
  the output path leaves the corruption it set out to prevent. `TestConnConcurrentWritesDoNotRace`
  drives output, control messages and pings concurrently over a real socket and observes
  genuinely corrupted frames without the lock — not merely a race warning.
- **ui (terminal)**: the socket takes `arraybuffer` and the message handler is
  synchronous. It previously awaited `Blob.text()` per frame, which decoded each frame in
  isolation (corrupting anything spanning frames), let frame order follow promise
  resolution rather than arrival, and left the cancellation check stale across the await.
  Control frames stay JSON text and are discriminated by `typeof event.data === "string"`,
  which keeps the existing non-JSON-text fallback intact.
- **ui (`terminalOutputDecoder`)**: a per-session streaming `TextDecoder`. A character
  split across frames is held and completed for *every* encoding, not just UTF-8. An
  unsupported encoding name falls back to UTF-8 rather than throwing. It always yields a
  string: the recorder needs one regardless (asciicast v2 output events are JSON strings
  by specification), so feeding the terminal and the recorder the same string is both
  simpler and what guarantees a recording replays as what the user saw.
- **ui (settings)**: character encoding is a persisted terminal preference alongside the
  existing appearance settings. Changing it applies to sessions opened afterwards — an
  open session's decoder holds stream state, so its scrollback is not retroactively
  re-decoded.

Invalid bytes still render as U+FFFD, but now in the browser, as a rendering decision,
after the bytes have survived the trip. That is what makes selecting GBK able to work.

## Testing

The server tests assert byte-for-byte fidelity through scripted read boundaries rather
than on frame counts or buffer sizes — the implementation details whose removal is the
point. `TestOutputWriterPreservesBytes` places a boundary mid-character and covers GBK,
Latin-1, all-continuation-bytes, a zero-byte read followed by data, and arbitrary binary;
the two pathologies pinned by the deleted `TestRedirToWs_Regression_*` tests are carried
over as explicit cases rather than dropped. `TestOutputWriterSendsBinaryFrames` pins the
frame opcode over a real connection, which the package's socket mock cannot show.

Run the server package with `-race` — `TestConnConcurrentWritesDoNotRace` is the gate on
the locking, and it fails loudly if the lock is narrowed back to the output writer.

Worth probing by hand: a device emitting GBK filenames with the encoding set to GBK, and
a `hexdump`/binary-heavy command to confirm the browser receives exactly what the device
wrote. The terminal component itself stays untested — it is thin wiring (frame-type
branch, decode, write, record), and testing it would mean introducing WebSocket, OPFS and
WebGL shims this app does not have.

### DIFF
--- a/server/ssh/web/conn.go
+++ b/server/ssh/web/conn.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"sync"
 	"time"
 	"unicode/utf8"
 
@@ -21,6 +22,12 @@ type Conn struct {
 	Socket Socket
 	// Pinger is reponsable to inform the server that a SSH session is open.
 	Pinger *time.Ticker
+
+	// writes serializes every frame written to the socket. Device output, JSON
+	// control messages and keep-alive pings are produced by different
+	// goroutines, and the frame writer underneath is not goroutine-safe: two
+	// concurrent writers otherwise interleave and corrupt a frame.
+	writes sync.Mutex
 }
 
 func NewConn(socket Socket) *Conn {
@@ -116,6 +123,9 @@ func (c *Conn) ReadMessage(message *Message) (int, error) {
 }
 
 func (c *Conn) WriteMessage(message *Message) (int, error) {
+	c.writes.Lock()
+	defer c.writes.Unlock()
+
 	buffer, err := json.Marshal(message)
 	if err != nil {
 		return 0, errors.Join(ErrConnReadMessageJSONInvalid)
@@ -130,6 +140,9 @@ func (c *Conn) WriteMessage(message *Message) (int, error) {
 }
 
 func (c *Conn) WriteBinary(data []byte) (int, error) {
+	c.writes.Lock()
+	defer c.writes.Unlock()
+
 	socket, ok := c.Socket.(*websocket.Conn)
 	if !ok {
 		// NOTE: If the underlying connection is not a websocket connection, fallback to a normal write.
@@ -151,11 +164,35 @@ func (c *Conn) WriteBinary(data []byte) (int, error) {
 	return wrote, nil
 }
 
+func (c *Conn) WritePing() error {
+	c.writes.Lock()
+	defer c.writes.Unlock()
+
+	socket, ok := c.Socket.(*websocket.Conn)
+	if !ok {
+		return nil
+	}
+
+	frame, err := socket.NewFrameWriter(websocket.PingFrame)
+	if err != nil {
+		return err
+	}
+
+	if _, err := frame.Write([]byte{}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (c *Conn) Read(buffer []byte) (int, error) {
 	return c.Socket.Read(buffer)
 }
 
 func (c *Conn) Write(buffer []byte) (int, error) {
+	c.writes.Lock()
+	defer c.writes.Unlock()
+
 	return c.Socket.Write(buffer)
 }
 
@@ -176,9 +213,7 @@ func (c *Conn) KeepAlive() {
 			return
 		}
 
-		if fw, err := socket.NewFrameWriter(websocket.PingFrame); err != nil {
-			return
-		} else if _, err = fw.Write([]byte{}); err != nil {
+		if err := c.WritePing(); err != nil {
 			return
 		}
 
--- a/server/ssh/web/conn_test.go
+++ b/server/ssh/web/conn_test.go
@@ -1,13 +1,19 @@
 package web
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
+	"net/http/httptest"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/shellhub-io/shellhub/server/ssh/web/mocks"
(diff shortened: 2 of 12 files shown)


gustavosbarreto:
The `Conn` write race is real: ping frames, JSON control messages and output share a frame
writer that is not goroutine safe. Unrelated to locale, changes no frame format, worth its
own PR now.

Two blockers on the rest.

**Wire format.** `/ws/ssh` is not in the OpenAPI spec and the frame format was never
published. Integrations were written against observed behaviour: binary frames always
valid UTF-8, never splitting a character. Per-frame decoding is the obvious client
implementation and would silently yield replacement characters. No version to bump, no way
to notify. Endpoint version, not a bugfix.

**The client bugs need no server change.** `Terminal.write()` takes `Uint8Array` and
decodes UTF-8 statefully across calls (`_utf8Decoder: Utf8ToUtf32`, 3 byte `interim`, one
per Terminal). `binaryType = "arraybuffer"`, synchronous handler, bytes to `term.write`:
split characters and ordering both fixed, wire untouched. The recorder keeps a per session
`TextDecoder({stream: true})` for the asciicast.

The encoding selector has no case left. #6763 is not a non-UTF-8 case, the escapes there
are `\344\270\255\346\226\207`, valid UTF-8 for `中文`, and #6772 fixes it. Independently
of that, the input path is UTF-8 only by construction: `ReadMessage` counts runes with
`utf8.RuneCountInString` and `CharacterSize` is 4. Selecting GBK would read GBK and type
UTF-8, so a filename created in that session comes out corrupted. The selector and its
store can go.

Split: mutex on its own, client fixes as a UI PR, locale in #6772.

otavio:
Checked every point and they all hold:

- **The `Conn` race is real.** `WriteMessage` writes through `Socket.Write`, `WriteBinary` through a `BinaryFrame` writer and `KeepAlive` through a `PingFrame` writer, from three different goroutines onto the same non-goroutine-safe writer. Driving all three concurrently over a real socket reports a data race *and* desyncs the frame parser outright.
- **`/ws/ssh` is absent from the OpenAPI spec** — no match in `openapi/spec/`. Nothing to version, agreed.
- **The client bugs need no server change.** `@xterm/xterm@6` types `write(data: string | Uint8Array)` and the bundle carries `this.interim = new Uint8Array(3)`, so the UTF-8 decode is stateful across `write` calls exactly as you describe. Bytes to `term.write` on a synchronous handler fixes both the split characters and the ordering with the wire untouched.
- **The encoding selector has no case left.** Beyond #6763 not being a non-UTF-8 case, the input path settles it: `ReadMessage` counts with `utf8.RuneCountInString` and `CharacterSize` is 4, so selecting GBK would read GBK and type UTF-8 — a filename created in that session would come out corrupted. Selector and store dropped.

Split as suggested:

- #6829 — the `Conn` mutex on its own.
- #6830 — the client fixes, UI only.
- #6772 — the locale, merged.

Closing this one; it is superseded by those three.

PR #6815, opened 2026-07-31
ci: group Dependabot updates instead of one pull request per module

gustavosbarreto:
## What

Collapses `.github/dependabot.yml` from 11 update entries to 5, grouping updates that belong together and putting an explicit ceiling on how many pull requests each entry may open.

## Why

Pointing Dependabot at the modules that actually exist (`e2d091c13`) released a backlog of 24 open pull requests in one day. Draining that batch showed the file's real problems, and all of them are structural rather than one-off.

The six `gomod` entries treated the modules as independent, which they are not. They share a workspace, so `go mod tidy` on any one of them rewrites the others' `go.sum`. A dependency several modules have in common therefore arrived as one pull request per module, each carrying the same edits and each conflicting with the rest once any one merged. `testcontainers` came as four pull requests, `go-envconfig` as two, and four of the batch had to be regenerated through `@dependabot recreate` after unrelated merges.

`react` and `react-dom` arrived separately, and React refuses to run unless the two are on the exact same version. Each half fails CI on its own, so neither could ever go green. The react-only half was closed rather than merged.

No entry set `open-pull-requests-limit`, which defaults to 5 and applies per entry rather than per file.

## Changes

- **gomod**: one entry over all six module directories, with `group-by: dependency-name` so a shared dependency produces a single pull request. The `caddy` group is kept and excluded from the catch-all, since a dependency matching more than one group is not placed in the first one it matches.
- **npm**: a `react` group covering `react`, `react-dom` and their `@types` packages, plus a catch-all for minor and patch updates that excludes the named groups.
- **docker**: the four Go images share a base, so they are grouped by dependency name for the same reason as the Go modules.
- **github-actions**: a single group replaces the `codeql-action`-specific one.
- **limits**: every entry now states `open-pull-requests-limit`.
- Catch-all groups carry `applies-to: version-updates`, so a security fix arrives on its own instead of waiting behind a broken neighbour in the same pull request.

## Notes

Go bumps now carry a `deps` commit scope rather than `server`, `gateway` or `pkg`. An update that spans the workspace has no single module to name.

`group-by: dependency-name` is generally available on github.com (`fpt` and `ghec`, and GHES 3.21 onward) and is documented as creating "a single pull request for each dependency update across all specified directories, rather than separate pull requests per directory". Its documented limitations are satisfied here: both entries keep a single package ecosystem, and the grouping applies to version updates only. Where directories carry incompatible version constraints for the same dependency, Dependabot falls back to separate pull requests rather than stalling.

It has not run against this repository yet, so the first grouped run is worth confirming on the repository's Dependabot page.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 
+# Every entry sets open-pull-requests-limit explicitly. The default is 5 per entry,
+# which multiplies by the number of entries here and lets a quiet week turn into
+# dozens of simultaneous pull requests.
+
 updates:
 - package-ecosystem: github-actions
   directory: "/"
@@ -7,10 +11,14 @@ updates:
     interval: weekly
   cooldown:
     default-days: 7
+  open-pull-requests-limit: 5
   groups:
-    codeql:
+    # applies-to keeps this to version updates, so a security fix still arrives on
+    # its own instead of waiting behind an unrelated action in the same pull request.
+    actions:
+      applies-to: version-updates
       patterns:
-        - "github/codeql-action/*"
+        - "*"
 
 - package-ecosystem: docker
   directories:
@@ -24,6 +32,15 @@ updates:
     default-days: 7
   commit-message:
     prefix: "docker"
+  open-pull-requests-limit: 5
+  groups:
+    # These four share the same base image, so without group-by a single bump
+    # arrives as four pull requests that all edit the same version.
+    images:
+      applies-to: version-updates
+      group-by: dependency-name
+      patterns:
+        - "*"
   # The Go image follows the toolchain pinned in go.mod, so it only moves when we
   # bump the toolchain ourselves.
   ignore:
@@ -37,6 +54,7 @@ updates:
     default-days: 7
   commit-message:
     prefix: "docker"
+  open-pull-requests-limit: 5
   ignore:
     - dependency-name: "node"
       update-types: ["version-update:semver-major"]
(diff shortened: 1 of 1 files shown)


gustavosbarreto:
Good catch to push on this rather than take the word for it, since an unrecognized key rejects the entry outright and Go updates would stop silently. Checked it against the docs source rather than a rendered page.

`group-by` is a real group key, documented in `content/code-security/reference/supply-chain-security/dependabot-options-reference.md` in `github/docs`:

> ### `group-by` (`groups`)
>
> Use `groups.<group-name>.group-by` to specify how Dependabot should group updates across multiple directories in a monorepo.
>
> * **Type:** String
> * **Accepted values:** `dependency-name`
> * **Applies to:** Configurations with multiple directories specified

It is version-gated in the docs, and the gate resolves to generally available on github.com — `data/features/dependabot-updates-group-by.yml`:

```yaml
# Reference: Issue #20890 - Dependabot can group updates by dependency name across multiple directories in a monorepo [GA]
versions:
  fpt: '*'
  ghec: '*'
  ghes: '>= 3.21'
```

The same section lists three limitations, all of which this configuration satisfies or tolerates:

- all directories must use the same package ecosystem, which holds for both the `gomod` and the `docker` entry
- it applies to version updates only, which is what the group already declares
- **if directories have incompatible version constraints for a dependency, Dependabot creates separate pull requests**

That last one is the part that matters for the risk you raised: the failure mode of the grouping itself is a fall back to the previous per-directory behaviour, not a stall.

Keeping your suggestion to confirm the first run on the repository's Dependabot page after merge, and updating the note in the description, which was more pessimistic than the evidence supports.

PR #6838, opened 2026-08-02
fix(gateway): mark the seat as having a pty only once the device says so

gustavosbarreto:
The seat is marked as having a pty on the way in, before the request has even
reached the device, and nothing rolls it back when the device refuses. A device
can refuse: no `/dev/ptmx` in a restricted container is enough, and OpenSSH does
the same thing in that situation — reply `SSH_MSG_CHANNEL_FAILURE` and let the
session carry on without a terminal.

A seat that claims a pty it does not have defeats two guards written for exactly
this case.

The namespace announcement is written straight into the client channel, guarded
by `seat.HasPty` precisely so it stays out of a non-tty stdout. With the seat
stale, a client piping the output gets `Connected to ... via ShellHub` in the
middle of whatever it was reading.

On Cloud and Enterprise the session is flagged as recorded, and `pty-output`
events are stored for a session that has no terminal. The guard at
`session.go` that reports `session won't be recorded because there is no pty`
never fires, so the recording is fed raw pipe bytes where it expects terminal
output.

The parsed request is now carried down to where the agent's answer arrives, and
the seat is marked there. The refusal is also logged, which it was not on this
side before.

## Reachability

Not reachable today: the agent runs on the library's emulated pty, which never
fails, so `pty-req` is always accepted. It becomes reachable as soon as the
agent allocates a real pty, which is why this lands first and on its own.

## Verification

Exercised end to end by forcing the device to refuse a pty, which produced the
two log lines that confirm the guards now hold, and the announcement stayed out
of the stream.

Ten released agent versions, v0.17.2 through v0.26.0, were driven through shell,
exec with and without a pty, heredoc, sftp and scp against this change. No
device refused a pty and nothing regressed.

### DIFF
--- a/server/ssh/server/channels/session.go
+++ b/server/ssh/server/channels/session.go
@@ -338,6 +338,7 @@ func DefaultSessionHandler() gliderssh.ChannelHandler {
 					ok, err := agent.Channel.SendRequest(req.Type, req.WantReply, req.Payload)
 					if err != nil {
 						logger.WithError(err).Error("failed to send the request from client to agent")
+						ptyRequested = nil
 
 						continue
 					}
@@ -349,7 +350,10 @@ func DefaultSessionHandler() gliderssh.ChannelHandler {
 					}
 
 					if ptyRequested != nil {
-						if ok {
+						// SendRequest reports false when the client asked for no
+						// reply, so only a client that wanted one tells us
+						// anything about the device's answer.
+						if ok || !req.WantReply {
 							sess.Seats.SetPty(seat, true)
 							sess.Event(PtyRequestType, *ptyRequested, seat) //nolint:errcheck
 						} else {
--- a/server/ssh/server/channels/session.go
+++ b/server/ssh/server/channels/session.go
@@ -231,6 +231,10 @@ func DefaultSessionHandler() gliderssh.ChannelHandler {
 		go func() {
 			defer wg.Done()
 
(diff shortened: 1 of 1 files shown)


gustavosbarreto:
Good catch on the `want_reply` gating — that was a real regression. `SendRequest`
returns false when no reply was asked for, which is the library saying there was
nothing to wait for, not the device refusing. Now the seat is only dropped when
the client actually asked for a reply and the device said no. Also clearing the
pending request when the forward itself errors, as suggested.

## Validation against released agents

Ten released agent versions were brought up as their published images and driven
through the gateway with this branch: shell with a pty, exec with and without
one, heredoc, sftp and scp.

| versions | result |
| --- | --- |
| v0.17.2 – v0.20.1 | all six checks pass |
| v0.21.7 – v0.26.0 | all pass except sftp |

54 of 60 checks pass, and **no device refused a pty**, so the seat was marked on
every session.

The six sftp failures are #6834: the agent looks up the user in the container's
`/etc/passwd` rather than the host's. It is already fixed and clears once a
release carries it. The older versions pass sftp because they need the host's
passwd bind-mounted to work at all, and that makes the lookup succeed too.

Worth noting for anyone testing older agents: up to v0.20.1 the agent mapped
`/host/etc/shadow` but left passwd pointing inside the container, so
authentication succeeds and the user lookup right after fails silently — the
session ends with nothing run and nothing logged. Fixed in bce4f5343.

PR #6840, opened 2026-08-02
fix(agent): report a shell's exit code and reap every session

gustavosbarreto:
A shell with a pty always reported success to the client. The agent never sent a status of its own, so the library filled in a zero. Exec and the heredoc shape both got this right, which is why the gap only showed up when you asked for a terminal. Nothing in the e2e suite covered it: four cases request a pty and then a shell, and none of them looked at the exit status.

Reaping the process now sits next to the command that started it, in all four handlers. That lets the `cmds` map go, and three defects go with it.

The map was written under the Sessioner mutex and read under the Server one, so two unrelated locks guarded the same map. Its key was the connection's exchange hash rather than the channel's, so two sessions multiplexed on one connection collided and only the second was ever killed. And a shell was registered only after utmp had been written, leaving a window in which the callback found nothing and never ran again, because it runs once per connection.

Exec installed its reaper after waiting on the command's output. A command that ignores a vanished client never closes that output, so for exactly the case the reaper exists for, it was never installed. SFTP had no reaper at all.

The fork bump carries the matching library fix: a command that wrote a burst and exited at once lost the end of it, because nothing ordered the copy out of the pty against the close of the channel. Reproducing it needs stressing, 400 runs caught it four times before the change and none after, and one of those lost 31300 of 200000 bytes.

Depends on shellhub-io/ssh@17dd395.

### DIFF
--- a/agent/go.mod
+++ b/agent/go.mod
@@ -81,4 +81,4 @@ require (
 
 replace github.com/shellhub-io/shellhub => ../
 
-replace github.com/gliderlabs/ssh => github.com/shellhub-io/ssh v0.0.0-20260802041333-0de04e2021d3
+replace github.com/gliderlabs/ssh => github.com/shellhub-io/ssh v0.0.0-20260802071412-17dd3955caac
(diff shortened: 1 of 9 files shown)


gustavosbarreto:
Both observations from the review were considered and left as-is.

The `reapOnDisconnect` error branch and `Heredoc`'s unguarded `ProcessState` are unreachable in practice: `ContextKeyConn` is set once per connection at handshake, before any channel is dispatched, and `ProcessState` is only nil when `Wait` was never reached. Changing either would trade a real behaviour for a hypothetical one, so they stay for a follow-up if they ever become reachable.

PR #6961, opened 2026-08-26
fix(agent): use `unless-stopped` restart policy in `install.sh`

### LINKED ISSUE #6958: fix(install): agent does not come back after reboot on Podman
fix(install): agent does not come back after reboot on Podman

`install.sh` creates the agent container with `--restart=on-failure`
(`install.sh:174` for Podman, `install.sh:262` for Docker). On Podman that policy
never starts the container at boot, so the device goes offline on the first reboot
and stays offline until someone logs in and starts it by hand.

## Podman

Podman has no daemon. Containers are started at boot by `podman-restart.service`,
which runs:

    podman start --all --filter should-start-on-boot=true

`should-start-on-boot` is true only for restart policy `always` or `unless-stopped`
(`man podman-ps`), so the agent container is never in that set.

Measured with `shellhubio/agent:v0.27.0-rc.3` and the flags the script uses:

    $ podman inspect shellhub-probe --format '{{.HostConfig.RestartPolicy.Name}}'
    on-failure

    $ podman ps -a --filter should-start-on-boot=true
    (empty)

    $ podman stop -t 3 shellhub-probe
    $ podman start --all --filter should-start-on-boot=true   # what the unit runs at boot
    (exit 0, nothing started)
    $ podman ps -a
    shellhub-probe   Exited (143)

Changing only the policy on the same container:

    $ podman update --restart=always shellhub-probe
    $ podman ps -a --filter should-start-on-boot=true
    shellhub-probe   Exited (143)
    $ podman start --all --filter should-start-on-boot=true
    shellhub-probe   Up 3 seconds

`podman-restart.service` ships with Podman and is disabled by default. The script
neither enables it nor checks for it.

## Docker

Docker is not affected today, but only because of the exit code the agent happens to
die with. Measured in a nested daemon, five containers, one daemon restart:

    NAMES            STATUS
    onfail-run       Up 8 seconds      # --restart=on-failure, killed by signal
    onfail-exit0     Exited (0)        # --restart=on-failure, exited 0
    always-run       Up 8 seconds
    unlessstop-run   Up 8 seconds
    norestart-run    Exited (137)

`on-failure` is honoured on daemon start only when the last exit was non-zero. The
agent qualifies because it has no SIGTERM handler in its main loop: the only
`signal.Notify` is `agent/main.go:455`, scoped to `waitForPairing`. It dies 143.

Adding graceful shutdown to the agent would make it exit 0 and silently stop coming
back after a reboot.

## Proposed change

1. `--restart=unless-stopped` on both container branches. Fixes Podman, and removes
   the exit-code dependency on Docker.

2. In `podman_install`, before `podman run`, abort when there is no boot path: no
   `systemctl`, no `podman-restart.service`, or the unit disabled. Print the command
   to enable it and point at `INSTALL_METHOD=docker` or `standalone`. `wsl_install`
   already aborts this way for its own preconditions.

3. Do not enable the unit silently. It starts every container on the host with that
   policy, not only ours.

4. `SKIP_BOOT_RESTART_CHECK=1` to install anyway, for hosts where the container
   lifecycle is managed elsewhere: a Quadlet, an orchestrator, a CI runner that
   never reboots. It skips the check in item 2 only, and still prints what was
   skipped.

Docker needs none of the checks. The daemon does the work.

luizhf42:
## What

Switches the agent container restart policy from `on-failure` to `unless-stopped` in both the Podman and Docker install paths, and adds a pre-flight check for `podman-restart.service` in the Podman path.

## Why

`--restart=on-failure` never starts the agent container on boot under Podman. Podman has no daemon — boot-time container startup is handled by `podman-restart.service`, which only picks up containers whose restart policy is `always` or `unless-stopped`. With `on-failure`, the agent goes offline on the first reboot and stays offline until manually started.

On Docker the current policy works only by accident: `on-failure` restarts after a daemon restart only when the last exit was non-zero, and the agent happens to die with 143 (unhandled SIGTERM). Adding graceful shutdown to the agent in the future would make it exit 0 and silently break this.

Closes #6958

## Changes

- **`install.sh` restart policy**: `--restart=on-failure` → `--restart=unless-stopped` on both the Podman (`podman run`) and Docker (`docker run`) branches. Fixes Podman and removes the exit-code dependency on Docker.
- **`install.sh` boot-restart pre-flight**: new `check_podman_boot_restart()` called at the top of `podman_install()`. Aborts when `systemctl` is missing or `podman-restart.service` is not enabled, with actionable error messages. The unit is never enabled silently — it affects all containers on the host, not just ours.
- **`SKIP_BOOT_RESTART_CHECK=1`**: escape hatch for managed environments (Quadlet, orchestrators, CI) where boot-time container lifecycle is handled externally. Skips the check and prints what was skipped.
- **`tests/install-boot-check_test.sh`**: shell test covering all five branches of the check function (no systemctl, service disabled, service enabled, skip+no systemctl, skip+disabled).

## Testing

Run the shell test: `bash tests/install-boot-check_test.sh` (5 pass). For end-to-end verification on a Podman host (Fedora/RHEL), install the agent, confirm `podman inspect` shows `unless-stopped`, and verify the container survives a reboot via `podman-restart.service`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -113,7 +113,36 @@ enroll_agent_interactively() {
   $_AGENT_CMD login || true
 }
 
+check_podman_boot_restart() {
+  [ "${SKIP_BOOT_RESTART_CHECK}" = "1" ] && {
+    printf "⚠️  SKIP_BOOT_RESTART_CHECK is set; skipping podman-restart.service check.\n"
+    printf "   The agent container will NOT auto-start on boot unless you manage it yourself\n"
+    printf "   (Quadlet, an orchestrator, etc.).\n"
+    return 0
+  }
+
+  if ! command -v systemctl >/dev/null 2>&1; then
+    printf "❌ ERROR: systemctl not found. Podman needs podman-restart.service to start\n"
+    printf "   the agent container on boot, which requires systemd.\n"
+    printf "   Set SKIP_BOOT_RESTART_CHECK=1 to install anyway (if you manage the container\n"
+    printf "   lifecycle yourself), or use INSTALL_METHOD=docker or standalone.\n"
+    exit 1
+  fi
+
+  if ! systemctl is-enabled podman-restart.service >/dev/null 2>&1; then
+    printf "❌ ERROR: podman-restart.service is not enabled. Without it the agent container\n"
+    printf "   will not start on boot.\n\n"
+    printf "   Enable it with:\n\n"
+    printf "     sudo systemctl enable podman-restart.service\n\n"
+    printf "   Then re-run this script. Set SKIP_BOOT_RESTART_CHECK=1 to install anyway\n"
+    printf "   (if you manage the container lifecycle yourself).\n"
+    exit 1
+  fi
+}
+
 podman_install() {
+  check_podman_boot_restart
+
   [ -n "${KEEPALIVE_INTERVAL}" ] && ARGS="$ARGS -e SHELLHUB_KEEPALIVE_INTERVAL=$KEEPALIVE_INTERVAL"
   [ -n "${PREFERRED_HOSTNAME}" ] && ARGS="$ARGS -e SHELLHUB_PREFERRED_HOSTNAME=$PREFERRED_HOSTNAME"
   [ -n "${PREFERRED_IDENTITY}" ] && ARGS="$ARGS -e SHELLHUB_PREFERRED_IDENTITY=$PREFERRED_IDENTITY"
@@ -171,7 +200,7 @@ podman_install() {
   $SUDO podman run -d \
     --name=$CONTAINER_NAME \
     --replace \
-    --restart=on-failure \
+    --restart=unless-stopped \
     --privileged \
     --pid=host \
     --security-opt label=disable \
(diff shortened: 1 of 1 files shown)


tests/install-boot-check_test.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
otavio: I see the dilemma of adding this, but if we go this way, then we should also, in sequence, extend to cover the rest of those cases that are not yet supported.

https://github.com/bats-core/bats-core

This would be a bad idea with these tests, and I do think that we can do that, but in my opinion that would be a separated pull request to extend this covering all the use cases that we already have.

otavio (changes requested):

tests/install-boot-check_test.sh
luizhf42: Agreed. Removed the hand-rolled test and created a follow-up issue, #6962, to track that work.

otavio (approved):

PR #6984, opened 2026-08-28
chore(lint): enable testifylint and thelper, and fix what they found

otavio:
## What

Enables `testifylint` and `thelper`, and fixes the 216 findings they report across 63 files. All of it is test code.

## Why

Fourth step of the lint expansion, after #6980, #6981 and #6982. These two are what make a failing test say the right thing: the assertion that failed, at the caller's line, instead of a nil panic three lines later.

## Changes

**One real bug.** Two `httptest` handlers called `require.NoError` — in `pkg/wsconnadapter` and `server/ssh/pkg/dialer`. `require` reaches `t.FailNow`, which is only defined on the test's own goroutine. A failed WebSocket upgrade in those handlers would not have stopped the test the way the code reads. They now assert and return.

**125 `require-error`.** The recurring shape is an error asserted with `assert`, then the value it produced dereferenced on the next line. When the assertion fails the test keeps running and panics on nil, so the report is a crash rather than the assertion that actually failed. `require` stops there instead.

`tests/environment` had this hand-rolled seven times as `if !assert.NoError(t, err) { assert.FailNow(t, err.Error()) }`, which is `require.NoError` written out longhand. Those collapse to the real thing.

**91 `thelper`.** Helpers that take a `*testing.T` now declare it, so a failure points at the caller rather than at a line inside the helper. The `storetest` sub-suite closures took `*testing.T` second, which `thelper` reads as a helper with its parameters the wrong way round — they take it first now and `runSubSuite`'s signature follows.

## How

The rewrites were driven from the linter's own `file:line` output rather than typed, then the full suite was run to confirm no assertion changed a result. Two mechanical slips the process caught and I fixed: a `t.Helper()` insertion that landed in a composite literal because the target was a single-line empty closure, and 58 duplicated `testify` imports from the import pass.

## Testing

Under `golang:1.26.7-alpine3.24`:

- `golangci-lint run ./...` reports `0 issues` for all six modules, and for `agent` under no tags, `-tags docker` and `-tags native`
- `go build ./...` clean everywhere
- `go test ./...` passes for root, `server`, and `agent` under both `docker` and `native`
- `go vet ./...` clean; `go mod tidy` a no-op

https://claude.ai/code/session_01D4BSojj4fmh3ZQSZWkGbD8

### DIFF
--- a/pkg/dockerutils/utils_test.go
+++ b/pkg/dockerutils/utils_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
 	"github.com/stretchr/testify/require"
 )
 
--- a/pkg/dockerutils/utils_test.go
+++ b/pkg/dockerutils/utils_test.go
@@ -5,6 +5,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestReadContainerID(t *testing.T) {
@@ -38,7 +40,7 @@ func TestReadContainerID(t *testing.T) {
 921 959 0:131 / /sys/devices/virtual/powercap ro,relatime - tmpfs tmpfs ro,inode64`: "7accf056f96dcfdeb486c705843147f979d3e0ce88f7375e145688e0d2890e33",
 	} {
 		id, err := parseContainerIDv2(strings.NewReader(in))
-		assert.NoError(t, err)
+		require.NoError(t, err)
 		assert.Equal(t, out, id)
 	}
 }
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -13,6 +13,14 @@ formatters:
   enable:
     - gofumpt
     - goimports
+    # goimports keeps whatever grouping it is handed, so an import added in its own
+    # block stays there and reads as a deliberate section. gci decides the grouping.
+    - gci
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
 
 linters:
   exclusions:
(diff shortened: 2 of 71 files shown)


pkg/dockerutils/utils_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/stretchr/testify/require"
```
otavio:
```suggestion
	"github.com/stretchr/testify/require"
```

And make sure that this error is fixed in all cases.

otavio: Addressed in 091b2bda2. The import script appended each `require` as its own blank-line-separated group rather than joining the existing testify group, and goimports preserves whatever grouping it is given, so nothing flagged it.

Fixed in all cases: 9 files here, plus 1 in shellhub-io/cloud#2523. I also checked every file this PR touches for any other import block split into more than two groups — the only remaining ones are five cloud test files that are grouped that way on master already, so I left them alone. Not pushed yet.

otavio: Follow-up: my previous reply said this was fixed, but I had only committed locally and never pushed — so both PRs still showed the defect on GitHub, which is what you were looking at. Both branches are pushed now.

Fixed at the source rather than per-instance: `gci` is now enabled as a formatter in 8abfaf345 (and the cloud equivalent). `goimports` preserves whatever grouping it is handed, which is precisely why a clean lint, a clean vet and a full test run all passed over this — `gci` decides the grouping instead. There are now zero split import blocks in either repo, including the ones that predated this work.

One cost worth knowing: it also flattens grouping that was deliberate. In this repo that is `gateway/modules.go`, where the caddy-dns providers had their own commented block. The comment survives and now names the imports it describes instead of relying on the blank line above them.

PR #7048, opened 2026-09-05
fix(ui): give the pending devices count a page to land on

### LINKED ISSUE #7043: Dashboard's Pending Devices card links to a page that redirects away
## What happens

The Dashboard has a **Pending Devices** stat card whose "View pending" action navigates to
`/devices?status=pending` (`ui/apps/console/src/pages/Dashboard.tsx:34-36`, card at `:86-93`).

That URL does not survive. The Devices page hides every status tab but `accepted`
(`ui/apps/console/src/pages/devices/index.tsx:134`) and an effect forces the filter back:

```tsx
useEffect(() => {
  if (params.status !== "accepted") {
    setFilter("status", "accepted");
  }
}, [params.status, setFilter]);
```

So clicking the card lands on the accepted list. The count it shows is real, the destination is
not — there is no pending-devices list to arrive at.

## Why it is like this

Acceptance moved to the install key: a pending registration is reviewed in the key's activity
(`/install-keys/:id/activity`, `InstallKeyEventReview.tsx`) or on the device's own page
(`DeviceDetails.tsx:232`). The server is unchanged — it still parks registrations as pending
(`api/services/device.go:280-296`). Only the list view went away, and the Dashboard card was
left pointing at it.

## What to do

The card is the only remaining entry point for "N devices are waiting on you", so it is worth
keeping — it just needs a destination that exists. Options:

- Point it at the install key whose activity has the pending registrations
- Point it at a filtered view of the registrations across keys, if one is wanted
- Drop the card, if pending devices are no longer a dashboard-level concern

Found while capturing documentation screenshots: the docs page for accepting devices described a
Pending tab that can no longer be reached.

otavio:
## What

Adds a review queue at `/pending-devices` listing every device waiting on a decision across all
install keys, and points the dashboard card at it. The card's count and the list are now the same
number from the same table.

## Why

The Dashboard's Pending Devices card linked to `/devices?status=pending`. The devices list hides
every status but accepted and an effect forced the parameter back, so the click landed on the
accepted list with no error and no explanation — a real count with no destination.

Acceptance had moved to the install keys area, which left no list of what was waiting: a pending
registration was reachable only by opening each key's activity in turn and scanning it.

Closes #7043

## Changes

- **new page**: lists by device status rather than by key, which is what lets it agree with the
  dashboard count. Each row names the key that enrolled it, so the context acceptance needs
  survives the change of axis. Route is `/pending-devices`, not `/devices/pending`, because
  `NavLink` matches by prefix — nested, the sidebar would light up two entries at once, and an
  exact match on Devices would stop it highlighting on the device detail and add-device pages.
- **links**: dashboard card and the AddDevice callout. The latter also called the destination a
  tab, which has not existed since acceptance moved.
- **devices list**: the forced status left the pending and rejected column sets, their tabs and
  the accept/reject/remove dialog unreachable, and the pending set duplicated the new queue line
  for line. Removed; the tab now links out instead of switching status.
- **pending count**: excluded containers, so the count matches the list it links to — the list
  injects `platform ne connector` unconditionally. Only the pending count moved; the registered
  count answers the same question for licence and firewall evaluation.
- **wizard**: the first-run screen keys off those counts, so a namespace holding only a pending
  container would have been offered a first device it already had.
- **docs**: reviewing a pending enrollment described the key's activity page as where it happens
  "rather than in the device list"; the two are now alternatives.

## Testing

The count-versus-list agreement rests on matching the list's operator exactly. The stats predicate
uses `<>` rather than `IS DISTINCT FROM` deliberately: `platform` is nullable and the list's `ne`
filter emits `<>`, so a null-safe comparison here would count a row the list does not show and
restore the disagreement. `TestPgStore/StatsStore` covers a pending container; the NULL case is not
expressible through the fixture and rests on that reasoning.

Worth probing: a namespace whose only entity is a pending container should still get the dashboard,
not the wizard.


otavio:
## QA validation — run against a live instance

Exercised this branch (`251bba1dc`) end-to-end against a running dev stack (`v0.27.0-rc.8`, UI bind-mounted
from the branch so the browser drove branch source), in **both community and enterprise editions**. Every
feature was reached through its real entry point — sidebar, dashboard card, devices list, add-device page —
rather than by calling endpoints directly.

**28 checks executed: 28 pass, 0 fail.**

### The server fix does what it claims

Seeded a namespace with 4 pending devices, one of them `platform='connector'`:

| | |
|---|---|
| `GET /api/stats` → `pending_devices` | **3** |
| `GET /api/devices?status=pending` → `X-Total-Count` | **3** |
| `GET /api/devices?status=pending&connector=true` | 1 |
| pre-fix query (`status='pending'` alone) on the same rows | **4** |

So the count and the list it is read alongside disagreed before and agree now. Still consistent at 13/13
after the fixture grew.

### Review queue

- Reached by clicking the sidebar entry; lists `status=pending`, sorted `created_at desc`.
- Pagination: 13 pending → 10 rows on page 1, 3 on page 2, URL `?page=2`, request
  `page=2&per_page=10&status=pending&sort_by=created_at&order_by=desc`.
- Install Key column verified on **all four** branches: key name linking to
  `/install-keys/:id/activity`, `Pairing code`, `Tenant-only registration`, and `—` for a null digest.
- Accept moved `ci-runner-01` out of the queue (3→2) and into the accepted devices list; Reject moved
  `legacy-box` out (2→1). Confirmed at the database level, and `pending_devices` dropped accordingly while
  the container stayed excluded.
- Row click → `/devices/:uid`. Empty state renders "No devices are waiting for review".

The review queue listing pending devices with the install key that enrolled each one

Page 2 of the same queue, and the empty state in a namespace with nothing pending:

Page two of the pending queue showing the remaining three devices
The pending queue empty state reading No devices are waiting for review

### The welcome-screen guard, both directions

| namespace | contents | result |
|---|---|---|
| `containerns` | stats all zero, one pending container | dashboard renders (**not** the wizard), after exactly one `GET /api/containers?page=1&per_page=1` |
| `emptyns` | genuinely empty | wizard renders |
| `dev` | stats non-empty | **zero** container requests — the common path keeps its single stats call |

Dashboard rendering for a namespace whose only entity is a pending container, with all device counts at zero
The welcome screen rendering for a genuinely empty namespace

### Also verified

`/devices?status=pending` no longer filters — the Accepted tab stays selected and the request goes out as
`status=accepted`; and the accepted list keeps its columns after a device is accepted from the queue:

The devices list at the status equals pending URL, still showing the Accepted tab
The devices list showing the newly accepted device with online dot, SSHID and connect column

The add-device page's link into the queue, on a non-codeless install method:

The add device page showing the sentence linking to the pending list


Devices list no longer status-filters (`?status=pending` is inert, requests `status=accepted`) and its
accepted columns are intact; the Pending / Install Keys links; the add-device "pending list" link (via a
non-codeless method); `DeviceDetails` enrollment source through the new hook; install-keys pages after the
`KeyValueChip` move; docs build green with the shot registered at `/pending-devices` (and the committed PNG
is a genuine capture of the new page, not stale); OpenAPI spec still validates and the client regenerates.

### `/admin/api/stats` — verified on an enterprise stack

Brought the stack up in enterprise mode (cloud `master` compiles clean against this branch with
`-tags enterprise`) and hit the admin endpoint, which aggregates instance-wide across every namespace:

| | |
|---|---|
| pending rows instance-wide (2 namespaces, 2 of them containers) | 15 |
| pre-fix query (`status='pending'` alone) | **15** |
| post-fix query (excluding `connector`) | **13** |
| `GET /admin/api/stats` → `pending_devices` | **13** |

So the admin surface carries the same correction, not just the namespace-scoped one. No license was needed:
`internal/admin/routes/routes.go` lists `StatsURL` under `licmiddleware.SkipLicense`, and the sibling
`/admin/api/auth/token` returns 402 on the same unlicensed instance, which confirms the exemption is what
let stats through.

<sub>Install-key lifecycle, device removal, live SSH, and the repo's own Vitest/Go suites were out of scope
for this run.</sub>

gustavosbarreto:
#7070 merged and closed #7043, taking a different route: instead of a page listing pending devices across keys, the install key list now says how many devices each key has waiting, in the usage column, and the dashboard card lands there.

Two things from this PR that were right and that #7070 did not carry:

**The dead code.** `4b0b34300` is the only change that fixed the issue from the inside, and it is still needed: `visibleTabs` keeps only the accepted tab and an effect forces the parameter back, so the pending and rejected column sets, their tabs and the accept/reject/remove dialog remain unreachable on master. Reopened as #7071, keeping your commit and authorship, with the pending tab pointing at the install keys list rather than the page this PR added.

**The `used_times` observation.** Chasing the count in the usage column turned up that the counter was only incremented where the server decided admission by itself, so `usage_limit` was never enforced for a manual-mode key. That became #7068 and #7069, which merged just before #7070.

PR #7060, opened 2026-09-08
fix(agent): stop an offline agent logging a line per retry attempt

### LINKED ISSUE #7058: Shellhub Agent ballooning log files
### Describe the issue

 When a device running shellhub agent is left running without an interent connection every few seconds it writes the following

`{"log":"2026/06/30 00:17:05.601858 ERROR RESTY Get \"https://SHELLHUBDOMAIN/info?agent_version=v0.16.0\": dial tcp: lookup SHELLHUBDOMAIN on 127.0.0.1:53: server misbehaving, Attempt 679\n","stream":"stderr","time":"2026-06-30T00:17:05.602080571Z"}`

To log file `/var/lib/docker/containers/0ca59847fa474c9850cbc1e9072b37f71a85fa57cb2b9a1abeb942340b1dc457/0ca59847fa474c9850cbc1e9072b37f71a85fa57cb2b9a1abeb942340b1dc457-json.log`

After a few months this causes the log file to reach several GB and fill up avilable storage. 

I've tried to stop this by adding 

```
{
  "log-driver": "json-file",
  "log-opts": {
    "max-size": "10m",
    "max-file": "3"
  }
}
```

to `nano /etc/docker/daemon.json`. But this has not had an effect. 

What's the best way to reduce to stop the amount of logging done by the shellhub agent? Or make sure the logs are rotated? Is switching to use dockers `local` log driver instead of `json-file` a good option?

otavio:
## What

An agent that cannot reach the server retries indefinitely, and logged a line for every attempt. A
whole outage now costs two lines — one when the server stops answering, one when it answers again —
and the per-attempt trail moves to debug level, where it is still there for anyone who turns it on.

## Why

Fixes #7058, where a device left without internet grew its Docker JSON log to several gigabytes and
filled the disk. The reported line had reached `Attempt 679`:

```
ERROR RESTY Get "https://.../info?agent_version=v0.16.0": dial tcp: lookup ...: server misbehaving, Attempt 679
```

That line is resty's own, from `request.go`, at warn level once per retry. It reached the device's
stderr because the client never installed a logger over it: `SetLogger` was called *before* the
options ran, guarded by a nil check on the very field `WithLogger` sets, so the guard was always
false and the call never happened. resty kept its own logger, writing outside logrus where
`SHELLHUB_LOG_LEVEL` could not reach it.

The reverse-tunnel loop had the same shape from the other direction: one error line every ten
seconds, about 8,600 a day, for as long as a device stayed offline.

## Commits

Seven commits, each self-contained and green on its own.

| Commit | What it does |
|---|---|
| `feat(agent): report a server's reachability as transitions, not as attempts` | New `pkg/connectivity`. Turns a count of attempts into the two moments worth a line. |
| `fix(agent): send the client's logging, and resty's, through the agent's logger` | The `SetLogger` ordering bug, and `LeveledLogger` rewritten printf-style. **This is the commit that closes #7058.** |
| `fix(agent): retry device authorization only on what an operator can clear` | `AuthDevice` stops retrying every refusal forever. Adds the backoff seam. |
| `fix(agent): report an outage once instead of once per retry attempt` | Reporting moves into `AddRetryHook`; dispatch, recovery and the nil-response guard. |
| `fix(agent): log a tunnel outage as a transition rather than once per dial` | The reverse-tunnel reconnect loop. |
| `fix(agent): propagate the device name after a ping re-authorizes, not before` | Independent bug: an inverted condition in the ping loop. |
| `test(agent): take the real backoff out of every test that retries` | `pkg/api/client` goes from 392s to 2.2s. |

## What it logs now

`pkg/connectivity` emits three lines. The first attempt of a run is warn, every one after it debug,
so an outage costs one line however long it lasts:

- `Cannot reach the server, retrying until it answers` — fields `attempt`, `error`
- `Cannot authorize the device, retrying until the server accepts it` — fields `attempt`, `error`
- `Recovered after retrying` — info, fields `attempts`, `after`

Callers that already count attempts pass the count in — resty's `Request.Attempt` is one. Callers
whose retry loop has no counter (the tunnel loop, the ping loop) use a `Tracker`, which keeps it.
Tunnel lines carry `transport=tunnel`, ping lines `transport=ping`, and client lines
`server_address`.

The two conditions are kept apart because they send an operator to different places: the network,
or the namespace and the device limit. A request that got no answer, and a `429` or `5xx` the
client-wide condition retried, are both a server that could not answer. Only a status the
request-level condition retried is a device refusal.

## Retry policy

`AuthDevice` retries just the three statuses an operator clears without touching the device: `404`
for a namespace that does not exist yet, `402` and `403` for one at its device limit. All three are
settled from the console while the agent waits, which is what makes waiting useful.

Everything else fails fast. The list is an allowlist, so a status nobody reasoned about — `400`,
`401`, `409`, `422` — goes back to the caller rather than being retried for the life of the
process. Transport failures, `429` and `5xx` fall through to the client-wide condition. `501` is
excluded from the server's fault: a server that does not implement a route will not grow the
ability by being asked again.

## Testing

`pkg/connectivity` (7 tests) covers the level policy, the fields, and `Tracker` across a run of
failures, a recovery, and the start of a fresh run. `pkg/api/client/connectivity_test.go` (6 tests)
covers the seam rather than the leaf helpers: which status is reported as which condition, `403`
retried, a request that never left not retried, a refusal not reporting a recovery, recovery only
after an outage, and the response body bounded.

Each of those six was verified red against the code before its fix — the defect reverted
individually, the test observed failing, then restored.

`go build`, `go vet` and `golangci-lint` are clean on both modules, and every commit builds and
tests green in isolation, not just the tip.

## Review notes

- The log lines above are the emitted format, read from `pkg/connectivity/connectivity.go`. They
  are not a captured session — the previous version of this description quoted a live run against
  field names (`unreachable_for`) and messages (`Reached the server again`) that no longer exist.
- SSH through the reverse tunnel is not verified end to end here. The edits to `listenV1`/`listenV2`
  swap one log call for another at the same position and do not touch the `a.listening <- true`
  ordering, but a reviewer with a working SSH path should confirm a session still opens after a
  reconnect.
- `Warnf` mapping to debug is deliberate and indiscriminate: resty's other per-request warnings, a
  body it could not unmarshal and Basic Auth over plain HTTP, are buried along with the retry line.
  `Errorf` stays at error, so the request resty finally gave up on is still visible.

### DIFF
--- a/agent/pkg/agentd/agent.go
+++ b/agent/pkg/agentd/agent.go
@@ -69,6 +69,7 @@ import (
 	"github.com/shellhub-io/shellhub/agent/server"
 	"github.com/shellhub-io/shellhub/pkg/api/client"
 	"github.com/shellhub-io/shellhub/pkg/clock"
+	"github.com/shellhub-io/shellhub/pkg/connectivity"
 	"github.com/shellhub-io/shellhub/pkg/envs"
 	"github.com/shellhub-io/shellhub/pkg/models"
 	"github.com/shellhub-io/shellhub/pkg/validator"
@@ -576,7 +577,8 @@ const (
 )
 
 // Listen serves connections until ctx is cancelled, using the transport named by the
-// configuration.
+// configuration. It requires a prior Authorize, whose token it reconnects with and whose logger it
+// reports through.
 func (a *Agent) Listen(ctx context.Context) error {
 	a.mode.Serve(a)
 
@@ -599,11 +601,15 @@ func (a *Agent) listenV1(ctx context.Context) error {
 
 	go a.ping(ctx, AgentPingDefaultInterval) //nolint:errcheck
 
+	logger := a.logger.WithField("transport", "tunnel")
+
 	ctx, cancel := context.WithCancel(ctx)
 	go func() {
+		tunnelServer := connectivity.NewTracker(logger)
+
 		for {
 			if a.isClosed() {
-				a.logger.Info("Stopped listening for connections")
+				logger.Info("Stopped listening for connections")
 
 				cancel()
 
@@ -612,7 +618,7 @@ func (a *Agent) listenV1(ctx context.Context) error {
 
 			ShellHubConnectV1Path := "/ssh/connection"
 
-			a.logger.Debug("Using tunnel version 1")
+			logger.Debug("Using tunnel version 1")
 
 			listener, err := a.cli.NewReverseListenerV1(
 				ctx,
@@ -620,20 +626,22 @@ func (a *Agent) listenV1(ctx context.Context) error {
 				ShellHubConnectV1Path,
 			)
 			if err != nil {
-				a.logger.Error("Failed to connect to server through reverse tunnel. Retry in 10 seconds")
+				tunnelServer.Lost(err)
 
-				time.Sleep(time.Second * 10)
+				time.Sleep(tunnelReconnectInterval)
 
 				continue
 			}
 			a.listener.Store(&listener)
 
-			a.logger.Info("Server connection established")
+			if !tunnelServer.Recovered() {
+				logger.Info("Server connection established")
+			}
 
 			a.listening <- true
 
 			if err := tun.Listen(ctx, listener); err != nil {
-				a.logger.WithError(err).Error("Tunnel listener exited with error")
+				logger.WithError(err).Error("Tunnel listener exited with error")
 			}
 
 			a.listening <- false
@@ -654,11 +662,15 @@ func (a *Agent) listenV2(ctx context.Context) error {
 
 	go a.ping(ctx, AgentPingDefaultInterval) //nolint:errcheck
 
+	logger := a.logger.WithField("transport", "tunnel")
+
 	ctx, cancel := context.WithCancel(ctx)
 	go func() {
+		tunnelServer := connectivity.NewTracker(logger)
+
 		for {
 			if a.isClosed() {
-				a.logger.Info("Stopped listening for connections")
+				logger.Info("Stopped listening for connections")
 
 				cancel()
 
@@ -667,7 +679,7 @@ func (a *Agent) listenV2(ctx context.Context) error {
 
 			ShellHubConnectV2Path := "/agent/connection"
 
-			a.logger.Debug("Using tunnel version 2")
+			logger.Debug("Using tunnel version 2")
 
 			listener, err := a.cli.NewReverseListenerV2(
 				ctx,
@@ -676,20 +688,22 @@ func (a *Agent) listenV2(ctx context.Context) error {
 				client.NewReverseV2ConfigFromMap(a.authData.Config),
 			)
 			if err != nil {
-				a.logger.Error("Failed to connect to server through reverse tunnel. Retry in 10 seconds")
+				tunnelServer.Lost(err)
 
-				time.Sleep(time.Second * 10)
+				time.Sleep(tunnelReconnectInterval)
 
 				continue
 			}
 			a.listener.Store(&listener)
 
-			a.logger.Info("Server connection established")
+			if !tunnelServer.Recovered() {
+				logger.Info("Server connection established")
+			}
 
 			a.listening <- true
 
 			if err := tun.Listen(ctx, listener); err != nil {
-				a.logger.WithError(err).Error("Tunnel listener exited with error")
+				logger.WithError(err).Error("Tunnel listener exited with error")
 			}
 
 			a.listening <- false
@@ -703,6 +717,8 @@ func (a *Agent) listenV2(ctx context.Context) error {
 
 // AgentPingDefaultInterval is the default time interval between ping on agent.
 const AgentPingDefaultInterval = 10 * time.Minute
+
+const tunnelReconnectInterval = 10 * time.Second
 
 func (a *Agent) ping(ctx context.Context, interval time.Duration) error {
 	a.listening = make(chan bool)
--- a/agent/pkg/agentd/agent.go
+++ b/agent/pkg/agentd/agent.go
@@ -730,6 +730,8 @@ func (a *Agent) ping(ctx context.Context, interval time.Duration) error {
 	<-a.listening // NOTE: wait for the first connection to start to ping the server.
 	ticker := time.NewTicker(interval)
 
+	authorization := connectivity.NewTracker(a.logger.WithField("transport", "ping"))
+
 	for {
 		if a.isClosed() {
 			return nil
@@ -766,6 +768,10 @@ func (a *Agent) ping(ctx context.Context, interval time.Duration) error {
 			}
 		case <-ticker.C:
 			if err := a.authorize(); err != nil {
+				authorization.Refused(err)
+			} else {
+				authorization.Recovered()
+
 				a.server.SetDeviceName(a.authData.Name)
 			}
 
--- a/pkg/api/client/client.go
+++ b/pkg/api/client/client.go
@@ -43,12 +43,13 @@ type Client interface {
 }
 
 type client struct {
-	scheme   string
-	host     string
-	port     int
-	http     *resty.Client
-	logger   *log.Logger
-	reverser reverser.Reverser
+	scheme    string
+	host      string
+	port      int
+	http      *resty.Client
+	logger    *log.Logger
+	retryWait func() time.Duration
+	reverser  reverser.Reverser
 }
 
 // ErrParseAddress is returned by NewClient when the server address is not a URL carrying scheme,
@@ -77,74 +78,44 @@ func NewClient(address string, opts ...Opt) (Client, error) {
 		const MinRetryAfterSecs int = 5
 		const MaxRetryAfterSecs int = 65
 
-		t := time.Duration(rand.IntN(MaxRetryAfterSecs-MinRetryAfterSecs)+MinRetryAfterSecs) * time.Second //nolint:gosec
-
-		log.WithFields(log.Fields{
-			"retry_after": t,
-		}).Warn("retrying request after a random time period")
-
-		return t
+		return time.Duration(rand.IntN(MaxRetryAfterSecs-MinRetryAfterSecs)+MinRetryAfterSecs) * time.Second //nolint:gosec
 	}
 
 	client := new(client)
+	client.retryWait = randomWaitTimeSecs
 	client.http = resty.New()
 	client.http.SetRetryCount(math.MaxInt32)
 	client.http.SetRedirectPolicy(SameDomainRedirectPolicy())
 	client.http.SetBaseURL(uri.String())
 	client.http.SetContentLength(true)
 	client.http.AddRetryCondition(func(r *resty.Response, err error) bool {
+		if r == nil {
+			return false
+		}
+
 		var netErr net.Error
 		if errors.As(err, &netErr) {
-			log.WithFields(log.Fields{
-				"url": r.Request.URL,
-			}).WithError(err).Error("network error")
-
 			return true
 		}
 
-		switch {
-		case r.StatusCode() == http.StatusTooManyRequests:
-			log.WithFields(log.Fields{
-				"status_code": r.StatusCode(),
-				"url":         r.Request.URL,
-				"data":        r.String(),
-			}).Warn("too many requests")
-
-			return true
-		case r.StatusCode() >= http.StatusInternalServerError && r.StatusCode() != http.StatusNotImplemented:
-			log.WithFields(log.Fields{
-				"status_code": r.StatusCode(),
-				"url":         r.Request.URL,
-				"data":        r.String(),
-			}).Warn("failed to achieve the server")
-
-			return true
-		}
-
-		return false
+		return serverAtFault(r.StatusCode())
 	})
-	client.http.SetRetryAfter(func(c *resty.Client, r *resty.Response) (time.Duration, error) {
+	client.http.SetRetryAfter(func(_ *resty.Client, r *resty.Response) (time.Duration, error) {
 		switch r.StatusCode() {
 		case http.StatusTooManyRequests, http.StatusServiceUnavailable:
 			retryAfterHeader := r.Header().Get(RetryAfterHeader)
 			if retryAfterHeader == "" {
-				return randomWaitTimeSecs(), nil
+				return client.retryWait(), nil
 			}
 
 			retryAfterSeconds, err := strconv.Atoi(retryAfterHeader)
 			if err != nil {
-				return randomWaitTimeSecs(), err
+				return client.retryWait(), err
 			}
-
-			log.WithFields(log.Fields{
-				"status":      r.StatusCode(),
-				"retry_after": retryAfterSeconds,
-				"url":         r.Request.URL,
-			}).Warn("retrying request after a defined time period")
 
 			return time.Duration(retryAfterSeconds) * time.Second, nil
 		default:
-			return randomWaitTimeSecs(), nil
+			return client.retryWait(), nil
 		}
 	})
 	client.http.SetRetryMaxWaitTime(MaxRetryWaitTime)
@@ -164,4 +135,21 @@ func NewClient(address string, opts ...Opt) (Client, error) {
 	client.http.SetLogger(&LeveledLogger{client.logger})
 
 	return client, nil
+}
+
+func serverAtFault(status int) bool {
+	if status == http.StatusTooManyRequests {
+		return true
+	}
+
+	return status >= http.StatusInternalServerError && status != http.StatusNotImplemented
+}
+
+func operatorCanResolve(status int) bool {
+	switch status {
+	case http.StatusNotFound, http.StatusPaymentRequired, http.StatusForbidden:
+		return true
+	default:
+		return false
+	}
 }
--- a/pkg/api/client/client.go
+++ b/pkg/api/client/client.go
@@ -3,16 +3,19 @@ package client
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math"
 	"math/rand/v2"
 	"net"
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	resty "github.com/go-resty/resty/v2"
 	"github.com/shellhub-io/shellhub/pkg/api/client/reverser"
+	"github.com/shellhub-io/shellhub/pkg/connectivity"
 	"github.com/shellhub-io/shellhub/pkg/models"
 	log "github.com/sirupsen/logrus"
 )
@@ -48,6 +51,7 @@ type client struct {
 	port      int
 	http      *resty.Client
 	logger    *log.Logger
+	logEntry  *log.Entry
 	retryWait func() time.Duration
 	reverser  reverser.Reverser
 }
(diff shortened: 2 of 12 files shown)


otavio:
## QA validation log (re-run against `c50b14991`)

Replaces the earlier log, which was run against `2c8e3927b` and quoted messages the current code no longer emits.

Run against a live agent in the containerised dev stack (agent `latest`, API `v0.27.0-rc.8`). Outages simulated by stopping `shellhub-gateway-1`; the agent runs `network_mode: host`, so a stopped gateway is a genuine connection refusal.

### Clean boot is quieter than before the rework

The duplicate `Connected to the server` is gone, and master's line is back:

```
INFO Listening for connections        mode=multi-user tenant_id=00000000-0000-4000-0000-000000000000
INFO Server connection established    transport=tunnel sshid=dev.c8-7f-54-63-39-f3@localhost namespace=dev
```

### Outage on an established agent — tunnel path

2m10s down, 13 reconnect attempts. One warning, one info:

```
WARN Cannot reach the server, retrying until it answers  attempt=1 error="read tcp ...: connection reset by peer" transport=tunnel
INFO Recovered after retrying  after=2m10.102310287s attempts=13 transport=tunnel
```

Reproduced on a second agent in the same window: identical shape, `attempts=13`, `after=2m10.074836885s`.

### Outage on boot — HTTP path through `AddRetryHook`

Agent started while the server was unreachable, `SHELLHUB_LOG_LEVEL=info`, 1m46s down, 4 attempts. Six lines total, one of which is the outage:

```
INFO    Starting ShellHub  mode=multi-user version=latest
WARN    Cannot reach the server, retrying until it answers  attempt=1 error="Get \"http://localhost:80/info?agent_version=latest\": dial tcp [::1]:80: connect: connection refused"
INFO    Recovered after retrying  after=1m46.046819112s attempts=4
INFO    Listening for connections
INFO    Server connection established  transport=tunnel sshid=dev.qa3@localhost
```

The `attempt` field is resty's `Request.Attempt`, which is the point of the rework — nothing counts it a second time.

### The debug trail

Same outage at `SHELLHUB_LOG_LEVEL=debug`. `attempt=1` warns, everything after it debugs, and resty's own per-attempt line comes through logrus rather than its stderr logger:

```
level=debug msg="Get \"...\": dial tcp [::1]:80: connect: connection refused, Attempt 1"
level=warning msg="Cannot reach the server, retrying until it answers" attempt=1
level=debug msg="Get \"...\": dial tcp [::1]:80: connect: connection refused, Attempt 2"
level=debug msg="Cannot reach the server, retrying until it answers" attempt=2
level=debug msg="Cannot reach the server, retrying until it answers" attempt=3
```

`RESTY`-prefixed lines in the whole run: 0. Attempt spacing (21:35:28, 21:36:23, 21:36:35) confirms the 5–65s backoff is intact.

### Test suite

`go test` and `golangci-lint` clean on both modules; `go.mod`/`go.sum` unchanged. The retry-policy cases that previously took 167s now run in 0.203s, since the backoff has a seam.

### Still not verified

SSH end to end. The dev SSH server issues an interactive device-login approval code, and authenticated API calls in this environment return `400` (unrelated to this change — it touches no server API code), so the device cannot be approved programmatically. Partial evidence the listener path is intact: the tunnel re-established after every outage in this run, and the server logged `v2 connection established` / `v2 connection bound` each time. A reviewer with a working SSH path should confirm a session still opens after a reconnect.

gustavosbarreto (commented):
Good diagnosis. Three real defects found here, and all three are worth landing:

1. `SetLogger` was called before the `opts` loop, so `client.logger` was always nil and the branch never fired. resty was logging to its own stderr logger, outside logrus, where `SHELLHUB_LOG_LEVEL` could not reach it.
2. `LeveledLogger` treated resty's variadic args as logrus key/value pairs, but resty's `Logger` is printf-style. `Errorf("%v", err)` would have panicked on the out-of-range read for the missing pair. It was dead code, so nobody hit it, but wiring it up as written would have.
3. `Client.execute` returns `(nil, err)` when its pre-request middleware fails, and the retry condition dereferenced `r` unconditionally.

The part I'd push back on is `pkg/connectivity`. The state the `Tracker` keeps behind a mutex already exists in resty, scoped to the request, because of how this client is configured.

With `SetRetryCount(math.MaxInt32)`, an outage is not a sequence of requests. It is one `Execute` that keeps retrying, and `resp.Request.Attempt` (`request.go:1057`) is the attempt counter for it. That makes the transition `Attempt == 1`, with no state machine and no lock:

```go
client.http.AddRetryHook(func(r *resty.Response, err error) {
	if r == nil {
		return
	}

	entry := logger.WithError(err).WithField("attempt", r.Request.Attempt)

	if r.Request.Attempt > 1 {
		entry.Debug(retryReason(r))

		return
	}

	entry.Warn(retryReason(r))
})

client.http.OnSuccess(func(_ *resty.Client, r *resty.Response) {
	if r.Request.Attempt > 1 {
		logger.WithField("attempts", r.Request.Attempt).Info("The server is answering again")
	}
})
```

`AddRetryHook` also matters for correctness, not just for shape. Reporting from inside a retry condition makes the bookkeeping depend on evaluation order. resty runs `append(r.retryConditions, r.client.RetryConditions...)` and breaks on the first `true` (`retry.go`, `Backoff`), so the request-level condition in `AuthDevice` returning `true` means the client-wide condition never runs. It works here, but only because of the order the conditions happen to be registered in, and nothing in the code says so.

One hook can also carry both messages, picking by what the response says: `r.StatusCode() == 0` is a server that did not answer, `r.IsError()` is a server that answered and refused. That removes the `Messages` struct, both exported mutable vars, and the second tracker.

For the outage duration, without reintroducing shared state:

```go
client.http.OnBeforeRequest(func(_ *resty.Client, r *resty.Request) error {
	if r.Attempt <= 1 {
		r.SetContext(context.WithValue(r.Context(), startedAt{}, time.Now()))
	}

	return nil
})
```

`executeBefore` runs on every attempt and `r.Attempt` is already 1 on the first, so the stamp is the start of the outage. `context.WithValue` keeps the parent's `Done()`, so `Backoff`'s cancellation is unaffected.

The tunnel loop is the one place this does not cover, since it is not resty. But `listenV1`/`listenV2` are a plain `for` in a single goroutine, so a local `attempt` counter reset on a successful listen does the same job without a package or a mutex.

Points to fix regardless of which shape you keep:

- `identity := req.Identity.MAC` in `AuthDevice` now runs on every call. It used to be a closure invoked only inside `if r.IsError()`. A nil `Identity` moves from an error-path panic to a happy-path one.
- The `r == nil` guard sits after the `errors.As(err, &netErr)` branch. The error it guards against comes from `executeBefore` wrapped in `wrapNoRetryErr`. If such an error is ever a `net.Error`, the condition returns `true` and the client retries forever a request resty marked as not retryable. The guard should be first.
- `Tracker` hardcodes the `failures` and `unreachable_for` field names while only the text comes from `Messages`. With `DeviceAuthorization` the log reads `Authorized the device unreachable_for=3m` for a server that answered every single time.
- A clean boot now logs `Connected to the server` twice, once from the client tracker on the first `GET /info` and once from the tunnel tracker, distinguished only by the `transport` field, plus `Authorized the device`. The description only demonstrates the outage path.
- `time.Now()` with a `//nolint:forbidigo` where the repo has `pkg/clock`. The test pays for it: it can only assert `Contains(recovery.Data, "unreachable_for")`, never a value.
- Not related to this PR, but this change makes it reachable: `ping` does `if err := a.authorize(); err != nil { a.server.SetDeviceName(a.authData.Name) }`. That condition looks inverted, and `AuthDevice` failing fast on 400/422 will now exercise it.

otavio:
@gustavosbarreto Thanks — the `Attempt` observation is the right one, and I want to take most of it. Verified the mechanics: `Attempt` is exported (`request.go:52`) and incremented at the top of the `Backoff` operation (`request.go:1057`), so `Attempt == 1` really is the first try, and with `SetRetryCount(math.MaxInt32)` an outage is one `Execute` rather than a sequence of them. All three defects you list stand, and one of them — the `r == nil` guard sitting after the `errors.As` branch — turned out to be worse than a style point, so it is going in either way. More on that below.

Where I want to push back slightly is on deleting the abstraction entirely, because the two designs disagree about less than it looks.

## They agree on the rule, not on where the count lives

Both want the same thing: first failure warns, repeats go to debug, recovery infos with a count. Mine stores the count in a `Tracker`; yours reads resty's. That is the whole disagreement.

What each buys:

**Yours** — no duplicated state, the right seam (a hook that fires on the retry decision, so conditions stay pure predicates and the ordering dependence disappears), per-request by construction so two trackers cannot both report one outage, and a much smaller exported surface.

**Mine** — one vocabulary across both transports, and a transition rule testable without HTTP. Those are not free to give up. The tunnel is `gorilla/websocket`, so your design needs a second mechanism for it, as you note. And once the rule lives inside resty, the only way to exercise it is through a real 5–65s backoff: the retry-policy tests I added already sleep 54s, up to ~130s, for exactly that reason.

## Taking both

Make the reporter stateless and let each call site supply the count:

```go
// no state, no mutex, no lifecycle
func Failed(logger logrus.FieldLogger, attempt int, err error)  // attempt == 1 -> Warn, > 1 -> Debug
func Recovered(logger logrus.FieldLogger, attempt int)
```

The resty path passes `r.Request.Attempt` from `AddRetryHook` — your point, no stored state, no mutex. The tunnel passes its own local counter reset on a successful listen — one vocabulary, both transports. Conditions go back to being pure predicates deciding policy only (`serverAtFault`, `operatorCanResolve`); the hook does all the reporting and picks its message the way you suggested, `StatusCode() == 0` for a server that did not answer versus `IsError()` for one that answered and refused.

That drops the `Tracker`, `Messages`, both exported vars and the second tracker, which was your objection, while keeping the tunnel covered and the tests fast.

Two details from your sketch I would keep and change:

- The `OnBeforeRequest` context stamp for the duration is worth keeping. It is the one clever bit, but "unreachable for 3h" is the number the reporter in #7058 actually cares about.
- `OnSuccess` fires on 4xx — `onErrorHooks` runs the success hooks whenever `err == nil`. Confirmed by running it: a 400 auth failure currently emits `info "Connected to the server"`. The recovery line needs guarding on `!r.IsError()` as well as `Attempt > 1`.

## Your other points

Already fixed in `d84db724a`, which landed after you reviewed `310f732ec`: the `identity := req.Identity.MAC` hoist.

Taking all of these:

- **Guard order.** More than a style issue. `Backoff` passes `unwrapNoRetryErr(err)`, so the marker is stripped before the condition sees it, and `parseRequestURL`/`createHTTPRequest` return a raw `*url.Error`, which satisfies `net.Error`. So `errors.As` matches first and we retry forever a request resty marked non-retryable. Since my version no longer dereferences `r` there, master's loud panic became a silent hang. Guard goes first.
- **Hardcoded `failures`/`unreachable_for` for both trackers.** Correct, and it reads exactly as badly as you say. The stateless reporter lets the call site choose its own fields.
- **Duplicate `Connected to the server` at boot.** Also correct, and the PR only ever demonstrated the outage path.
- **`time.Now()` with a nolint where `pkg/clock` exists.** Taking it.
- **`ping`'s inverted `authorize()`.** Confirmed, and my fail-fast on 400/401/409/422 is what makes it live — worse, those statuses now return without a log line, so a de-authorized device loops silently. Separate commit.

Reworking as: guard order and the `Warnf` doc first, then the stateless reporter plus `AddRetryHook`, then the tunnel counter, then `ping`, then tests with a seam for the backoff so they stop sleeping. Shout if you would rather I go all the way to no package at all — happy to, I just did not want to lose the tunnel or the fast tests without flagging it.

Also: the QA log above is stale. It was run against `2c8e3927b`, before the authorization tracker existed, and it quotes a debug message the current code no longer emits. I will replace it once the rework is in.

otavio:
@gustavosbarreto The rework is pushed (branch rewritten into seven commits — `3866eced7`). Almost all of your review is in. One place I did not follow you, and one place your sketch needed a correction; both below, plus the test-to-production ratio since it is lopsided enough to want explaining.

## Taken as written

- **The count comes from resty.** `r.Request.Attempt` on the HTTP side, no stored state, no mutex, nothing counting a second time. The `Tracker` that kept a count behind a lock is gone from that path entirely.
- **`AddRetryHook` instead of reporting from a condition.** Your correctness argument, not just shape: `Backoff` walks `append(r.retryConditions, r.client.RetryConditions...)` and breaks on the first `true`, so reporting inside one made the bookkeeping depend on registration order. Conditions are now pure predicates (`serverAtFault`, `operatorCanResolve`) and the hook does the reporting.
- **One hook carrying both messages**, picked from what the response says.
- **The `OnBeforeRequest` context stamp** for the outage duration. Kept exactly as you wrote it, including that `context.WithValue` preserves the parent's `Done()`.
- **`Messages`, both exported mutable vars, and the second tracker** — deleted, and they did not come back.
- **Guard order.** Moved first. As noted, this was worse than style: `Backoff` hands the condition `unwrapNoRetryErr(err)`, so a request that never left arrives as a raw `*url.Error`, which satisfies `net.Error`, and `errors.As` matched before the nil check. Master panicked; the intermediate version hung forever.
- **`OnSuccess` on 4xx** — now guarded on `IsError()` as well as the attempt count.
- **`ping`'s inverted `authorize()`** — its own commit, `fix(agent): propagate the device name after a ping re-authorizes, not before`.
- **The hardcoded `failures`/`unreachable_for` field names** — gone. Fields are `attempt`, `attempts`, `after`.
- **Duplicate `Connected to the server` at boot** — gone. A clean boot logs nothing from the client path and master's `Server connection established` from the tunnel.

## Where I did not follow you

You suggested the tunnel keep "a local `attempt` counter reset on a successful listen ... without a package or a mutex", and in my earlier reply I said I would do exactly that. I did, and then backed it out.

The reason is arity. There is not one non-resty loop, there are three: `listenV1`, `listenV2`, and — once `AuthDevice` stops retrying every refusal — the `ping` loop, which now reaches a failed re-authorization every tick. Each needs the same five lines:

```go
if attempt == 0 {
    lostAt = time.Now()
}
attempt++
// ... on success:
if attempt > 0 { Recovered(logger, attempt, time.Since(lostAt)) } else { logger.Info(...) }
attempt = 0
```

`listenV1` and `listenV2` carried that byte-for-byte identically, and `ping` would have made three copies of the transition rule. So `pkg/connectivity` has a `Tracker` again — but it is not the one you objected to:

- **No mutex.** It documents that it belongs to the single goroutine running the loop it counts, which is true of all three call sites.
- **Not on the resty path.** `client.go` calls the package-level `Lost`/`Refused`/`Recovered` with `r.Request.Attempt`, exactly as you sketched. Two trackers cannot both report one outage because the HTTP side has none.
- **It is a thin holder**, not a state machine: the transition rule still lives in one `level(attempt)` function that both the free functions and the `Tracker` go through.

If you would still rather see three hand-rolled counters and no `Tracker`, say so and I will do it — it is a small change in one direction or the other. I went this way because "the rule lives in three places" was the thing I could not talk myself out of.

## One correction to the sketch

`r.StatusCode() == 0` for a server that did not answer versus `r.IsError()` for one that answered and refused is not sufficient, and I shipped it wrong the first time before catching it.

A `429` or a `502` has a non-zero status and `IsError() == true`, so it lands in the refusal branch and logs `Cannot authorize the device, retrying until the server accepts it` for a server that is simply down. That sends an operator to the namespace and the device limit for a fault that is not there.

The hook now asks the same question the conditions ask: no status, **or** `serverAtFault(status)` (429, 5xx except 501), is a server that could not answer. Only a status the request-level condition retried is a device refusal.

## On `time.Now()` and `pkg/clock`

I said I would take this one and then did not, so it deserves an explanation rather than silence. The repo's own `forbidigo` message is:

> use `clock.Now()` so a test can control it. If this is an elapsed-time measurement or a deadline, keep `time.Now` and add `//nolint:forbidigo` saying which.

Both remaining uses are elapsed-time measurements, so the suppression is the sanctioned path, not an evasion of it. Your actual objection was that the test pays for it — that it can only assert the field exists, never a value. That part is fixed by signature rather than by `clock`: `Recovered(logger, attempt, elapsed)` takes the duration as a parameter, so `TestRecovered` asserts `after` is exactly `3m50s`.

## Test lines versus production lines

871 lines changed, and only about a third of it is production code:

| | files | insertions | deletions |
|---|---|---|---|
| Production | 6 | 271 | 110 |
| Tests | 6 | 599 | 5 |

Three things drive that.

**The production change is substantially a deletion.** 110 lines removed against 271 added, and the biggest single edit — `pkg/api/client/client.go` — is mostly *removing* per-attempt logging from the retry conditions and the backoff, then adding back a hook, a helper and two predicates. `pkg/connectivity` is 103 lines. The behaviour change is genuinely small; that is the point of it.

**There were no tests at these seams at all.** The retry policy, the level policy, the dispatch, and the logger adapter had zero coverage before this branch — the adapter was dead code that would have panicked the first time it was called. `pkg/connectivity` is 103 production lines against 206 test lines, and `pkg/api/client/connectivity_test.go` is 245 new lines covering six behaviours that previously had none.

**Table-driven Go tests are verbose out of proportion to what they assert.** `level(attempt)` is five lines of production code; pinning it across the boundary and both sides of it is a forty-line table. The 94 lines added to `client_public_test.go` are almost entirely the six-case status table for which refusals an operator can clear.

Worth saying explicitly: each of the six tests in `connectivity_test.go` was verified red against the code before its fix — defect reverted individually, test observed failing, then restored. They are not written-after-the-fact assertions that the current behaviour is the current behaviour.

The seventh commit also takes the real 5–65s backoff out of every test that retries, which is why the package now runs in 2.2s instead of 392s.

gustavosbarreto (approved):
Approving. This fixes #7058 and the three defects behind it were all real: `SetLogger` running before the options were applied, so `client.logger` was always nil and resty logged to its own stderr logger outside logrus where `SHELLHUB_LOG_LEVEL` could not reach it; `LeveledLogger` reading resty's printf arguments as key/value pairs, which would have panicked on `Errorf("%v", err)` the moment the first fix made it reachable; and the retry condition dereferencing a `*Response` that resty returns nil when its pre-request middleware fails.

The rework also reads well. The retry condition is a predicate again, the reporting happens in one place, the message is chosen from what the response says, and the start time rides on the request context because resty resets `Request.Time` per attempt.

Three follow-ups, none of them blocking this.

**`pkg/connectivity` is more package than the problem needs.** 46 lines of executable code behind 8 exported symbols, plus 206 lines of test, to pick a log level from an `attempt` the caller already holds. There is no state being kept and no invariant being protected outside the `Tracker`, whose state is an `int` and a `time.Time`. Written where it is used, the same behaviour is around 60 lines: a level and a message in the retry hook, a counter in each tunnel loop, and nothing at all in `ping`, where `AuthDevice` already retries internally and the loop ticks rather than retries. `pkg/api/client/connectivity_test.go` is worth keeping either way, since it tests client behaviour rather than the package.

**`retryWait` is a test seam on the production struct.** The tests are in the same package and already reach `client.http` directly for httpmock, so overriding resty's own retry knob there does the same job without adding a field to what ships on the device.

**The ping tracker's wording followed it somewhere it is not true.** It says "Cannot authorize the device, retrying until the server accepts it", but that loop does not retry until accepted, it waits for the next `AgentPingDefaultInterval` tick. Accurate where `AuthDevice` uses it.

And one thing worth its own issue rather than a comment. The agent has three loops that independently discover that the server is gone: the HTTP client retrying inside resty, the tunnel reconnect loop, and `ping`. None of them knows the others already found out, which is why the `Tracker` doc has to say that two of them covering one server each report the same outage. The log volume is the symptom; the absence of a single place that knows whether the agent is connected is the cause, and `a.listening` as a channel and `CheckUpdate` running on a 24h timer with no idea whether there is a network are the same gap showing up elsewhere. If that ever becomes work, it should start from `a.listening` rather than from logging, because anything that only picks a log level is this package again under a different name.

PR #7069, opened 2026-09-10
fix(server): charge an install key use when a device is accepted

### LINKED ISSUE #7068: Manual-mode install keys never consume their usage limit
## What happens

A key in `manual` mode admits devices without ever spending its allowance. `used_times` stays at 0 no matter how many devices are accepted, so `isOverused()` (`pkg/models/install_key.go:191`) never becomes true, `IsValid()` keeps returning true, and `usage_limit` has no effect in that mode.

The namespace's `legacy` system key is created in manual mode (`server/api/store/pg/namespace.go:61`), so keyless tenant-only registration is on the affected path by default.

## Reproduction

1. Create an install key with `mode: manual` and `usage_limit: 1`.
2. Register two devices with it. Both land `pending`, as expected for manual mode.
3. Accept both from the key's activity page.
4. Both devices are `accepted`. The key still reads `0 / 1` and continues to accept registrations.

```sql
select k.name, k.mode, k.usage_limit, k.used_times,
       count(d.id) filter (where d.status = 'accepted') as accepted_devices
from install_keys k
left join devices d on d.install_key_id = k.key_digest
group by 1, 2, 3, 4;

     name      |  mode  | usage_limit | used_times | accepted_devices
---------------+--------+-------------+------------+------------------
 manual-review | manual |           1 |          0 |                2
```

## Why

`InstallKeyIncrementUsage` has two callers:

- `server/api/services/enrollment.go:127`, inside `applyEnrollmentDecision` on the `enrollAccept` branch
- `server/api/services/install-key.go:512`, the deferred enrollment webhook callback

`applyEnrollmentDecision` only runs where the server decides admission itself during a registration request: first registration (`server/api/services/auth.go:337`), re-registration of a removed device (`auth.go:380`), and `reconcileEnrollment` (`enrollment.go:199`), which is restricted to webhook and allowlist keys by `ReconcilableOnAuth` (`pkg/models/install_key.go:164`).

Accepting from the console is `PATCH /api/devices/:uid/accept`. There is no dedicated route for it: it matches `UpdateDeviceStatusURL = "/devices/:uid/:status"` (`server/api/routes/device.go:22`, registered at `server/api/routes/routes.go:160`) with `:status = accept`, and lands in `service.UpdateDeviceStatus`. That method writes the device status and calls `InstallKeyEventStampDecision` to freeze the verdict on the history event (`server/api/services/device.go:292`). It does not touch the key's counter.

The counter therefore tracks admissions the server granted automatically, not admissions granted. Coverage by mode:

| Mode | How the accept happens | `used_times` |
|---|---|---|
| `automatic` | at registration | incremented |
| `allowlist`, MAC matches | at registration | incremented |
| `allowlist`, left pending | reconciled on a later agent phone-home | incremented |
| `webhook`, synchronous accept | at registration | incremented |
| `webhook`, deferred callback | callback | incremented |
| `webhook`, timed out then reconciled | later phone-home | incremented |
| pairing code | at registration | incremented |
| `manual` | operator accepts from the queue | **not incremented** |
| any mode, `rejected` then accepted | operator accepts | **not incremented** |

## What to do

Charge a use wherever a device becomes accepted, rather than only where the server decided the admission by itself.

`InstallKeyIncrementUsage` is already the atomic gate: it updates under `WHERE usage_limit = 0 OR used_times < usage_limit` and returns `ErrNoDocuments` when the key is exhausted (`server/api/store/pg/install-key.go:184`). `InstallKeyDecrementUsage` (`install-key.go:207`) exists for the rollback and is used that way by the webhook callback.

Two decisions this needs:

- **What an exhausted key does to a manual accept.** `applyEnrollmentDecision` logs and leaves the device pending, which is right when no one is waiting on the response. An operator clicking Accept needs an error saying the key is out of uses, not a silent no-op.
- **Existing data.** Installations may hold manual keys whose accepted device count already exceeds `usage_limit`; after the fix those keys stop admitting new devices. Either backfill `used_times` from the accepted device count per key, or leave the counter as it is so only new accepts are charged. The first is truthful and immediately blocks over-limit keys; the second is quieter and lets them drift for one more cycle.

Related: `used_times` is monotonic. `InstallKeyDecrementUsage` is only used to undo a failed accept, so removing an accepted device does not return the use. The counter means "admissions granted", not "devices currently held".

gustavosbarreto:
An install key's `used_times` was incremented only where the server decided the admission by itself during a registration request: the automatic mode, an allowlist match, a webhook answering accept, the deferred webhook callback, and the reconcile that runs on a later agent phone-home. Accepting a device from the review queue goes through `UpdateDeviceStatus`, which stamped the verdict on the history event and left the counter alone.

The consequence was that `usage_limit` had no effect on a key in `manual` mode. A single-use manual key admitted devices forever, because `isOverused()` reads `used_times` and that never grew. The namespace's `legacy` key is created in manual mode, so keyless tenant-only registration was on that path too.

The charge now lives in the device status transaction, which is the one place that owns a device becoming accepted. Every path that accepts a device gets it, whoever made the decision. The store's increment was already the limit check (`WHERE usage_limit = 0 OR used_times < usage_limit`), so an exhausted key surfaces as `ErrInstallKeyExhausted` and the accept fails with 403 rather than silently doing nothing. Because the charge runs inside the transaction, a failed status write rolls the use back on its own, so the enrollment paths and the webhook callback no longer reserve and release it by hand.

Automatic enrollment behaviour is unchanged: an exhausted key still leaves the device pending, logged, since no one is waiting on that response.

### Note for reviewers

Existing installations may hold manual keys whose accepted device count already exceeds `usage_limit`. This change does not backfill `used_times`, so those keys keep admitting devices until the counter catches up with the limit through new accepts. Backfilling from the accepted device count would be truthful and would block them immediately, at the cost of turning an upgrade into a provisioning stop for whoever relied on the limit not working.

Closes #7068

### DIFF
--- a/server/api/services/device.go
+++ b/server/api/services/device.go
@@ -298,6 +298,23 @@ func (s *service) UpdateDeviceStatus(ctx context.Context, req *requests.DeviceUp
 	return nil
 }
 
+func (s *service) chargeInstallKeyUse(ctx context.Context, tenantID, installKeyID string) error {
+	if installKeyID == "" {
+		return nil
+	}
+
+	key := &models.InstallKey{ID: installKeyID, TenantID: tenantID}
+	if err := s.store.InstallKeyIncrementUsage(ctx, key); err != nil {
+		if errors.Is(err, store.ErrNoDocuments) {
+			return ErrInstallKeyExhausted
+		}
+
+		return err
+	}
+
+	return nil
+}
+
 func (s *service) updateDeviceStatus(req *requests.DeviceUpdateStatus) store.TransactionCb {
 	return func(ctx context.Context) error {
 		namespace, err := s.store.NamespaceResolve(ctx, store.NamespaceTenantIDResolver, req.TenantID)
@@ -381,6 +398,10 @@ func (s *service) updateDeviceStatus(req *requests.DeviceUpdateStatus) store.Tra
 				if err := s.validateDeviceAcceptance(ctx, namespace); err != nil {
 					return err
 				}
+			}
+
+			if err := s.chargeInstallKeyUse(ctx, namespace.TenantID, device.InstallKeyID); err != nil {
+				return err
 			}
 		}
 
--- a/server/api/services/device_test.go
+++ b/server/api/services/device_test.go
@@ -2371,6 +2371,83 @@ func TestUpdateDeviceStatus_licenseEvaluator(t *testing.T) {
 	storeMock.AssertExpectations(t)
 }
 
+func TestUpdateDeviceStatus_keylessDeviceSpendsNoKey(t *testing.T) {
+	envstest.SetEdition(t, envs.Community)
+
+	savedHooks := deviceMergeHooks
+	deviceMergeHooks = nil
+	t.Cleanup(func() { deviceMergeHooks = savedHooks })
+
+	now := time.Date(2025, 1, 15, 12, 0, 0, 0, time.UTC)
+	clockMock := clockmock.NewMockClock(t)
+	clockMock.On("Now").Return(now)
+	prevClockBackend := clock.DefaultBackend
+	t.Cleanup(func() { clock.DefaultBackend = prevClockBackend })
+	clock.DefaultBackend = clockMock
+
+	storeMock := storemock.NewMockStore(t)
+	queryOptionsMock := storemock.NewMockQueryOptions(t)
+	storeMock.On("Options").Return(queryOptionsMock).Maybe()
+
+	ctx := context.Background()
+	const tenantID = "00000000-0000-0000-0000-000000000000"
+
+	device := &models.Device{
+		UID:      "keyless",
+		Name:     "keyless",
+		TenantID: tenantID,
+		Status:   models.DeviceStatusPending,
+		Identity: &models.DeviceIdentity{MAC: "aa:bb:cc:dd:ee:ff"},
+	}
+	accepted := *device
+	accepted.Status = models.DeviceStatusAccepted
+	accepted.StatusUpdatedAt = now
+
+	storeMock.
+		On("NamespaceResolve", ctx, store.NamespaceTenantIDResolver, tenantID).
+		Return(&models.Namespace{TenantID: tenantID, MaxDevices: -1}, nil).
+		Once()
+	storeMock.
+		On("DeviceResolve", ctx, mock.Anything, store.DeviceUIDResolver, "keyless").
+		Return(device, nil).
+		Once()
+	queryOptionsMock.On("WithDeviceStatus", models.DeviceStatusAccepted).Return(nil).Once()
+	storeMock.
+		On("DeviceResolve", ctx, mock.Anything, store.DeviceMACResolver, "aa:bb:cc:dd:ee:ff", mock.AnythingOfType("[]store.QueryOption")).
+		Return(nil, store.ErrNoDocuments).
+		Once()
+	storeMock.
+		On("DeviceResolve", ctx, mock.Anything, store.DeviceHostnameResolver, "keyless", mock.AnythingOfType("[]store.QueryOption")).
+		Return(nil, store.ErrNoDocuments).
+		Once()
+	storeMock.On("DeviceUpdate", ctx, &accepted).Return(nil).Once()
+	storeMock.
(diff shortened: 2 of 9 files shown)


otavio:
Two axes reviewed against the repo rules and #7068. Core fix is right; these need addressing:

**Blocking**
- **The operator still doesn't see the reason.** `ErrInstallKeyExhausted` correctly reaches 403 with
  `install key usage limit reached`, but `ui/apps/console/src/utils/acceptErrors.ts` maps every 403 to
  "You do not have permission to accept devices in this namespace." The issue asked for "an error saying
  the key is out of uses" — right now it's a wrong error instead of a silent no-op.
- **Dead reserve/release protocol.** `InstallKeyDecrementUsage` has zero callers after this change, but the
  interface (`server/api/store/install_key.go:62`), the pg impl and the mock remain, and their doc comments
  still describe the reserve/release the diff deleted. Same staleness in `InstallKeyIncrementUsage`'s doc
  and `enrollment_e2e_test.go:402`. Remove it or rewrite the docs.

**Should fix**
- `chargeInstallKeyUse` is unexported, so its comment isn't one of the two exceptions `code-style.md` allows.
  Its one non-obvious fact — the store's increment *is* the limit check — didn't reach the commit message;
  move it there and drop the comment.
- `removed → accepted` charges a second use for a device that was already charged. Only `accepted` is
  guarded, so that edge falls through.
- The no-backfill decision is argued in the PR body only. It belongs in the commit message, where
  `git blame` reaches it.

**Test gaps**
- The issue's own reproduction (manual key, `usage_limit: 1`, accept two) has no e2e test, though
  `enrollment_e2e_test.go` already has the `installKey`/`usedTimes` seam that would prove the counter —
  a better place for this than a mock test asserting the store call.
- Nothing covers `rejected → accepted` charging, or the pairing path *not* double-charging.

gustavosbarreto:
Addressed what this PR made, pushed. Three findings I'm leaving out, with reasons.

**Dead reserve/release protocol** — right, I made it dead. `InstallKeyDecrementUsage` is gone: interface, pg impl and mock. Nothing outside this repo referenced it. One correction: `InstallKeyIncrementUsage`'s doc is not stale. It says it increments atomically only while the key has usage left and returns `ErrNoDocuments` once overused, which is still exact and now carries more weight, since that predicate is the only limit check left. Reworded the two places that described where the reservation happened (`enrollment_e2e_test.go:402` and its subtest name).

**`chargeInstallKeyUse` comment** — dropped. The non-obvious fact moved to the commit message.

**No-backfill decision** — moved into the commit message, where `git blame` reaches it.

**Test gap on the issue's reproduction** — agreed, and the e2e seam was the right call: `TestEnrollmentE2E_AcceptSpendsAUse` builds a manual key with `usage_limit: 1`, enrolls two devices, asserts the counter is still 0 while both wait, then accepts one (counter 1) and asserts the second accept fails with `ErrInstallKeyExhausted`, stays pending, and spends nothing. It also covers `rejected → accepted` spending a use, through an allowlist key that rejected the MAC. That replaces the mock cases that asserted the same thing indirectly; what stays as a mock test is the one negative with no observable side effect, a device with no `install_key_id` reaching accepted without any call to the store's increment.

**403 copy in the console — out of scope here.** The finding is real: `acceptErrors.ts` maps every 403 to the permission sentence, so the server's message never reaches the operator. But that mapping already mistranslates any other 403 from that endpoint, it predates this change, and it is a UI file in a server fix. It also cannot be fixed without deciding something this PR should not smuggle in: the error body carries only `message`, no machine-readable code, so the options are a distinct status for the exhausted case, a code in the error body, or keying UI copy off server prose. That is API shape and wants its own issue. If you want it, open one.

**`removed → accepted` charging twice — not introduced here.** On master, re-registration of a removed device already went through `applyEnrollmentDecision` (`auth.go:380`) and incremented again. And under the semantics #7068 states, `used_times` means admissions granted and removal does not return the use, so charging a re-admission is correct rather than an unguarded edge. Wanting "devices the key currently holds" is a different model and a different change.

**Pairing not double-charging** — covered behaviour this PR does not alter. `TestEnrollmentE2E_UsageLimitUnderConcurrency` already pins the counter against a limit on the automatic path.

PR #7070, opened 2026-09-10
feat: say which install key has devices waiting for a decision

### LINKED ISSUE #7043: Dashboard's Pending Devices card links to a page that redirects away

gustavosbarreto:
The Dashboard's Pending Devices card pointed at `/devices?status=pending`, a URL the devices list undoes: it hides every status tab but accepted and an effect forces the parameter back. The count was real, the destination was not.

Acceptance lives in the install keys area, so that list is where the card should land. It just had nothing to say. With two devices sitting pending, both keys read `0 / ∞`, because `used_times` only counts devices already accepted. The one page where a pending registration is reviewed gave no sign that anything was waiting, and finding one meant opening each key's activity in turn.

So the list now says it. The install key payload carries `pending_devices`, and the usage cell shows it: a device awaiting a decision is a claim on the same allowance an accepted one spends, so that column is the only place able to say that accepting the queue would exhaust the key. While devices wait they take the cell's headline and the spend moves below the bar. On a key with a real limit the bar draws the projection, and hatches when accepting everything waiting would pass the limit.

The count is over devices with status pending rather than history events without a verdict, because an event outlives the device it registered and a removed pending device would keep inflating it. Counting devices also makes the sum across a namespace's keys equal the `pending_devices` figure the stats endpoint reports, so the card and the page agree.

The Dashboard card and the callout under the install command both point at the install keys list now, and the docs stop describing a Pending tab that has not existed since acceptance moved.

Supersedes #7048, which answered the same issue with a dedicated `/pending-devices` page and a sidebar entry.

Closes #7043
Closes #7048

### DIFF
--- a/openapi/spec/components/schemas/installKey.yaml
+++ b/openapi/spec/components/schemas/installKey.yaml
@@ -63,6 +63,11 @@ properties:
     type: integer
     description: How many devices have registered with the key.
     example: 3
+  pending_devices:
+    type: integer
+    description: How many devices the key registered that are still awaiting a decision.
+    minimum: 0
+    example: 1
   last_used_at:
     type: string
     format: date-time
--- a/pkg/models/install_key.go
+++ b/pkg/models/install_key.go
@@ -106,6 +106,8 @@ type InstallKey struct {
 	UsageLimit int `json:"usage_limit"`
 	// UsedTimes is how many devices have enrolled with the key.
 	UsedTimes int `json:"used_times"`
+	// PendingDevices counts the key's devices still awaiting a decision.
+	PendingDevices int `json:"pending_devices"`
 	// LastUsedAt is when a device last enrolled with the key.
 	LastUsedAt *time.Time `json:"last_used_at"`
 	// Ephemeral marks devices enrolled with the key for automatic removal once offline past
--- a/server/api/store/pg/entity/install-key.go
+++ b/server/api/store/pg/entity/install-key.go
@@ -37,6 +37,8 @@ type InstallKey struct {
 	CreatedAt          time.Time  `bun:"created_at"`
 	UpdatedAt          time.Time  `bun:"updated_at"`
 	ExpiresAt          *time.Time `bun:"expires_at,nullzero"`
+	// PendingDevices is counted by InstallKeyList; it is not a stored column.
+	PendingDevices int `bun:"pending_devices,scanonly"`
 }
 
 // InstallKeyFromModel projects an install key into its row form.
@@ -100,6 +102,7 @@ func InstallKeyToModel(entity *InstallKey) *models.InstallKey {
 		Reusable:           entity.Reusable,
 		UsageLimit:         entity.UsageLimit,
 		UsedTimes:          entity.UsedTimes,
+		PendingDevices:     entity.PendingDevices,
 		LastUsedAt:         entity.LastUsedAt,
 		Ephemeral:          entity.Ephemeral,
 		EphemeralTimeout:   entity.EphemeralTimeout,
--- a/server/api/store/pg/install-key.go
+++ b/server/api/store/pg/install-key.go
@@ -75,6 +75,13 @@ func (pg *Pg) InstallKeyConflicts(ctx context.Context, sc scope.Scope, target *m
 	return conflicts, len(conflicts) > 0, nil
 }
 
+const pendingDevicesExpr = `(
+	SELECT COUNT(*) FROM devices d
+	WHERE d.install_key_id = install_key.key_digest
+	  AND d.namespace_id = install_key.namespace_id
+	  AND d.status = 'pending'
+) AS pending_devices`
+
 // InstallKeyList implements [store.InstallKeyStore].
 func (pg *Pg) InstallKeyList(ctx context.Context, sc scope.Scope, opts ...store.QueryOption) ([]models.InstallKey, int, error) {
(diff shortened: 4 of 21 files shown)


otavio:
A few things to fix before merge:

**Correctness**
- `used_times` is incremented only at auto-accept (`services/enrollment.go:127`) and the
  webhook callback (`services/install-key.go:512`). The console accept path
  (`UpdateDeviceStatus`) never increments it and never checks `usage_limit`, so the
  projection bar and the `· N over` warning describe a limit that path does not enforce.
- `enrollmentInstallKey` (`services/auth.go:183-191`) can return `""` and `install_key_id`
  is `nullzero`, so such a pending device is counted by the stats endpoint and by no key —
  the sum-equals-stats claim has a hole.
- `useAcceptDevicePairing` (`hooks/useDeviceCode.ts:27`) accepts a pending device but does
  not invalidate `installKeyList`; the count goes stale on the pairing path.
- `pending_devices` has no `omitempty`, so the schema's "omitted from a single-key read" and
  the model's "List only" are not true as written.

**Standards**
- The comment on `pendingDevicesExpr` (`store/pg/install-key.go:78-85`) is on an unexported
  declaration and its text is already verbatim in commit `00db2554`; same for the `base.css`
  comment. `code-style.md` allows only doc comments on exported symbols and suppression
  reasons.
- `install_key_tests.go:97-101` uses raw `"pending"`/`"accepted"`/`"rejected"`; every other
  caller of `WithDeviceStatus` uses `models.DeviceStatus*`.
- `device-not-appearing.mdx` — the body now points at Install Keys, but the heading is still
  `## Is it in Pending?`.

**Worth a look**
- `devices(install_key_id, namespace_id, status)` is unindexed, so the correlated count runs
  per key per page.
- The pending arithmetic sits inline in `UsageMeter` while its siblings live behind
  `helpers.ts`, which is why the new test has to assert through a rendered row.

gustavosbarreto:
Pushed. What changed, and what did not.

**`enrollmentInstallKey` can return `""`.** Real, and it breaks the sum-equals-stats claim as written. It does not happen today: every namespace gets a legacy key at creation and migration 013 backfilled the existing ones. But `install_key_id` is nullable and the claim was stated as an invariant, so the commit message now scopes it to every device that has a key and names the gap.

**`useAcceptDevicePairing` not invalidating `installKeyList`.** Added. The pairing accept admits a device without going through the device mutations, so the count went stale on that path.

**`pending_devices` wording.** There is no single-key read that serializes `models.InstallKey`, so the sentence described a contract that does not exist. Dropped from the schema description and the model doc.

**Missing index.** Migration 026 adds a partial index on `devices (namespace_id, install_key_id) WHERE status = 'pending'`. Partial because that is the only status the count reads, and pending is a small, short-lived slice of a namespace's devices. It does not reintroduce what migration 020 removed: that was a presence heartbeat writing `last_seen`/`disconnected_at`, where an indexed column changing disqualifies HOT. Neither column is in this index, and `status` changes only on a decision.

**Arithmetic inline in `UsageMeter`.** Moved to `getWaitingInfo` in `helpers.ts`, with an `it.each` over five limit/spent/waiting combinations including the exact-at-limit boundary. It was also why the other test had to assert through a rendered row.

**The comments.** Both removed. Why the count is over devices rather than over history events is in the server commit message; why the hatch takes `currentColor` is in the UI one. What stays in the diff is five doc comments on exported declarations, each with an identical sibling in its own file.

The migration keeps its comment, matching every other migration in that directory, 020 included. Say the word and it moves to the commit message too.

**`used_times` and the projection.** That is #7069. The counter is incremented only at auto-accept and the webhook callback, and the console accept path does not enforce `usage_limit`; #7068 tracks it, #7069 fixes it. The sequencing point is right: the projection and the `· N over` warning are only truthful once #7069 lands, so it should merge first.

**Raw status strings.** Changed to `models.DeviceStatus*`. For the record, `storetest` is split on this: 32 raw against 29 typed.

**`device-not-appearing.mdx` heading.** Now "Is it waiting to be accepted?".

PR #7072, opened 2026-09-10
fix(ui): link a device to the source it registered through

gustavosbarreto:
`REGISTERED VIA` on the device page rendered `<DeprecatedBadge />` alone for an enrollment whose source is the tenant-only path, so the field answered `Deprecated` and named no source at all. That path is what the install script uses when no install key is given, which makes it the most common device in a namespace.

The name now comes from `enrollmentSourceName`, which shares the two system-source labels with `installKeyDisplayName`, so what a device reports and what the keys list shows cannot drift apart. It links to that key's activity, where this device's registration record lives: what the agent reported, the key fingerprint, the decision. All three sources link, including the two system keys, which have activity pages like any other.

The deprecation badge does not come along. Whether the mechanism is on its way out is a fact about the key, and the keys list is where it is actionable; on a device page it answers a question nobody asked there.

```
REGISTERED VIA          REGISTERED VIA
Deprecated              Tenant-only registration   -> /install-keys/<digest>/activity
```

### DIFF
--- a/ui/apps/console/src/pages/DeviceDetails.tsx
+++ b/ui/apps/console/src/pages/DeviceDetails.tsx
@@ -1,5 +1,10 @@
 import { useEffect, useState } from "react";
-import { useParams, useNavigate, useSearchParams } from "react-router-dom";
+import {
+  Link,
+  useNavigate,
+  useParams,
+  useSearchParams,
+} from "react-router-dom";
 import Breadcrumb from "@/components/common/Breadcrumb";
 import { LABEL_BASE } from "@/utils/styles";
 import {
@@ -17,8 +22,10 @@ import {
 } from "../hooks/useDeviceMutations";
 import { useNamespace } from "../hooks/useNamespaces";
 import { useInstallKeys } from "../hooks/useInstallKeys";
-import { resolveEnrollmentSource } from "@/pages/install-keys/helpers";
-import { DeprecatedBadge } from "@/pages/install-keys/StatusChip";
+import {
+  enrollmentSourceName,
+  resolveEnrollmentSource,
+} from "@/pages/install-keys/helpers";
 import { useAuthStore } from "../stores/authStore";
 import { useTerminalStore } from "../stores/terminalStore";
 import ActionDialog from "@/components/common/ActionDialog";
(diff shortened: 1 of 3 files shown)


otavio (dismissed):

otavio:
A few things to fix before merge.

1. `helpers.test.ts:271-278`: the "agrees with what the keys list calls the same
   key" case now compares `SYSTEM_SOURCE_NAMES` to itself. It passes if both
   labels are wrong and only fails if the constant lookup breaks. The `it.each`
   above already pins the expected strings, so drop this case.
2. `DeviceDetails.tsx:314`: `—` to `&mdash;` is unrelated to the fix, and line
   338 in the same file still uses the literal. Revert it or do it everywhere in
   a separate commit.
3. The badge rule is still duplicated. `DeviceDetails` tests
   `enrollment.kind === "legacy"`, `InstallKeysTable.tsx:145` tests
   `!isPairingKey(key)`. Only the label is shared, so the two views can still
   disagree on whether a source is deprecated. Returning the label and the
   deprecated flag from one function would close it.
4. No test covers the row that regressed.
   `pages/devices/__tests__/DeviceDetails.test.tsx` renders this page, but
   nothing asserts that a legacy enrollment shows the name and the badge. The
   badge-only regression would come back silently.
5. The added `<span className="flex items-center gap-2">` wraps children that
   `InfoItem` already puts in `<dd className="flex items-center gap-1">`. Drop
   it, or match `gap-1`.

gustavosbarreto:
Pushed. The change is not what you reviewed: the badge is gone from the device page entirely, and the source is now a link to the key's activity.

**The tautological test.** Right, and it went further than you asked. Both sides of that assertion read `SYSTEM_SOURCE_NAMES`, so it could only fail if the lookup broke, and the `it.each` above it asserted a constant against itself in three colours. The whole block is gone.

**No test covers the row that regressed.** This is the one that mattered. It now lives in `pages/devices/__tests__/DeviceDetails.test.tsx`, which renders the page and already mocks `installKeyList`: a device with a legacy `install_key_id` must show a link named `Tenant-only registration` pointing at `/install-keys/legacy-digest/activity`. That is the regression, asserted through what a person sees, at the network boundary and nowhere deeper.

**`&mdash;`.** Reverted. Unrelated, and half-done at that.

**The duplicated badge rule.** Moot now, and the reason is worth stating: the device page no longer decides whether a source is deprecated, because it no longer says so. `DeviceDetails` tests nothing about deprecation; `InstallKeysTable` remains the only place that does. There is no second rule left to disagree with the first.

**The wrapper span.** Gone with the badge. The field is a single `Link`, so `InfoItem`'s own `<dd className="flex items-center gap-1">` is doing the layout.

PR #7087, opened 2026-09-13
refactor(server): give the reverse tunnel a dial seam

### LINKED ISSUE #7086: refactor(server): give the reverse tunnel a dial seam
## Problem Statement

Everything the SSH gateway does after it reaches a device is untested, and not for want of anyone
trying. The reverse tunnel has no seam. Every layer of the SSH path holds the concrete dialler and
uses exactly one method on it, so there is no interface to substitute and no fake agent anywhere in
the tree. A test that wants to reach past the dial has two options: stand up a real agent over a
real tunnel, or reach around the dialler and write the session's unexported connection field.

The tree does the second, twice. The one test covering `connect` assigns the agent connection
directly. The banner handler carries a struct of function fields whose only purpose is to let a test
substitute the dial. Both are workarounds for the same missing seam, and both test past the
interface rather than through it.

The cost falls on the largest and most-changed code in the server. Session finalisation, the
237-line session-channel request loop, and the web-terminal bridge have no unit coverage at all, and
two of them carry suspected defects that no test can currently reach.

## Solution

Name the single operation the SSH path performs on a reverse tunnel, and let a fake satisfy it.

`TunnelDialer` is a one-method interface in the dialler module, satisfied by the real dialler and by
two test adapters: a fake agent that completes an SSH handshake over an in-memory pipe, and a stub
that fails on demand. The tunnel-extension registration point narrows to the same interface, which
also stops an out-of-tree extension reaching the tunnel registry through an exported field.

What a caller wants from a tunnel and how that is negotiated stay on opposite sides of the seam. The
fake agent does not speak the V1 or V2 bootstrap. Those stay behind the seam and keep the tests they
already have, next to the code that speaks them.

## User Stories

1. As a maintainer, I want the reverse tunnel behind a named interface, so that a test can
   substitute it without a live agent.
2. As a maintainer, I want a fake agent that completes an SSH handshake in process, so that session
   finalisation and the channel request loop become reachable from a unit test.
3. As a maintainer, I want a stub that fails with each of the dial's error modes, so that callers'
   failure paths are assertable rather than assumed.
4. As a maintainer, I want the `connect` test to drive the public path instead of assigning an
   unexported field, so that it breaks when the contract breaks and not when the field moves.
5. As a maintainer, I want the banner handler's ad-hoc dial substitution retired, so that there is
   one way to fake a dial instead of two.
6. As a maintainer, I want to tell the three dial failures apart at the call site, so that a log
   line, and later an operator-facing message, can distinguish a device that holds no tunnel from
   one that holds a tunnel and does not answer.
7. As an author of an out-of-tree tunnel extension, I want the registration point to hand me the
   ability to dial and nothing else, so that I cannot register or evict a tunnel by accident.
8. As an operator, I want every SSH and web-terminal behaviour to be unchanged, so that this work is
   invisible to me.
9. As a cloud maintainer, I want the web-endpoint proxy to compile and behave unchanged, so that the
   narrowing costs a signature and nothing more.

## Implementation Decisions

**The seam.** One interface, declared in the dialler module because all three consumers use the
whole of it and none can avoid importing the module anyway. `Target` is a sealed interface, its only
method unexported, so naming a target already requires the import. A consumer-declared interface
would therefore decouple nothing and produce three copies to keep in sync.

```go
type TunnelDialer interface {
    DialTo(ctx context.Context, tenant, uid string, target Target) (net.Conn, error)
}
```

**What narrows to it.** The session's dialler field, the SSH HTTP handlers' dialler field, the SSH
server and its banner-handler constructors, and the tunnel-extension function type. The composition
root keeps the concrete dialler, because the Prometheus collector reads the tunnel registry through
it.

**What does not move.** The tunnel registry stays concrete and reachable through the dialler's
exported field. It is genuinely deep, roughly sixty lines building a yamux client, evicting the
displaced tunnel and running a keepalive loop, but nothing fakes it today and nothing is asking to.
One adapter is a hypothetical seam. Narrowing the extension point is what stops an extension
reaching it, which is the part that has a reason now.

**Error modes become part of the interface.** The dial fails three distinguishable ways: an empty
tenant or device, which is a programming error; no tunnel held, which is how an offline device is
observed; and a tunnel that does not complete the exchange. Today every caller collapses all three.
The interface states them and callers branch where they can act on the difference. No message an
operator sees changes in this work.

**The domain term.** `CONTEXT.md` gains **reverse tunnel**: the connection a device holds open to
the server, which the server dials back through. The direction is inverted from every other
connection in the system, the device being the listener, and that is the fact a reader is most
likely to have backwards. It is also what a later architecture pass would need in order not to
propose that the server simply connect to the device.

**Commits.** One pull request against `master`, three commits, in order: the seam and the
narrowings, with no behaviour change; the test harness together with its first callers; the error
contract. A fix for anything the new tests catch goes in its own commit behind those. Splitting this
way means the weakest commit can be dropped from the branch without rewriting the others.

**Cloud lands in lockstep.** Cloud's `go.mod` replaces both shellhub modules with the sibling working
tree, so narrowing the extension point breaks its build immediately. Cloud's diff is two lines, a
struct field and a closure parameter, on a branch sharing this one's name, because cloud CI resolves
the shellhub source by branch match. Cloud reaches no part of the dialler other than `DialTo`, so
nothing else is affected, and its existing proxy tests pass an untyped nil dialler, which satisfies
an interface parameter exactly as it satisfied a pointer.

## Testing Decisions

**One new seam, two existing ones reused.** Everything here is driven through `TunnelDialer`. The
channel-handler interface already exists and already has a hand-written fake; it is reused
unchanged. The generated service mock is reused unchanged. Narrowing the SSH path's view of the
service layer is a separate finding and is deliberately not bundled here.

**Two adapters, not one with a mode flag.** A fake agent for the paths that must succeed, and a
stub returning a configured error for the paths that must fail. A single fake with a failure switch
would grow, and the point of the seam is that something real varies across it.

**The harness.** A sibling helper package to the dialler module, matching the house pattern that
`envstest`, `storetest` and `dbtest` established: normal build files rather than `_test.go`, helpers
taking `*testing.T` first, `t.Helper()` and `t.Cleanup` for teardown, no build tags. It is linted
with the full set, so every exported symbol needs a doc comment. Where such a package carries an
invariant, the convention is that it ships its own test enforcing it.

The fake agent is new code. Nothing in the repository stands up an SSH server on the agent side
today, and `NewServerConn` appears nowhere. The nearest prior art is the dialler module's own target
tests, which script a responsive fake agent over an in-memory pipe for the multistream negotiation,
and the web bridge's test helper that writes an ephemeral ed25519 host key to a temporary directory.
Both are worth reading before writing this.

**Acceptance.** Named behaviours, not a count.

1. The dial's target dispatch, driven through the fake. Currently unreachable.
2. Session finalisation, which writes a raw HTTP request onto a connection the SSH client already
   owns, while the close handler does the same job properly through a close target. A test here is
   expected to fail first.
3. The close handler, through the stub, asserting the target it was handed.
4. The `connect` test migrated off the unexported-field assignment onto the seam.
5. The banner handler's function-field dial substitution retired in favour of the interface.

Deliberately follow-ups rather than acceptance, so that the pull request stays reviewable: the
authentication path's `fallthrough`, which can leave a session registered with the API but not
authenticated, and the session-channel request loop, whose agent-forwarding goroutine defers a close
inside a loop and so leaks a channel per forward.

**Prior art to match.** The tree's hand-rolled fake conventions are settled and should be followed
rather than reinvented: `stub` for passive substitutes, `fake` for behavioural stand-ins,
`recording` for spies that carry assertions, spy state guarded by a mutex or atomics because tests
run under `-race`, and table tests whose cases are sentence-shaped descriptions.

## Out of Scope

- **Any operator-visible message.** Telling an offline device from an unreachable one in the SSH
  banner needs a new banner kind, a new embedded message file, and a mapping on the web path. It is
  a product decision about copy and belongs in its own change. This work only makes the distinction
  available.
- **Splitting the dialler module in two.** The registration side deserves a seam the day something
  needs to fake it. Not before.
- **Narrowing the SSH path's view of the service layer.** Seven files take the whole service
  interface, 115 methods, to call 21 of them, and every SSH test loads a 7,699-line generated mock.
  That is a separate finding from the same architecture pass and wants its own change.
- **The two suspected defects,** beyond landing the failing tests that expose them. If either fix
  turns out to be non-trivial, it becomes its own issue with the test skipped and linked.
- **The V1 and V2 bootstrap.** It keeps its existing tests, on the far side of the seam.

## Further Notes

An earlier SSH test harness was deleted in August for having no callers, with the reasoning that an
unused test harness is worse than none because it reads as the established pattern when the real
helpers live elsewhere. That deletion sets the bar this work has to clear, and it is a fair one. The
harness proposed here differs in three ways: it fakes the agent rather than the gateway, it runs
over an in-memory pipe rather than a real listener, and it lands in the same commit as the tests that
use it. If it ever stops having callers, it should be deleted again.

The reverse tunnel's dial path already carries a deadline, added when a handshake was found to block
indefinitely against an agent that never answered. That work is the reason the handshake is
separable from the transport at all, and the new harness should not re-test it.

Found by the `improve-codebase-architecture` scan. The same pass produced two larger findings not
filed here: namespace scope being a compile-time obligation on reads and a convention on writes, and
the service-interface narrowing noted above.

otavio:
## What

Put the reverse tunnel behind `TunnelDialer`, a one-method interface a test can substitute, and
use it to reach three areas of the SSH path that had no unit coverage at all. Fixes one defect
the new tests caught.

## Why

Everything the SSH gateway did after it reached a device was untested. Every layer held the
concrete dialler and used exactly one method on it, so there was no interface to substitute and
no fake agent anywhere in the tree. A test that wanted to reach past the dial had two options:
stand up a real agent over a real tunnel, or reach around the dialler and write the session's
unexported connection field.

The tree did the second, twice. The one test covering `connect` assigned the agent connection
directly. The banner handler carried a struct of function fields whose only purpose was to let
a test substitute the dial. Both were workarounds for the same missing seam.

Closes #7086

## Changes

- **The seam.** `TunnelDialer` is declared in the dialler module rather than by each consumer,
  because `Target` is sealed. Naming a target already requires the import, so a
  consumer-declared copy would decouple nothing and leave three definitions to keep in sync.
- **What narrowed.** The session's dialler field, `NewSession`, the SSH server and both
  banner-handler constructors, `Handlers.Dialer`, and `TunnelExtension`. An out-of-tree
  extension is now handed the ability to dial and nothing else.
- **The registry stayed concrete.** `Handlers` reaches it through its own `Tunnels` field. The
  agent endpoints that register a tunnel cannot get at it through an interface that only dials,
  and the composition root still holds the concrete dialler for the Prometheus collector.
- **The test harness.** A fake agent that completes an SSH handshake in process, and a stub
  that fails on demand. Two adapters rather than one with a mode flag, because the point of the
  seam is that something real varies across it.
- **The error contract.** `ErrUnreachable` joins `ErrInvalidArgument` and `ErrNoConnection`, so
  a caller can tell a device holding no tunnel from one holding a tunnel it does not answer on.
- **The fix.** Finalisation sends the session close on its own connection.

## Testing

The test harness needed a buffered in-memory transport. A plain `net.Pipe` cannot carry an SSH
handshake. Both ends write their version string before reading one, and an unbuffered pipe
deadlocks on that first exchange. `pipe.go` queues writes and drains them from a goroutine.
Reads and deadlines pass through, so a caller can still bound a handshake.

Two things need a closer look.

**Commit 2 is red on its own.** `TestFinishLeavesTheSSHTransportIntact` and its two neighbours
land failing, and commit 4 fixes them. The branch is deliberately not bisectable through that
range. This is what the issue asked for, so the fix can be dropped without rewriting the
commits under it.

**The finalisation fix changes the wire.** The old close was written as a `DELETE`, but the
agent's V1 route is `GET:///ssh/close/:id` (`agent/pkg/agentd/handlers.go:229`), so it has been
arriving as an unroutable request for as long as it has been written that way. It also went
onto the connection the SSH client owns, which breaks the transport under any seat still on it.
A unit test cannot cover either half, so please confirm this against a real agent on both
transport versions.

Cloud lands the matching change on a branch of this name. Its `go.mod` replaces both shellhub
modules with the sibling tree, so merging this alone reddens cloud CI.

Out of scope, and left as follow-ups: the authentication path's `fallthrough`, which can leave
a session registered with the API but not authenticated, and the session-channel request loop,
whose agent-forwarding goroutine defers a close inside a loop.


otavio:
**On the `CONTEXT.md` gap: the entry exists, but not where this checkout can see it.**

`shellhub/CONTEXT.md` is a symlink to `../claude/shellhub/CONTEXT.md`, in the separate
`shellhub-io/claude` repo. A CI checkout of this repo alone cannot resolve it, which is why the
grep for "tunnel" came back empty.

The **Reverse tunnel** section is written and says what the issue asked for, including the
inverted direction:

> **Reverse tunnel** — the connection a device holds open to the server, which the server dials
> back through to reach the device. The direction is inverted from every other connection in the
> system: the device is the listener and the server is the caller.

It landed as `0c02af4 docs(shellhub): name the reverse tunnel in the glossary` on the
`docs/reverse-tunnel-glossary` branch of `shellhub-io/claude`, which is pushed but not yet
merged. So the requirement is met and the definition is not duplicated into Go doc comments, but
it is gated on that branch merging. Flagging it here so it does not get lost.

otavio:
Addressed in `aa103cf`, `58ce8c8` and `2a943e2`.

**Correctness, both taken.**

- *Teardown blocking on a dial.* `closeOnAgent` now runs on its own goroutine with its own
  bounded context, so the client connection's `Close` no longer waits on a device holding a
  wedged tunnel. The close does not need to be synchronous with the client's FIN or with
  `DeactivateSession`. The finalisation tests wait for the dial rather than assuming it has
  happened when `Finish` returns.
- *Dead pump reporting success.* `asyncWriter` records the pump's error and returns it from
  every later `Write`. `pipe_test.go` drives the exact path described, a deadline firing while a
  write is parked, and asserts the next write fails.

**Standards, all taken.**

- `Describe` is now exported from the dialler module and both call sites use it, leaving only
  the log level at each. The close handler keeps `Warning` for a device holding no tunnel.
- The three-way contract moved onto `TunnelDialer`. `(*Dialer).DialTo` documents only the nil
  target case it adds.
- `Handlers`' doc no longer claims the split stops a reaching handler from registering. It names
  `TunnelExtension` as the narrowing that bites, and says the split is for substituting the dial
  in a test.
- `Finish`'s "callers can defer it" is gone.

**Spec, both taken.**

- Acceptance 3: the stub path now asserts the close target it was handed, not only the error.
- `TestFinishRunsOnce` asserts the deactivation count with `AssertNumberOfCalls`, so a double
  deactivation fails the test whose name promises it will not happen.

The `CONTEXT.md` gap is answered separately: the entry exists in the `claude` repo that
`shellhub/CONTEXT.md` symlinks into, on a branch that has not merged yet.

Two notes I agree with and have not acted on. `Agent.broken`/`served` being per-agent rather
than per-dial is a real trap, but no test drives two `serve()`-backed dials against one `Agent`
and inventing the case now would be speculative. The `s.agent.conn != nil` guard does proxy for
"did we reach the device", and it is byte-for-byte present on `master`, so it stays out of a
branch that is already carrying one behaviour fix.

PR #7090, opened 2026-09-14
docs(manage): fold recording into sessions, and play one

gustavosbarreto:
Stacked on #7089 — merge that one first and this rebases onto master on its own.

Three things about session recording, which was a page of its own describing infrastructure a reader on Cloud does not have.

**The storage section is gone.** It named `SHELLHUB_OBJECT_STORAGE_*`, MinIO and what a database dump leaves out — self-hosting, on a page about using the product. Backup already lists the object store as its own thing to back up and the reference has the variables; the one detail missing there, that the store grows with session activity, moved to Backup, where somebody sizing a disk is reading.

**Recording is now part of Sessions.** A recording does not exist without a session, it is played from the session's own detail page, and both pages carried their own version of what a session does not capture. It is a section with its own heading rather than a line in passing: the two mechanisms, playing one, what it holds, and what decides whether it exists. The eight pages that linked it link the section, and the old address redirects.

**The page plays a recording instead of showing a picture of one.** A recording is a terminal moving through time, which is precisely what a screenshot cannot carry. The page replays a real session — nginx down because another process holds port 80, found and fixed in under a minute — from an asciicast in the repository, using the package the console itself plays recordings with, driven by a copy of the console's controls. The console's component is React and would be the only hydrated island on this site, so this is a plain script.

Two defects surfaced while doing it, both invisible until something other than a code block appeared in the prose. The copy-button script wraps every `<pre>` it finds, and the player draws its terminal into one, so it was lifting the terminal out of the player; and the code-block styles painted padding and a border onto it.

What is turned on also stops being a section with a screenshot of a switch: it is a namespace setting, one sentence. And the privacy paragraph, which listed what a terminal can show, is replaced by the controls that exist — recording is off until somebody turns it on, and playback is administrator and owner only.


gustavosbarreto:
Folded into a single docs branch (#7091 shortly) instead of stacking on a merged PR.

